PR #485, opened 2015-01-23
Add full circle area function

leogargu:
*Side note*: `RS_Ellipse` handles full ellipses and ellipse arcs, but for circles we have `RS_Circle` for full circles only and `RS_Arc` for circular arcs. Is there any reason to keep these as separate classes? Why not add `angle1` and `angle2` to `RS_Circle` like in `RS_Ellipse`?

### DIFF
--- a/librecad/src/lib/engine/rs_circle.cpp
+++ b/librecad/src/lib/engine/rs_circle.cpp
@@ -743,6 +743,21 @@ LC_Quadratic RS_Circle::getQuadratic() const
     return ret;
 }
 
+
+/**
+* @brief Returns area of full circle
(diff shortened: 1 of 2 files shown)


lordofbikes:
Thanks for your contribution

lordofbikes:
Concerning your side note, these classes are related to DXF entities. In DXF definition are entities CIRCLE, ARC and ELLIPSE, which are handled by these classes.

Second point is possibly performance. Computing with full circles is often much simpler than with arcs. In a combined class we'll have a lot of if/else checking for start angle equals end angle or start is 0° and end is 360°. Iterating through a document this summarizes and may slow down performance. With separate classes checking the angles can be avoided for circles.

Armin

Issue #500, opened 2015-02-02
Statusbar selector


kromanenko:
LibreCAD starts with Wtatusbar. 
Check in menu View: statusbar is not checked there. 
![statusbar](https://cloud.githubusercontent.com/assets/10499375/6011125/51400834-ab41-11e4-822a-f929f032129b.png)

Environment: Ubuntu 14,04
master, on commit ae5cc64
Windows, build: 2.0.7-33-g9ccfb49

PR #507, opened 2015-03-27
Fixing issue #500. Setting the checkbox for the Statusbar option in the view menu.

korhadris:
The Statusbar option in the View menu is now checked by default to reflect that the status bar is displayed by default. This should fix issue #500.


dxli:
thanks!

PR #510, opened 2015-04-02
Some hacks to the formulars to convert between radians,degrees and gradi...

Ngassa:
Thought I should optimize some of the formulars for conversions between radians, gradians and degrees in the rs_math.cpp file.


dxli:
Hi Ngassa,

Yes, this is the right direction, and please try to explore as much as possible.

Thanks!

PR #538, opened 2015-06-17
MakerCamSVG to support splines

dinkel:
This extends the MakerCamSVG exporter with spline support. It converts closed and open degree 2  and 3 splines to a series of quadratic resp. cubic bézier curves (the only thing supported by SVG). It is an exact, although not easily reversable, conversion.

Degree 1 splines are not supported because they aren't correctly displayed in LibreCAD anyways (it would just be polylines and not these strange chopped corners and it seems to me, that it would result in some confusion if they don't look the same.

The DXF format would allow to give weights to the different control points (DXF would allow NURBSs), which is not implemented in LibreCAD. This isn't a terribly bad thing for the SVG converter, because one cannot transform NURBS into bézier curves without loss.

It would be cool if at least one experienced developer could have a look at this, before I would push it to master.

### DIFF
--- a/librecad/src/lib/generators/lc_makercamsvg.cpp
+++ b/librecad/src/lib/generators/lc_makercamsvg.cpp
@@ -581,7 +581,7 @@ void LC_MakerCamSVG::writeSplinepoints(LC_SplinePoints* splinepoints) {
     writeQuadraticBeziers(splinepoints->getControlPoints(), splinepoints->isClosed());
 }
 
-void LC_MakerCamSVG::writeCubicBeziers(std::vector<RS_Vector> control_points, bool is_closed) {
+void LC_MakerCamSVG::writeCubicBeziers(const std::vector<RS_Vector> &control_points, bool is_closed) {
 
     std::vector<RS_Vector> bezier_points = calcCubicBezierPoints(control_points, is_closed);
 
@@ -602,7 +602,7 @@ void LC_MakerCamSVG::writeCubicBeziers(std::vector<RS_Vector> control_points, bo
     xmlWriter->closeElement();
 }
 
-void LC_MakerCamSVG::writeQuadraticBeziers(std::vector<RS_Vector> control_points, bool is_closed) {
+void LC_MakerCamSVG::writeQuadraticBeziers(const std::vector<RS_Vector> &control_points, bool is_closed) {
 
     std::vector<RS_Vector> bezier_points = calcQuadraticBezierPoints(control_points, is_closed);
 
@@ -623,7 +623,7 @@ void LC_MakerCamSVG::writeQuadraticBeziers(std::vector<RS_Vector> control_points
     xmlWriter->closeElement();
 }
 
-std::vector<RS_Vector> LC_MakerCamSVG::calcCubicBezierPoints(std::vector<RS_Vector> control_points, bool is_closed) {
+std::vector<RS_Vector> LC_MakerCamSVG::calcCubicBezierPoints(const std::vector<RS_Vector> &control_points, bool is_closed) {
 
     std::vector<RS_Vector> bezier_points;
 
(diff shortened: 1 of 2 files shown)


librecad/src/lib/generators/lc_makercamsvg.cpp
```diff
@@ -543,6 +550,193 @@ void LC_MakerCamSVG::writeEllipse(RS_Ellipse* ellipse) {
...
+
+    if (is_closed) {
+
+        for (int i = 0; i < (control_points_size - 1); i++) {
+            bezier_points.push_back(control_points[i]);
+
+            bezier_points.push_back((control_points[i] * 2.0 + control_points[i + 1]) / 3.0);
+            bezier_points.push_back((control_points[i] + control_points[i + 1] * 2.0) / 3.0);
+        }
+
```
dxli: I am not deep into this.

Is the linear interpolation form, (p0, (2/3)*p0 + (1/3)*p1, (1/3)*p0 + (2/3)*p1, p1), leading to quadratic spline points (lc_splinepoints)?

If not, we may want to consider a second order spline interpolation.

librecad/src/lib/generators/lc_makercamsvg.cpp
```diff
@@ -543,6 +550,193 @@ void LC_MakerCamSVG::writeEllipse(RS_Ellipse* ellipse) {
...
+        // Auxiliary points for easier calculation
+        bezier_points.insert(bezier_points.begin(), ((control_points[control_points_size - 1] + control_points[0] * 2.0) / 3.0));
+        bezier_points.push_back((control_points[0] * 2.0 + control_points[1]) / 3.0);
+
+        bezier_points_size = bezier_points.size();
+
+        for (int i = 1; i < (bezier_points_size - 1); i += 3) {
+            bezier_points[i] = ((bezier_points[i - 1] + bezier_points[i + 1]) / 2.0);
+        }
+
```
dxli: I see, this loop leads to second oder bezier, and this answers my comment on line 642

dxli: splinepoints class holds a quadratic spline passing given list of points(control points).

I feel the new cubic bezier won't pass precisely those initial control points any more.

dxli:
Hi dinkel,

I like your implementation, and it shows we could have a rather straight forward framework for spline/beziers.

If we can improve the old rs_spline, for;

1, keep only the control points, without having to approximate the spline by sequence of line segments;
2, implement getIntersection() with line, splinepoints(2nd order), and other conic curves;
3, getTangent(), etc;

For bezier, please also have a look at:

http://pomax.github.io/bezierinfo/

Thanks!

dxli

dinkel:
Hi dxli,

Thanks for your extensive reply.

First I have to make a correction to my pull request comment: In SVG both quadratic and cubic bézier curves are supported. I always knew and worked with this, but the sentence turned out to be very misleading (foreign languages can be tricky at times).

I intentionally did not touch the implementation(s) in the LibreCAD core (RS_Spline and LC_SplinePoints) and only used as little logic from both as necessary. This is, because I also think that there is some work to do on the splines implementation, but I did not want to interfere with possible work (or thinking) being done currently on this topic. Also I was a bézier/spline newbie before working on the exporter.

I agree, that it seems doable, to have a more complete framework for splines and your bullet points seem very reasonable and in the correct ordering to have a path to work. One thing I would prepend, would be to merge RS_Spline and LC_SplinePoints into only one implementation.

I would offer to work on a better spline implementation, but I guess this would take some time. What do you think about that? I remember you mentioned that @rvt is to be contacted before working on it to exchange ideas?

Concerning this pull request: Do you think I could merge it to master?


Best regards,

Dinkel

dxli:
Hi Dinkel,

The code looks clean enough to me, and if it works for some basic test cases at your side, we should merge it. We will do bug fixing and code cleanup continuously for sure.

If you want the new implementation in v3, we can work with Ries. If you want to start with v2, then, Ries won't get involved.

Regards,

dxli

dinkel:
Hi dxli,

Just merged the new functionality to the exporter. I did quite a few local and unfortunately all manual tests to make sure hopefully everything works as expected.

Concerning the spline-rewrite: I will have a look at v2 first and try to understand the stuff (both the math and the functionality).

dxli:
Hi Dinkel,

We can work together here :)

It's very interesting to get spline/bezier for both v2 and v3. (we will do v2 first, and port it to v3 later).

dxli

PR #539, opened 2015-07-08
Fixed regression of ellipse creation in MakerCamSVG arising from f123224

dinkel:
After isArc() was renamed to isEllipticArc() in f123224, the generation of ellipses ia SVG elements failed in "Export to MakerCAM SVG".

This is a simple fix, that renames the function call also in lc_makercamsvg.cpp


dinkel:
As this is a totally trivial fix, I didn't see the need for someone to review it, and therefore merged it on my own - I hope this is OK?

PR #540, opened 2015-07-09
- Removed return statement at the end of…

mhcuervo:
…lineWidth2dxfInt(enum lineWidth lw). All cases are covered in the switch construct, and also a default return value was specified.

### DIFF
--- a/libraries/libdxfrw/src/drw_base.h
+++ b/libraries/libdxfrw/src/drw_base.h
@@ -398,7 +398,6 @@ public:
         default:
             return -3;
         }
-        return static_cast<int> (lw);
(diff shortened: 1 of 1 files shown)


lordofbikes:
I'd prefer a more clear solution like this:
```
    default:
        break;
    }

    return -3;
}
```

Rallaz should check this, if he want to apply this to *libdxfrw* library first and then update LibreCAD or patch both, library and LibreCAD.

Anyhow, thanks for contributing.

mhcuervo:
Should I close this pull request?

mhcuervo:
-

PR #541, opened 2015-07-11
Minor correction on libdxfrw.

mhcuervo:
libdxfrw: Removed unnecessary return statement from "lineWidth2dxfInt(enum lineWidth lw)".


lordofbikes:
Thanks!

PR #542, opened 2015-07-12
implemented QColorDialog and added more default colors

r-a-v-a-s:
* added a variety of default colors to each ComboBox
* made the default colors words instead of unexpected hex codes
* implemented QColorDialog
* rearrangement of the dialog window, because the language ComboBox cramps the color section

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -953,8 +953,8 @@ void QC_ApplicationWindow::initActions(void)
 	actionFactory.addGUI(subMenu, actionHandler, RS2::ActionDrawText);
 
     // Hatch:
-	actionFactory.addGUI(menu, actionHandler, {RS2::ActionDrawHatch,
-											   RS2::ActionDrawImage}, RS2::ToolBarMain);
+    actionFactory.addGUI(menu, actionHandler, {RS2::ActionDrawHatch,
+                                               RS2::ActionDrawImage}, RS2::ToolBarMain);
     // Dimensioning actions:
     //
 #ifdef __APPLE1__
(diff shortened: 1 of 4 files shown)


lordofbikes:
Nice work!

A few thoughts:
* some of the color names are more cryptic than the HTML color codes to me
* the color dialog already has standard colors, the default color combo boxes are obsolete then
* add a kind of color indicator instead (QLabel/QFrame, background color set to selection), then we have a visual representation of the selection
* make the current setting the initial color of the color dialog
* I'd like a possibility to reset to standard color when change was a mistake or doesn't match
* if we set the first custom color of the color dialog to the LibreCAD standard, we can achieve this
* for custom colors be sure to set QColorDialog::DontUseNativeDialog to support our OS X users

There is a new *Icons* section in the preferences dialog, I assume this is for another feature, so we should add the new section with this other feature.

r-a-v-a-s:
> some of the color names are more cryptic than the HTML color codes to me

:open_mouth: I doubt many people share your perspective here.

> the color dialog already has standard colors, the default color combo boxes are obsolete then

True; however, I think it's nice to have the option to type in the names.

> add a kind of color indicator instead (QLabel/QFrame, background color set to selection), then we have a visual representation of the selection

Agreed.

> make the current setting the initial color of the color dialog

Agreed.

> I'd like a possibility to reset to standard color when change was a mistake or doesn't match
if we set the first custom color of the color dialog to the LibreCAD standard, we can achieve this

Agreed.

> for custom colors be sure to set QColorDialog::DontUseNativeDialog to support our OS X users

I did use it. Maybe I'm not understanding what you mean.

> There is a new Icons section in the preferences dialog, I assume this is for another feature, so we should add the new section with this other feature.

Yes the new layout was first achieved while working on my new toolbar system.

r-a-v-a-s:
> add a kind of color indicator instead (QLabel/QFrame, background color set to selection), then we have a visual representation of the selection

I decided against this. Call me a minimalist or lazy, but I don't think the added code complexity is worth it when you can just click the colordialog button and see what it looks like.

Unless there is some error, this is my final product. I will be moving on to the issue of the layer list / block list selection color.

PR #545, opened 2015-07-14
added layer/block selection color option

r-a-v-a-s:
The `.ui` file shows a lot of changes,
but this was mostly an automatic update.
I had to break the grid layout and then reapply it to fix a bug.
I'm guessing Qt has tweaked the grid layout since it was first applied in our program.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -1408,6 +1408,13 @@ void QC_ApplicationWindow::initView() {
         addDockWidget(Qt::BottomDockWidgetArea, dw);
 
     RS_DEBUG->print("  done");
+
+    RS_SETTINGS->beginGroup("/Appearance");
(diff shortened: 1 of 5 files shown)


dxli:
wondering whether we can add http://doc.qt.io/qt-4.8/qcolordialog.html to select colors.

lordofbikes:
QColorDialog was recently implemented by ravas for *Application Preferences*, see commit 08fe8f132ca7b22ff34cf7d6674bbbf813e948bd / #542

r-a-v-a-s:
I used the same `set_color` function, which uses the QColorDialog.

The change to the `.ui` can be seen at the bottom of the color section.

appearance

r-a-v-a-s:
If no problems are noted, then I will merge my requests after 24 hours.
To reduce this time reviewers can comment: "No problems noticed."
To increase this time reviewers can state when they plan on reviewing.

dxli:
I don't see any problem here.

PR #547, opened 2015-07-15
added link to wiki

r-a-v-a-s:
This launches the default browser and goes to our wiki.
The links under Help → Manual do not work.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -3673,5 +3673,5 @@ void QC_ApplicationWindow::updateWindowTitle(QWidget *w)
 
 void QC_ApplicationWindow::goto_wiki()
 {
-    QDesktopServices::openUrl(QUrl("http://wiki.librecad.org/index.php/Main_Page"));
+    QDesktopServices::openUrl(QUrl("http://wiki.librecad.org/"));
 }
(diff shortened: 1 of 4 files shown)


librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -3664,3 +3670,8 @@ void QC_ApplicationWindow::updateWindowTitle(QWidget *w)
         w->setWindowTitle(w->windowTitle()+m_qDraftModeTitle);
     }
 }
+
+void QC_ApplicationWindow::goto_wiki()
+{
+    QDesktopServices::openUrl(QUrl("http://wiki.librecad.org/index.php/Main_Page"));
```
lordofbikes: Keep the URL as short as possible:

`QUrl("http://wiki.librecad.org/")`

When we move the Wiki, the direct page link becomes invalid.

r-a-v-a-s: Any idea when the move will happen?

r-a-v-a-s:
I couldn't resist the commit summary... partly because I have a 14 letter German last name. :innocent: :de:

dxli:
is it okay for us to refer the user manual link to wiki as well?

librecad/src/main/qc_applicationwindow.cpp
lordofbikes: Not yet, limited time and the heat here delays my progress on troubleshooting the content transfer tool.
Hope I can finish this in my holidays in August/September.

lordofbikes:
I'd agree, linking to the Wiki is better than the empty manual now.
Or let us change the manual content and add the Wiki link there.

r-a-v-a-s:
I think this is a sufficient temporary fix.

I want to talk about working on the help manual on the forum soon.

PR #561, opened 2015-09-04
Extension of command-line autocompletion

Rupicapra-rupicapra:
Implemented bash-behavior for autocompletion : when typing 'sp<tab>', the CL will automatically prompt 'spline' (longest common root for commands starting with 'sp' -> 'spline' & 'spline2').

I would like to pack this piece of code into a function which would return a QString (it would seem more clean). May I use a QG_CommandWidget private method for this ? (e.g. QString autoCompletion(QStringList  const & shortCmdList);).
Rupi

### DIFF
--- a/librecad/src/ui/forms/qg_commandwidget.cpp
+++ b/librecad/src/ui/forms/qg_commandwidget.cpp
@@ -189,24 +189,16 @@ void QG_CommandWidget::setNormalMode() {
 }
 
 QString QG_CommandWidget::getRootCommand( const QStringList & cmdList ) {
-    QString longestString;
     QString shortestString;
     int lengthShortestString(0);
-    int lengthCmdList = cmdList.count();
 
-    // Finding which is the longest string
-    for(QStringList::const_iterator it = cmdList.begin(); it != cmdList.end(); ++it) {
-        if((*it).length() > longestString.length()) {
-            longestString = (*it);
-        }
-    }
-    int lengthLongestString = longestString.length();
+    shortestString = cmdList.at(0);
+    lengthShortestString = shortestString.length();
 
-    // Finding which is the shortest string
-    lengthShortestString = longestString.length();
     for(QStringList::const_iterator it = cmdList.begin(); it != cmdList.end(); ++it) {
         if((*it).length() < lengthShortestString) {
             shortestString = (*it);
+            lengthShortestString = shortestString.length();
         }
     }
     lengthShortestString = shortestString.length();
@@ -218,7 +210,7 @@ QString QG_CommandWidget::getRootCommand( const QStringList & cmdList ) {
 
     for(i = 0; i < lengthShortestString; ++i) {
         for(QStringList::const_iterator it = cmdList.begin(); it != cmdList.end(); ++it) {
-            if(longestString.at(i) != (*it).at(i)) {
+            if(shortestString.at(i) != (*it).at(i)) {
                 common = false;
                 break;
             }
@@ -227,7 +219,7 @@ QString QG_CommandWidget::getRootCommand( const QStringList & cmdList ) {
             ++pos;
         }
     }
-    return longestString.left(pos);
+    return shortestString.left(pos);
 
 }
 
--- a/librecad/src/ui/forms/qg_commandwidget.cpp
+++ b/librecad/src/ui/forms/qg_commandwidget.cpp
@@ -157,47 +157,7 @@ void QG_CommandWidget::tabPressed() {
             leCommand->setText(reducedChoice.first());
         }
         else if (reducedChoice.count()>0) {
-        //TODO: unix-like behaviour for autocompletion
-            QString longestString = "";
-            QString shortestString = "";
-            int lengthShortestString(0);
-            int lengthReducedChoice = reducedChoice.count();
-
-            // Finding which is the longest string
-            for(QStringList::Iterator it = reducedChoice.begin(); it != reducedChoice.end(); ++it) {
-                if((*it).length() > longestString.length()) {
-                    longestString = (*it);
-                }
-            }
-            int lengthLongestString = longestString.length();
-
-            // Finding which is the shortest string
-            lengthShortestString = longestString.length();
-            for(QStringList::Iterator it = reducedChoice.begin(); it != reducedChoice.end(); ++it) {
-                if((*it).length() < lengthShortestString) {
-                    shortestString = (*it);
-                }
-            }
-            lengthShortestString = shortestString.length();
-
-            // Now we parse the reducedChoice list, character of each item by character.
-            int i(0);
-            int pos(0);
-            bool common = true;
-
-            for(i = 0; i < lengthShortestString; ++i) {
-                for(QStringList::Iterator it = reducedChoice.begin(); it != reducedChoice.end(); ++it) {
-                    if(longestString.at(i) != (*it).at(i)) {
-                        common = false;
-                        break;
-                    }
-                }
-                if(common == true) {
-                    ++pos;
-                }
-            }
-
-            QString proposal = longestString.left(pos);
+            QString proposal = this->getRootCommand(reducedChoice);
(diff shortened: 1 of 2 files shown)


librecad/src/ui/forms/qg_commandwidget.cpp
```diff
@@ -155,7 +157,48 @@ void QG_CommandWidget::tabPressed() {
             leCommand->setText(reducedChoice.first());
         }
         else if (reducedChoice.count()>0) {
+            QString longestString = "";
+            QString shortestString = "";
```
dxli: the initialization is not necessary, because default ctor of QString generates an empty string

librecad/src/ui/forms/qg_commandwidget.cpp
```diff
@@ -155,7 +157,48 @@ void QG_CommandWidget::tabPressed() {
...
+            for(i = 0; i < lengthShortestString; ++i) {
+                for(QStringList::Iterator it = reducedChoice.begin(); it != reducedChoice.end(); ++it) {
+                    if(longestString.at(i) != (*it).at(i)) {
+                        common = false;
+                        break;
+                    }
+                }
+                if(common == true) {
+                    ++pos;
+                }
```
dxli: This algorithm would search character by character, so for length L with N choices, has a cost of O(N L)

I think we can do it faster here by bisection to make it O( N lg L):
the code is not tested yet

the range should be within [typed.size(), shortestString.size()]:
```C++
int low=typed.size();
int high=lengthShortestString;
QString proposal=shortestString.left(low);
while(high>low){
    int mid=(low+high)/2;
    bool common=true;
    proposal=shortestString.left(mid);
    for(auto const& str: reducedChoice){
          if(!str.startWidth(proposal)){
          common=false;
          break;
    }
    if(common){
         low=mid+1;
    }else{
         high=mid-1;
    }
}
}
```

librecad/src/ui/forms/qg_commandwidget.cpp
```diff
@@ -155,7 +157,48 @@ void QG_CommandWidget::tabPressed() {
             leCommand->setText(reducedChoice.first());
         }
         else if (reducedChoice.count()>0) {
+            QString longestString = "";
+            QString shortestString = "";
+            int lengthShortestString(0);
+            int lengthReducedChoice = reducedChoice.count();
+
+            // Finding which is the longest string
```
dxli: a common string to propose is a substring of the shortest, and the longest string can never be proposed.

If it's the case, we don't have to search for the longest.

librecad/src/ui/forms/qg_commandwidget.cpp
Rupicapra-rupicapra: Ok, I see. An old habit from the times I learned C ('always initialize your variables', he said...).

I packed that code into a dedicated method that returns the longest common root on another branch. I would like to have your opinion on [this commit](https://github.com/Rupicapra-rupicapra/LibreCAD/commit/07e447b43b15c42954f3bccd2f776cf49c7b9468). If it seems OK to you, I can replace this pull request by one about these changes.
I put the method in the private part 'by default', because I don't think it's necessary to access it from the outside of the command-line, but maybe it's a mistake.

dxli:
since this autoCompletion() should be a slot, to be triggered by a "Tab" pressed event, for example.

librecad/src/ui/forms/qg_commandwidget.cpp
dxli: It's still true, unless you know there's a default ctor for type. Otherwise, it's undefined behavior for primitive types, the same as in C, or a compiling time error (no default ctor).

you don't have to make a new pull request, as you can just keep pushing your commits to this branch.

librecad/src/ui/forms/qg_commandwidget.cpp
Rupicapra-rupicapra: Yep, I also realized it. I changed the code so it does not needs to use longest string. I'm currently working on algorithm optimization.

librecad/src/ui/forms/qg_commandwidget.cpp
Rupicapra-rupicapra: I have a little question : what does this syntax stands for ?
`auto const& str: reducedChoice`
Is it an equivalent for a 'foreach' loop / a 'for' loop in python (eg `for string in reducedChoice`) ?

dxli: yes, but foreach() is Qt preprocessor, while [range-based loop](http://en.cppreference.com/w/cpp/language/range-for) is c++11:

for(auto str: reducedChoice) is okay, because of QString default data sharing, but there's no drawback by specifying it's a const reference.

dxli: we started to support range-based loop by adding begin()/end() member functions to our containers like: RS_EntityContainer RS_VectorSolutions

Rupicapra-rupicapra: Hum, this code gives me a strange behavior : 'sp/tab/' -> `spline` -> /tab/ -> nothing (instead of keeping `spline`); 'z/tab/' -> `zoom` -> /tab/ -> `zoomp` -> /tab/ -> `zoompan` (instead of staying on `zoom`).

Rupicapra-rupicapra: Okay, thanks for the explanation.

librecad/src/ui/forms/qg_commandwidget.cpp
Rupicapra-rupicapra: Yep, I know, but I keep this habit for all data types, even for those which does not need it.

I pushed some commits to fix the behavior I had with the bisection, and created a slot for this method.
I didn't know that pushing on a branch after a PR added new changes to that PR. It's useful to know this.

dxli:
Do you have something else Work-in-Progress?

Otherwise, we can merge your work into to the master branch. I think we should port this to the 2.0 branch as well.

Rupicapra-rupicapra:
Not yet on this part.
I would like to work on a zsh-like behavior (tab once -> common root; <tab> -> first command in reduced list; <tab> -> next command...), but it seems much more complicated. So I think I will wait that next point is finished before working on it.
I also want to work on the 'circle with center and radius' module : adding a command-line shortcut for this tool (maybe `cc` or `cr`, I have to check if they are already used or not) and a command-line option to change radius (like in 'parallel line' tool, where you can give distance in the prompt)).
But, as it is not really linked (excepted the command shortcut), I think you can merge.

Thanks for your help on this topic. I learned a lot.

dxli:
adding commands/short commands is relatively easy following the format in the ctor of rs_commands

about command lines, there have been some many ideas emerged in past:

* simple improvement, can we supply everything in one line:
```bash
line 0,0 @1,1
circle 0,0 5
ellipse 0,0 @1,1 @-0.5,0.5 @1,0 @0,1
```

for ellipse, it's not so clear, but I'm trying to supply all coordinates along the action rs_drawEllipseAxis, but it's clear enough for line 2points, circle center/radius, etc.

* python binding, scripts by python;

* accepting commands by piping, so can be easily adopted to other programming languages

Just some brainstorming.

Have a nice day!

Rupicapra-rupicapra:
Yep, but I planned to wait for the option to set the fixed radius on command line to be functional. To do everything I want to do about this command in the same time.

It seems possible, but I don't really know how. Maybe it would need to overload the ctor for line & circle (I will have a deeper look into the code later).

Python scripting would be great, but I don’t think I have the skills to develop an interface C++ / Python*. Anyway, I think it's very interesting, so I'll look at the existing code and see.
* I have some knowelge of 2 Python libs designed to generate / edit dxfs (dxfwrite & ezdxf), if needed.


What do you mean by 'accepting commands by piping' ? Getting parameters from an external application / language (like LibreOffice accepts commands from Python typed from the outside of the app via an API) ?

Have a nice day too.

PR #564, opened 2015-09-08
Some changes into UI to allow more keyboard-driven actions

Rupicapra-rupicapra:
Shortcuts for drawing order added (suggestion).
Shortcut for "New Layer" added.
Focus in the Layer Dialog is given to the "Layer Name" field at opening.


librecad/src/actions/rs_actionlayersadd.cpp
```diff
@@ -37,6 +37,7 @@ RS_ActionLayersAdd::RS_ActionLayersAdd(RS_EntityContainer& container,
 
 QAction* RS_ActionLayersAdd::createGUIAction(RS2::ActionType /*type*/, QObject* /*parent*/) {
     QAction* action = new QAction(QIcon(":/ui/layeradd.png"), tr("&Add Layer"), NULL);
+    action->setShortcut(QKeySequence("Ctrl+L"));
```
dxli: Please add a tooltip as well

PR #567, opened 2015-09-10
gui update

r-a-v-a-s:
http://forum.librecad.org/action-factory-and-toolbars-td5711930.html
http://forum.librecad.org/new-toolbars-td5712125.html

Primary reasons to adopt this:
- easier for developers to understand / modify
- avoids going in and out of categories constantly
- users can decide what tools are exposed and where
- allows for user defined custom icon sizes


dxli:
great!

Let's merge this!

PR #571, opened 2015-09-16
Helper, shortcut & bug correction for "Circle with center & radius" tool

Rupicapra-rupicapra:
##### Help tip
When in mode "circle with center and radius", typing "radius"
in the command-line allow the user to set the radius without moving
hands off the keyboard, but the helper message didn't mentionned that
possibility.

##### Bug
When setting radius from commandline, pressing enter to validate
causes LibreCAD to draw a circle of selected radius at relative zero
position (if no mouse activity into the container) or crosshair
position (if mouse activity into the container).
I added a `switch` test to the `trigger()` method to test the mode, and
if the user is in mode "SetRadius", circle is not drawn & relative zero
is not moved.

##### Command shortcuts
`circlecr` and `cc`

### DIFF
--- a/librecad/src/actions/rs_actiondrawcirclecr.cpp
+++ b/librecad/src/actions/rs_actiondrawcirclecr.cpp
@@ -73,7 +73,15 @@ void RS_ActionDrawCircleCR::trigger() {
 									  *data);
     circle->setLayerToActive();
     circle->setPenToActive();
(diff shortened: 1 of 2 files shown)


dxli:
Thanks!

Rupicapra-rupicapra:
You're welcome.

PR #574, opened 2015-09-21
anti-aliasing

r-a-v-a-s:
`RS_SETTINGS->readNumEntry("/Appearance/Antialiasing")`
best case amounts to 6 function calls, 3 expression evaluations and a map lookup;
therefore, using the dedicated bool variable `antialiasing` is much cheaper.

By design this only applies anti-aliasing to the static objects, not previews or overlays.

### DIFF
--- a/librecad/src/lib/engine/rs_settings.cpp
+++ b/librecad/src/lib/engine/rs_settings.cpp
@@ -147,21 +147,14 @@ QByteArray RS_Settings::readByteArrayEntry(const QString& key,
 
 }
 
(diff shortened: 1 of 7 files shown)


r-a-v-a-s:
pushing this without the indentation change

dxli:
@r-a-v-a-s If you do change a line, it's reasonable to use tab indentation. Armin is right, if the change is only about indentation, it's getting confusing.

Looks like we should share our Qt-Creator editor settings

PR #577, opened 2015-09-23
Fixed bug in hatch dialog box

Rupicapra-rupicapra:
Bug introduced by commit 776c8ae4dbd86ae658202a2174895ce340b0b1dd (`git bisect` information)
Changing from `if(hatch==nullptr)` to `if(hatch)` caused the hatch
dialog box not to show when called.
It seems that this change fixed it... Will see.

Because I reverted a change which happened few times ago, I prefer asking : why was this line changed ? Does it have a goal in particular, that I would have broken by reverting ?

### DIFF
--- a/librecad/src/ui/qg_dialogfactory.cpp
+++ b/librecad/src/ui/qg_dialogfactory.cpp
@@ -1773,7 +1773,7 @@ bool QG_DialogFactory::requestTextDialog(RS_Text* text) {
  * Shows a dialog to edit pattern / hatch attributes of the given entity.
  */
 bool QG_DialogFactory::requestHatchDialog(RS_Hatch* hatch) {
-	if (hatch) return false;
+	if (hatch==nullptr) return false;
 
     RS_PATTERNLIST->init();
 


dxli:
@Rupicapra-rupicapra :+1: 

Thanks a lot Jerry! Looks like it was me who broke it :-1:

Rupicapra-rupicapra:
You're welcome.
That may happen. Maybe you wanted to replace `if(hatch==nullptr)` by `if(!hatch)` like you did for text tool on [line 1758](https://github.com/LibreCAD/LibreCAD/commit/776c8ae4dbd86ae658202a2174895ce340b0b1dd#diff-bccd21648950850b5d8f3ae3eebe4e3fR1758) but forgot the `!` (text tool is OK with this change). I'll try to change to `if(!hatch)`, and if it's OK I'll include it in a further PR.

dxli:
nullptr can be converted to false by implicit conversion

nullptr was introduced to avoid implicit conversion to int, so, nullptr != 2 is an compiler error
```C++
if  (! nullptr) //ok
if (nullptr == NULL) //ok, both converted to boolean by implicit conversion
if (nullptr == 2 ) // compiler error
```

PR #591, opened 2015-10-21
Added void-type to signals

Rupicapra-rupicapra:
Signals without void type was causing compilation failure.

### DIFF
--- a/librecad/src/lib/gui/rs_graphicview.h
+++ b/librecad/src/lib/gui/rs_graphicview.h
@@ -464,8 +464,8 @@ private:
 	bool m_bIsCleanUp=false;
 
 signals:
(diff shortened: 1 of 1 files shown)


r-a-v-a-s:
>_< Thanks! Weird that it didn't complain when I compiled.

Rupicapra-rupicapra:
You're welcome.
Do you use `-fpermissive` (g++) flag or equivalent (another compiler) at compilation ? If yes, it "hides" this kind of errors.

r-a-v-a-s:
Thanks for the tip. The only thing I changed from the defaults in Qt Creator is adding `-j` to the make arguments. I didn't see `-fpermissive`.

PR #600, opened 2015-11-09
sundry user-interface improvements

Osndok:
Each one is in a separate commit to (hopefully) be more reviewable.


r-a-v-a-s:
Welcome! It seems like mostly agreeable changes.
We definitely need more distinct icons for several actions.

Some tips for future contributions:
- consider smaller pull requests (otherwise if one commit is rejected, then everything will be)
  - you can make new branches and submit a pull request for each one
- consider posting about things you are thinking about changing (issue section or on the forum)
  - this can mitigate overlapping work or rejection

Osndok:
FYI, you should be able to cherry-pick commits straight from my fork, I wasn't trying to imply that you have to merge them all.

Osndok:
I think there is an issue with the added buttons... since they are "clones" of the other tasks, I think that they have somehow displaced three of the other line-drawing options.

dxli:
@Osndok,

Any idea about SVG icons?

I feel it's easy to implement, and could be life saver for HD displays.

Osndok:
@dxli I like the theory of SVG icons, but I'm not 100% sure that "SVG" (per se) would be the best bet. Many of the icons are so simple, for example, that they could even be hard-coded (line+circle?).

More to the point of this thread, as @r-a-v-a-s asked, I am going to close this (now conflicting) pull request, and re-issue it as several smaller, more tasty, morsels.

PR #607, opened 2015-11-12
Better zoom-field navigation through basic ZUI "flight" theory

Osndok:
When scrolling with the mouse wheel, you should find:
+ Far less "inch worming" (e.g. mouse top-left, zoom in, mouse bottom-right, zoom out)
+ Zooming out tends to reveal more logical/useful canvas
+ Zooming in tends to center objects that were under the cursor
+ Zooming in on objects near the edge of the view port should now be in a useful position more often
- Seasoned veterans might need to get used to the "drift" or "not working around the old zoom mechanic".
- Not as effective for "just playing with the grid lines" [i.e. zoom_in != inverse(zoom_out)]

Relatedly:
- Zooming is still a bit too choppy, IMO.
- Snapper/Guides are still not redrawn after a zoom event.

### DIFF
--- a/librecad/src/ui/qg_graphicview.cpp
+++ b/librecad/src/ui/qg_graphicview.cpp
@@ -333,7 +333,7 @@ bool QG_GraphicView::event(QEvent *event) {
 
             // It seems the NativeGestureEvent::pos() incorrectly reports global coordinates
             QPoint g = mapFromGlobal(nge->globalPos());
-            RS_Vector mouse = toGraph(RS_Vector(g.x(), g.y()));
+            RS_Vector mouse = toGraph(g.x(), g.y());
             setCurrentAction(new RS_ActionZoomIn(*container, *this, direction,
 												 RS2::Both, &mouse, factor));
         }
@@ -450,7 +450,7 @@ void QG_GraphicView::wheelEvent(QWheelEvent *e) {
         return;
     }
 
-    RS_Vector mouse = toGraph(RS_Vector(e->x(), e->y()));
+    RS_Vector mouse = toGraph(e->x(), e->y());
 
 #if QT_VERSION >= 0x050200
     QPoint numPixels = e->pixelDelta();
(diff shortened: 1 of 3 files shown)


librecad/src/ui/qg_graphicview.cpp
```diff
@@ -544,14 +544,38 @@ void QG_GraphicView::wheelEvent(QWheelEvent *e) {
 
     // zoom in / out:
     else if (e->modifiers()==0) {
+
+		/*
+		 * FIXME - there is surly a better way to get the center of the main view port. 
+		 */
+		RS_Vector mainViewCenter = toGraph(RS_Vector(getWidth()/2, getHeight()/2));
```
dxli: c++11 way should be handly

```C++
RS_Vector mainViewCenter = toGraph({getWidth()/2, getHeight()/2});
```

librecad/src/ui/qg_graphicview.cpp
```diff
@@ -544,14 +544,38 @@ void QG_GraphicView::wheelEvent(QWheelEvent *e) {
...
+
 		if (e->delta()>0) {
+			const double zoomInOvershoot=1.20;
+
+			RS_Vector effect=RS_Vector(mouse);
```
dxli: copy ctor can be simplified as

```C++
RS_Vector effect{mouse};
// or equivalent
//auto effect = mouse;
```

dxli:
zooming is choppy, probably because the default zooming factor in RS_ActionZoomIn is too large

```C++
 RS_ActionZoomIn(RS_EntityContainer& container,
					RS_GraphicView& graphicView,
                    RS2::ZoomDirection direction = RS2::In,
					RS2::Axis axis = RS2::Both,
					RS_Vector const* pCenter = nullptr,
					double factor = 1.25);
```
the factor = 1.25 might be reduced to 1.1, for example.

dxli:
the idea is to reveal a little bit from outside of viewport when zooming around mouse center is request.

I found it a great idea, and let's try to make it robust and intuitive enough.

Osndok:
You are right, changing the zoom factor did help a lot. I find that 1.10 is too slow, but 1.137 seems about right. I wonder, though, if there is a hidden tunable in QT that limits the update period for the scroll wheel, or if this is heavily dependent upon machine speed.

PR #610, opened 2015-11-12
Separate action icons for drawing lines from start vs end (workflow optimization)

Osndok:
...additionally, the resulting toolset has been re-arranged to "make sense to me".

librecad-tools-screenshot-n2

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -3144,25 +3144,32 @@ void QC_ApplicationWindow::menus_and_toolbars()
 
     list_a.clear();
 
+	/**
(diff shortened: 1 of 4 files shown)


r-a-v-a-s:
It would be useful if we could figure out a way to do this using the plugin system,
because I don't think these are worth the extra "noise"; although I realize they are worth it for you.

The mirroring of icons is an interesting approach;
however, if we did adopt these actions, then the desired solution would be new SVG icons.
Having all the tool icons in one folder will allow for a custom icon system (among other things);
and using SVG allows for scaling in the emergent age of 4-8k screens,
and allows for easier editing.

I'm closing this due to conflicting commits.
See LC_WidgetFactory to re-implement.

PR #637, opened 2016-01-30
Some minor performance enhancements

midzer:
Hi.
I ran the source code through latest cppcheck and found few parts where some optimizations are possible.

Project built nicely on Qt 5.5.1 under Debian Jessie.

### DIFF
--- a/librecad/src/lib/gui/rs_painter.h
+++ b/librecad/src/lib/gui/rs_painter.h
@@ -50,7 +50,7 @@ class QBrush;
  */
 class RS_Painter {
 public:
-    RS_Painter(): offset(RS_Vector(0.0,0.0)) {
+    RS_Painter(): offset{0.0,0.0} {
         drawingMode = RS2::ModeFull;
         drawSelectedEntities=false;
     }
--- a/librecad/src/lib/gui/rs_painter.h
+++ b/librecad/src/lib/gui/rs_painter.h
@@ -50,9 +50,8 @@ class QBrush;
  */
 class RS_Painter {
 public:
-    RS_Painter() {
+    RS_Painter(): offset(RS_Vector(0.0,0.0)) {
         drawingMode = RS2::ModeFull;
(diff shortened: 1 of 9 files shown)


librecad/src/lib/gui/rs_painter.h
```diff
@@ -50,9 +50,8 @@ class QBrush;
  */
 class RS_Painter {
 public:
-    RS_Painter() {
+    RS_Painter(): offset(RS_Vector(0.0,0.0)) {
```
dxli: @midzer 

```C++
RS_Painter(): offset{0.0,0.0}) {
```

dxli:
@midzer 

sorry for being late in reply :( I was a little bit busy lately

Yes, found an ugly part in LC, and thanks a lot :)

we really want continuous code review, and hopefully upgrade our legacy codebase in the process.

librecad/src/lib/gui/rs_painter.h
midzer: i dont know {} syntax to initialize a variable tbh

dxli: try to use aggregate initialization, even though for most time, it doesn't matter (the case here).

interesting difference:

```C++
struct A;
A a(); // declare a function A a() -> A {...}
A b{}; // define a variable a by default ctor of A
vector<int> v(2, 1); // v contains {1, 1}
vector<int> w{2, 1}; // w contains {2, 1}
```

midzer: thanks for clarification!
i did not get offset being an array :)

dxli:
@midzer 

please update your branch there, so we can quickly merge and close this one :+1:

midzer:
@dxli what do you mean by updating my branch? doing a fetch upstream & merge on my patch1 branch to update to head?

dxli:
@midzer 

Thanks

Issue #640, opened 2016-02-07
'Cancel' is the default button in the Move dialog


kfoltman:
As it is, the Modify/Move dialog box has 'Cancel' as the default button, which means that users that prefer to use a keyboard are likely to accidentally cancel the operation by pressing Enter.

On the other hand, Modify/Scale has OK as the default.

kfoltman:
After a short investigation, this is caused by the buttonBox widget (qg_dlgmove.ui) being the first, not the last element in the parent layout element. Moving the buttonBox to the end of the corresponding layout solves the problem.

Issue #639, opened 2016-02-07
selecting two polylines and "convert to polyline" leaves stray segments


kfoltman:
Steps to reproduce:

1. draw 5 connected lines using Draw / Line / 2 points
2. select first two lines
3. convert to polyline
4. select the remaining 3 lines
5. convert them to polyline
6. select the 2 resulting polylines
7. convert to polyline
8. select the resulting polyline (now it works fine)
9. deselect all (via Ctrl+K) - this part is important, I think
10. select the resulting polyline again (note the different display this time)
11. move it somewhere
12. notice the extra bunch of polyline segments

kfoltman:
Actually, step 8 only appears to work fine because some of the objects are still selected after step 7.

So, the shorter steps to reproduce:
1. draw a 2-segment polyline
2. draw a 3-segment polyline starting at the last point of the first one
3. select both
4. convert to polyline
5. mv 0,0 0,100, select "delete original" in the dialog
6. notice the remaining segments of the old polylines

kfoltman:
This seems to happen even if converting a single, 3-point polyline to a polyline (supposed to be a null or almost-null operation). It's more obvious to see with antialiasing enabled in settings, because single lines are then drawn in gray and duplicate/overlapping lines are drawn in white (i.e. twice). The other, less convenient way to notice the problem is to use the Info/List entities function to list the vertices of the produced polyline.

I think what may trigger the bug is the fact that in RS_ActionPolylineSegment::convertPolyline, selectedEntity is a polyline (RS2::EntityPolyline), but the entries added to 'remaining' list in the first loop are actually the individual segments of the polyline, i.e. RS2::EntityLine, due to the way iteration works. The end result is that the list of segments to chain contains both entire polyline mixed with polyline segments.

One fix that I've found was to use ResolveNone instead of ResolveAllButInserts in convertPolyline - in that case, polylines are not divided into segments in that iteration, which is probably a good thing, because it's not possible to select single line of a polyline anyway. However, it also means that other entity containers (inserts is one example I can think of) might not work correctly - and I don't even know what the correct behaviour should be in that case.

kfoltman:
The same goes for the RS_ActionPolylineSegment::init, it should also be a ResolveNone - otherwise another edge case gets broken (merging two polylines will only add the first segment of one polyline to be merged).

kfoltman:
Some edge cases that may be worth testing, and seem to work after changing the iteration mode:

[...]

PR #642, opened 2016-02-07
proposed fixes for bugs 639, 640, 643, 644 + a way to change radius on a polyline (including removal)

kfoltman:


kfoltman:
Note: bug639, bug640 should be issue #639, issue #640 - didn't know about two separate conventions for SF and Github tickets when I checked in that code, and AFAIK there's no easy way to change it now.

r-a-v-a-s:
Thanks :sunglasses: :+1: 
Everything looks OK to me... @dxli ?

"...AFAIK there's no easy way to change it now."
You could place the issue url in a comment on the associated commit to form a link.

dxli:
@kfoltman Thanks a lot.

I am out f town today, and i will review it tomorrow.

librecad/src/actions/rs_actionmodifyround.cpp
```diff
@@ -80,11 +81,16 @@ void RS_ActionModifyRound::trigger() {
...
 
         RS_Modification m(*container, graphicView);
+		RS_Entity* radius = nullptr;
+		if (changeRadius) {
+			radius = RS_Information::getRadiusArc(entity1, entity2);
+		}
 		m.round(pPoints->coord2,
 				pPoints->coord1,
-                (RS_AtomicEntity*)entity1,
+				(RS_AtomicEntity*)entity1,
```
dxli: minor coding style: try to avoid C-style casting, here entity1 and entity2 are both AtomicEntities, so static_cast is preferred over C-Style

librecad/src/actions/rs_actionpolylinesegment.cpp
```diff
@@ -54,9 +54,9 @@ void RS_ActionPolylineSegment::init(int status) {
 	if (container->countSelected(true, entityType)) {
 		//find a selected entity
 		//TODO, find a better starting point
-		for (RS_Entity* e = container->firstEntity(RS2::ResolveAllButInserts);
+		for (RS_Entity* e = container->firstEntity(RS2::ResolveNone);
```
dxli: we can use range based loop for ResolveNone

librecad/src/lib/information/rs_information.cpp
```diff
@@ -109,12 +110,12 @@ bool RS_Information::isTrimmable(RS_Entity* e1, RS_Entity* e2) {
                     e1->getParent()==e2->getParent()) {
 
                 // in the same polyline
-                RS_EntityContainer* pl = e1->getParent();
+                RS_Polyline* pl = (RS_Polyline *)e1->getParent();
```
dxli: getParent() is in base class, no need to cast?

librecad/src/lib/information/rs_information.cpp
```diff
@@ -109,12 +110,12 @@ bool RS_Information::isTrimmable(RS_Entity* e1, RS_Entity* e2) {
...
                 int idx1 = pl->findEntity(e1);
                 int idx2 = pl->findEntity(e2);
                 RS_DEBUG->print("RS_Information::isTrimmable: "
                                 "idx1: %d, idx2: %d", idx1, idx2);
-                if (abs(idx1-idx2)==1 || abs(idx1-idx2)==int(pl->count()-1)) {
+                if (abs(idx1-idx2)==1 || (pl->isClosed() && abs(idx1-idx2)==int(pl->count()-1))) {
```
dxli: Are the indices be reliable indicators for neighbors?

Looks like we'd better create API in polyline to handle this.

I'm worried about the indices after deletion/insertion.

kfoltman: Not sure - but at least that's how it worked before, and I haven't noticed any other bugs related to that. I've only changed it to ignore the 'first and last index' (wraparound) case for open polylines, and it at least seems to work as expected.

librecad/src/actions/rs_actionpolylinesegment.cpp
kfoltman: Do you agree that it needs to be ResolveNone though? I did a quick look at the types of entities that exist, and I don't think any of those needs to be descended into - certainly not blocks, because they're supposed to be separate entities and not to be changed by any operations on the main drawing. Maybe I missed something there, though.

librecad/src/lib/information/rs_information.cpp
kfoltman: The cast is applied to the result, in order to be able to assign it to a polyline pointer. I need it to be polyline pointer and not just a generic entity container, because I need access to the isClosed method. However, it might well be a static cast as you said above - the object type is verified beforehand.

librecad/src/actions/rs_actionpolylinesegment.cpp
dxli: polylines are of lines/arcs (wondering why not elliptical arcs as well), but doesn't accept, for example, another polyline as a member.

Maybe, some day, there can be a poly-polyline, but it's not attractive either.

librecad/src/lib/information/rs_information.cpp
dxli: looks like it's an API issue by design. The API requires that neighbors have consecutive indices and are actually connected by end/start points.

Probably too difficult to change this.

Ideally, we should provide an iterator here to guarantee looping through neighbors.

librecad/src/lib/information/rs_information.cpp
```diff
@@ -143,6 +144,31 @@ bool RS_Information::isTrimmable(RS_Entity* e1, RS_Entity* e2) {
...
+		e1->getParent()==e2->getParent()) {
+		RS_Polyline* poly = (RS_Polyline*)e1->getParent();
+		int idx1 = poly->findEntity(e1);
+		int idx2 = poly->findEntity(e2);
+		unsigned dist = abs(idx1 - idx2);
+		int mid = -1;
+		if (dist == 2)
+			mid = idx1 + (idx2 - idx1) / 2;
+		else if (poly->isClosed() && dist == poly->count() - 2U)
+			mid = (std::max(idx1, idx2) + 1) % poly->count();
```
dxli: a little bit confusion here for closed polylines:

```
0 - 1
|   |
3 - 2
```

given idx1=3, idx2=1, this algorithm gives mid = 0, seems to be perfect
but how could I handle idx1=1, idx2=3, and expected mid=1

```C++
if (idx2 < idx1) idx2 += poly->count();
mid = (idx1 + (idx2 - idx1)/2)%poly->count();
```

librecad/src/actions/rs_actionpolylinesegment.cpp
lordofbikes: A polyline, in terms of DXF guidelines, is a list of points (vertices X,Y). Each vertex has a bulge parameter which describes the arc to the next vertex. A straight line is the special case when the bulge value is 0. Thus the DXF format isn't capable of elliptical arcs in polylines or 'child'-polylines.

librecad/src/lib/information/rs_information.cpp
lordofbikes: In terms of DXF guidelines the vertices and so the indices must be arranged in sequence. I'm not sure about the LC entity container implementation for polylines, but it makes sense to keep the entities in sequence in the container too.

librecad/src/actions/rs_actionpolylinesegment.cpp
kfoltman: I think it can't be elliptical arcs because the arcs in polylines are represented with the single 'bulge' coordinate in the DXF format, not with real objects like in LibreCAD. I've ran into this problem before when trying to scale polyline with radiuses with different X and Y scale - the drawing fell apart, and the reason why was: it would turn the arcs into elliptical arcs, and that's not supported in polylines.

But, I think, this particular loop is over the objects in the whole drawing, not in a polyline. Previously, all polylines in the drawing were resolved, which caused unwanted effects: imagine the situation where the selected object is a polyline, but while iterating it's resolved into its components. So, the final list of segments would contain both the polyline as a whole object (added outside of this loop) and the components (added by this loop). That caused the resulting polyline to contain the same segments twice, hence the bug report.

This change means that the polylines will be treated as "opaque" objects with only start and end. This is okay because in LibreCAD one can only select a polyline in its entirety, not the lines/arcs that it's made of. So, the usable selected objects for this operation will be either lines, arcs or polylines.

Maybe there's a version of polyline object in DXF with separate x/y radius? I'm not very familiar with the format.

Hope the explanation makes sense - I'm not good at putting this in words.

librecad/src/lib/information/rs_information.cpp
kfoltman: I think the assumption of sequential order makes sense in case of open polylines - logically, you might treat them as sequences of points (and radiuses, in cases of arcs), even though they're represented as lines and arcs internally. It shouldn't be possible to have 'gaps' within a polyline - the next line's start should always be the preceding line's end.

So, the existing iterator should guarantee looping through neighbours. :)

Whether that assumption is held consistently within the project is another question, but at least I haven't found any visible bugs in that area during normal use.

librecad/src/lib/information/rs_information.cpp
kfoltman: Right, there are two solutions there, and my code always picks up a specific one, without allowing the user to pick the one they want.

Also, it's probably not a good UI anyway, because it would be quicker to just point to an arc directly instead of first and second entity. I can reimplement it that way if you prefer it. So, the fillet command would create a fillet if used on two lines, or modify/delete the fillet if used on an arc. What do you think?

librecad/src/actions/rs_actionpolylinesegment.cpp
kfoltman: @lordofbikes sorry, haven't seen your reply before writing mine - but looks like we're in agreement here.

librecad/src/actions/rs_actionmodifyround.cpp
kfoltman: I was just changing the existing code and trying to "blend in" style-wise with surrounding code :)

kfoltman:
Okay, so this set of changes requires some sanitization. I'll try to clean it up, fix the issues pointed out and submit another one.

PR #653, opened 2016-02-12
Added Backspace as shortcut for Delete. 

vlachoudis:
A tiny change to add a Backspace together with Delete as shortcuts for deleting.
Useful for mac keyboards (like on my linux) without a delete button.

Get the geometry from the activedMdiSubWindow before it is created was generating a coredump


r-a-v-a-s:
Thanks! :sunglasses:

PR #654, opened 2016-02-12
Edit-Copy Display snapping points while moving

vlachoudis:
Display the two cross hairs and show the snapping point while copy

### DIFF
--- a/librecad/src/actions/rs_actioneditcopy.cpp
+++ b/librecad/src/actions/rs_actioneditcopy.cpp
@@ -70,6 +70,11 @@ void RS_ActionEditCopy::trigger() {
     RS_DIALOGFACTORY->updateSelectionWidget(container->countSelected(),container->totalSelectedLength());
 }
 
+void RS_ActionEditCopy::mouseMoveEvent(QMouseEvent* e) {
+	if (getStatus()==SetReferencePoint)
+		RS_Vector mouse = snapPoint(e);
+}
+
 void RS_ActionEditCopy::mouseReleaseEvent(QMouseEvent* e) {
     if (e->button()==Qt::LeftButton) {
         RS_CoordinateEvent ce(snapPoint(e));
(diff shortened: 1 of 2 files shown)


librecad/src/actions/rs_actioneditcopy.cpp
```diff
@@ -70,6 +70,11 @@ void RS_ActionEditCopy::trigger() {
     RS_DIALOGFACTORY->updateSelectionWidget(container->countSelected(),container->totalSelectedLength());
 }
 
+void RS_ActionEditCopy::mouseMoveEvent(QMouseEvent* e) {
+	if (getStatus()==SetReferencePoint)
+		RS_Vector mouse = snapPoint(e);
+}
+
 void RS_ActionEditCopy::mouseReleaseEvent(QMouseEvent* e) {
```
dxli: thanks :+1:

librecad/src/actions/rs_actioneditcopy.h
```diff
@@ -7,7 +7,7 @@
 **
 **
 ** This file may be distributed and/or modified under the terms of the
-** GNU General Public License version 2 as published by the Free Software 
+** GNU General Public License version 2 as published by the Free Software
```
dxli: let's try to avoid empty patches

vlachoudis: Its my configuration on the vim editor. It strips all useless spaces (like at the end of line or before a tab). I have to find out how to revert the file :) Could you simply apply the first part only?

r-a-v-a-s: Note `virtual void mouseMoveEvent(QMouseEvent* e);`

Some git GUIs make it easy to revert the file 
https://git-scm.com/downloads/guis
or https://git-scm.com/book/en/v2/Git-Basics-Undoing-Things

dxli:
okay, I will do it by a patch

dxli:
@vlachoudis 

Added your patch by e3b5395

Please note I also added more C++11 style cleanup: "override" suffix

Looks like we will have to enable C++14 soon, to use std::make_unique(), for example

Thanks!

librecad/src/actions/rs_actioneditcopy.h
vlachoudis: Thanks I am new to git.

Issue #648, opened 2016-02-11
When moving vertexes on polyline with radiuses, all arcs radiuses are modified, even on unconnected arcs


kfoltman:
Steps to reproduce:
1. Draw a polyline with several segments
2. Use Modify/fillet to add a radius to one of the corners
3. Move any vertex of the polyline (preferably one not directly connected to the arc)
4. Watch the radius change

kfoltman:
Proposed fix: https://github.com/kfoltman/LibreCAD/commit/abb31fd1e17b2228d6c83fc3a38cea3942088c93

PR #655, opened 2016-02-13
(rebased) This doesn't contain the radius change/removal feature, because it needs some more thinking. However, I've added a fix for issue #648 instead.

kfoltman:


r-a-v-a-s:
I say we merge now and worry about tweaking later;
people are waiting for these fixes.

dxli:
@kfoltman 

Thanks a lot!

PR #672, opened 2016-03-12
Remove some conflicts between accelerator keys in menus + update translations

kfoltman:


r-a-v-a-s:
Looks fine to me...
@lordofbikes Can this be synced with the translation server?
I don't have an understanding of how that works.

lordofbikes:
The modification of the *.ts files will mess up the translation server. Please undo these changes, then it's ok for me too.

The translation servers git repo is synced by a script. The script syncs the pootle database, merge the changes to the local repo and then sync the repo with github. Then lupdate is run on the pootle server, to fetch latest strings. Finally the lupdate changes are committed and synced with the pootle database and github.
Modification of ts files outside of that script lead into merge conflicts because of the pootle sync scripts. That implies a lot of manual work to get it working again.

Armin

kfoltman:
OK, I've removed the second commit. Do you think it will not affect translations badly? I mean, the 'source' parts of the translations will be affected by my changes, which I thought may break the translations - but maybe there's some mechanism to compensate for that.

r-a-v-a-s:
I might not understand the system, but it seems like we need a way for developers to inject modified translation files. This is one use case -- another would be if a developer adds something new and wants to use a script to add initial translations -- another would be for fixing context 
(e.g. qc_applicationwindow -> lc_widgetfactory).

It seems like we would need a place to submit the files, and a full day dedicated to the injection.
On the dedicated day the pootle server would be disabled and the files pushed to this repository.
Then developers can modify those, and submit them somewhere for injection.

r-a-v-a-s:
> For those that use QT Linguist (download), feel free to download the .ts file, work on it, and upload it back to the server using the web interface. Please make sure to choose "merge" your translations instead of overwrite in case someone added a translation to a part of the program you didn't work on.

https://github.com/LibreCAD/LibreCAD/wiki/How-to-translate-LibreCAD

Does this work still?

lordofbikes:
I don't see any reason for these needs.

The system is very simple. In the source files add, remove or modify the strings to be translated in the `QObject::tr()` method. The Qt tool [lupdate](http://doc.qt.io/qt-5/linguist-manager.html) then scans all the source and header files to update the TS files listed in the `TRANSLATIONS` section of the *.pro files. It also takes care about context and to mark unused strings as obsolete in the new TS files.
In the build process the Qt tool [lrelease](http://doc.qt.io/qt-5/linguist-manager.html) is used to create the QM files for deployment.

As we have more than 50 language files for LibreCAD, a developer must take care of all the files to avoid inconsistency. This is very costly and error-prone by hand. And using lupdate outside the pootle scripts leads into the mentioned merge conflicts.

It's worthy of praise if a developer wants to commit initial translation for newly added strings, but he can do this for a single or maybe a couple of languages but surely not for all. And then the recommend way is to create an account on the [translation server](http://translate.librecad.org) and do the translation there.

Remains the update frequency between github and pootle. To do this on a daily basis will lead into extra work and to unnecessary work at times for translators. When strings change often, especially at the master branch for the moment, a translator will translate a sentence a couple of times until the final wording is found and possibly typos are fixed. One can rapidly loose his passion if this happens frequently.
We try to update the pootle server when new releases are announced and just before they are released to minimize unnecessary work for translators. We also do updates infrequently on requests in the [translation sub forum](http://forum.librecad.org/LibreCAD-translate-f5710645.html) if it makes sense.

Surely, there is an ample scope for optimisation here. We should find a better way for release management including feature fixes and a freezing points for translation issues. And a way to communicate this in time to the translators.

Concerning the way described in the wiki, I don't think that this still works. The article is from 2012, Qt4 and maybe some Qt3 dependencies. I'll check this and update the wiki if necessary. 
I'll also add readme files to the ts folders with some instructions.

kfoltman:
Well, in this case the change was to update the strings that have been changed in the source, in order to avoid making a whole lot of translations obsolete. I did it using a script to scan all the translations and do the changes, in order to be sure all the old translations are still used after the change. This problem may arise any time someone updates the accelerator keys or does some other changes that only affect source strings but should still preserve the translations.

I think being able to do that without messing up the translations saves a bit of unnecessary work for the translators.

lordofbikes:
I merged this manually because of a conflict in  librecad/src/ui/lc_widgetfactory.cpp

PR #674, opened 2016-03-16
use lrelease as determined by QT_SELECT and qtchooser

maqifrnswa:


maqifrnswa:
renamed branch, will resubmit

PR #686, opened 2016-03-21
dxfrw memory leak

dxli:
Raw pointer usage in dxfrw seems to be causing lots of memory leak

here's one attempt to avoid memory leak by switching to shared_ptr

please review, but do not merge this one:

* the fix might contain errors
* the fix is still incomplete


maqifrnswa:
i'm closing and reopening to run the static analyzer on this pull request. I just pushed a commit that will run the static analyzer on all pull requests but only update the analysis website after commits. Logs are available if you click on "show all checks"

dxli:
already pushed to master branch

PR #737, opened 2016-06-03
Committer: Equoda <Equoda@users.noreply.github.com>

Equoda:
fix: blocks moving/pasting bug

Solve the problem with copy-paste or move blocks bug cause some blocks disappears.
Most changes are in ```librecad/src/lib/modification/rs_modification.cpp```.
```RS_Modification::paste``` function has been rewritten.

Closes #715

PR #738, opened 2016-06-03
Committer: Equoda <Equoda@users.noreply.github.com>

### LINKED ISSUE #714: Unable to copy hatch area
Hatch doesn't copy by "Edit -> Copy | Edit -> Paste".

To reproduce:

1. Create draft
2. Create "Rectangle" shape
3. Select created shape
4. Do "Tools -> Hatch"
5. Select shape with hatch
6. "Edit -> Copy"
7. "Edit -> Paste"

Result: no hatch in pasted shape.

Librecad build info:

Version: 2.1.0-beta
Tag: 2.1.0-beta-8-g9723828
Compiler: GNU GCC 4.8.4
Compiled on: May 16 2016
Qt Version: 4.8.6
Boost Version: 1.54.0

Notes:

Draft is created from template. Here is: 
[draft-A4-book-mm.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/265915/draft-A4-book-mm.dxf.zip)

Equoda:
fix: moving/pasting hatch bug

Solve the problem with hatch moving/pasting cause hatch disappears.
This is a quick solution.
There are some minor issues to solve: hatch doesn't appear on the same layer as original after moving.

Closes #714

### DIFF
--- a/librecad/src/lib/engine/rs_blocklist.cpp
+++ b/librecad/src/lib/engine/rs_blocklist.cpp
@@ -212,7 +212,13 @@ void RS_BlockList::editBlock(RS_Block* block, const RS_Block& source) {
  * \p nullptr if no such block was found.
  */
 RS_Block* RS_BlockList::find(const QString& name) {
-    //RS_DEBUG->print("RS_BlockList::find");
+    try {
+        RS_DEBUG->print(RS_Debug::D_DEBUGGING, "RS_BlockList::find(): %s", name.toLatin1().constData());
+    }
+    catch(...) {
+        RS_DEBUG->print(RS_Debug::D_DEBUGGING, "RS_BlockList::find(): wrong name to find");
+        return nullptr;
+    }
 	// Todo : reduce this from O(N) to O(log(N)) complexity based on sorted list or hash
 	//DFS
 	std::vector<RS_BlockList const*> nodes;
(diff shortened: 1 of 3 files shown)


r-a-v-a-s:
What layer does the hatch appear on?

For an understanding of layer transfer, you might look at:
- `RS_ActionModifyAttributes::trigger()`
- `QG_DlgAttributes::updateData()`

You can add another commit to the branch, and it will automatically be part of this pull request.

r-a-v-a-s:
For any future commits please make the title / summary / message (whatever you want to call it) something descriptive.

We want to be able to look at the [commit list][1] and have some clue of what the commit is about.
Multiple commits titled `Committer: Equoda <Equoda@users.noreply.github.com>` doesn't help;
and the system already tags it with your name and links to your profile.

[1]:[https://github.com/LibreCAD/LibreCAD/commits/master]

lordofbikes:
To change the commit messages of all three pull requests, what is recommended, please see [github help - Changing a commit message](https://help.github.com/articles/changing-a-commit-message/).
The paragraph [Amending the message of the most recently pushed commit](https://help.github.com/articles/changing-a-commit-message/#amending-older-or-multiple-commit-messages) is what you have to do for your pull requests #737, #738 and #739.
As you work on your own fork, you can safely ignore the warning there about forced pushing. This is necessary in this case.

There*s no need for new pull requests, the changes will update the existing ones.

Equoda:
So, the first line of commit description is placed as header/summary in commits log? The template for github commit description therefore is: ```<changes>; #<track_number>\n<details>```?

OK, about wrong hatch layer. This error is due to ```RS_Hatch::update()``` function which doesn't preserve hatch sub-entities attributes. If it was intentionally or not, but new ```RS_EntityContainer* hatch``` object is initialized currently with ```nullptr``` layer.

Here is a patch. I will push it to my fork later: 
[librecad-hatch-moving.2016.06.06.diff.zip](https://github.com/LibreCAD/LibreCAD/files/300166/librecad-hatch-moving.2016.06.06.diff.zip)

r-a-v-a-s:
It wouldn't build for me because rs_modification.h didn't include QHash.
I guess it's a lesson in not trusting travis ci...
I guess I'll test in a temporary branch from now on...
although I didn't officially sign-up to be responsible for pull requests. >_<

I've noticed two issues already.

1. objects when selected only show the 'endpoints' and not the dashed lines
1a. this is strange... as I assume you would have noticed this

2. ctrl+c, then ctrl+v, then click = crash
2a. this isn't the proper procedure as it skips setting the reference point; 
however all crashes must be eliminated

r-a-v-a-s:
My mistake... the selection issue an issue on my end...

The copy / paste issue was not.

r-a-v-a-s:
Link to the issue in the commit message (as opposed to the pull request).

Also, I don't like the `Closes #xxx` feature.
It might close it... it might need to be reconsidered...
it's preferable to just manually close the issue in my opinion.

PR #768, opened 2016-06-18
Fix for Trackpads without smooth/hi-res scrolling

rubdos:
Trackpads that did not have hi-res scrolling fell back on the
default (mouse) behaviour. This makes behaviour inconsistent between
machines (or even DE's: Gnome X.org has hi-res, Gnome Wayland+XWayland
does not).

This patch assumes that if
  `options->device == "Trackpad"`
  and
  no smooth scrolling
the "scroll wheel" (`QWheelEvent::angleDelta()`) is used instead of
the "trackpad" (`QWheelEvent::pixelDelta()`).

Division by four is an arbitrary constant to make a compromise
between smooth scrolling and scrolling too fast.

This patch makes working with LibreCAD under XWayland the same
experience as under X.org afaik. More work is needed for native
Wayland (`librecad -platform wayland`) support.

Note on the first commit in this PR: it fixes wrong indentation of the altered code. Was too ugly to work with :)

### DIFF
--- a/librecad/src/ui/qg_graphicview.cpp
+++ b/librecad/src/ui/qg_graphicview.cpp
@@ -467,7 +467,6 @@ void QG_GraphicView::focusInEvent(QFocusEvent* e) {
     QWidget::focusInEvent(e);
 }
 
-
 /**
  * mouse wheel event. zooms in/out or scrolls when
  * shift or ctrl is pressed.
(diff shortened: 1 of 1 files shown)


rubdos:
@rvt : if someone ever mentions a regression w.r.t. trackpads, please mention me in the respective issue.

Issue #760, opened 2016-06-13
Pasted blocks in Block List widget do not contain proper hatch data


Equoda:
If retrieving block from "Block List" widget returned block object doesn't correlate with containing in the draft. This is a true for pasted blocks only.

Steps to reproduce:

1. Create a new draft
2. Create a shape ("Rectangle")
3. Create a hatch within 2. (select, "Hatch")
4. Create a block on 2. and 3. (select, "Create Block -> b1")
5. Create a new draft
6. Copy-paste created block from 4. to the new draft (pasted block name: "b1")
7. In Block List widget select "b1", choose "Edit Block"

Result: no hatch in the block opened for editting

An error occurs somehow while creating ```RS_BlockList*``` for "Block List" widget. There is proper hatch data in the original draft, but not when one call ```QC_DialogFactory::requestEditBlockWindow``` function: returned ```RS_Hatch``` object as a part of returned ```RS_Block``` contains ```NULL``` ```hatch``` property.

Equoda:
It's all about:

Version: 2.2.0-alpha-19-gd5b8b19
Compiler: GNU GCC 4.9.2
Compiled on: Jun 13 2016
Qt Version: 5.3.2
Boost Version: 1.55.0

On:  Debian 8 x86_64 GNU/Linux

Equoda:
Fixed by pull request #771

PR #770, opened 2016-06-19
[WIP] Fix all but one compile warning.

rubdos:
Most of them were misleading indentations, caused by mixed use of tabs
and spaces (or even caused by misleading amount of spaces).

Please have a look over here:
https://github.com/LibreCAD/LibreCAD/blob/master/librecad/src/lib/modification/rs_modification.cpp#L538

This thing causes

```
lib/modification/rs_modification.cpp: In member function ‘void RS_Modification::paste(const RS_PasteData&, RS_Graphic*)’:
lib/modification/rs_modification.cpp:539:14: warning: statement has no effect [-Wunused-value]
         data.asInsert;
         ~~~~~^~~~~~~~
```

Afaik, there is no such thing as a `nullptr` exception in C++, but perhaps I am wrong and there exists some compiler that throws them? Either case, having a reference that could be `nullptr`, is _very_ bad design.

I propose on entirely getting rid of that `try` block.

After fixing that warning, LibreCAD compiles without warnings on `-Wall`, GCC 6.1!

### DIFF
--- a/libraries/jwwlib/src/jwwdoc.cpp
+++ b/libraries/jwwlib/src/jwwdoc.cpp
@@ -1759,7 +1759,7 @@ CDataList JWWBlockList::GetBlockList(unsigned int i)
     for(unsigned int k=0; k < FBlockList.size(); k++)
         if(i == FBlockList[k]->m_n_Number)
             return *(PCDataList)FBlockList[k];
-	return {};
+    return {};
 }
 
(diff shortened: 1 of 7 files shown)


Equoda:
This warning is a my error. I will fix it along with #760

rubdos:
@Equoda: great, thanks.

Issue #727, opened 2016-05-19
librecad is unusably slow with much text


jirislaby:
When I open this dxf:
http://www.fi.muni.cz/~xslaby/sklad/test.dxf
librecad cannot be used reliably. It is way too slow when moving and operating. I assume it is because of the plenty of text there.

Note that I deleted most of the drawings. If I do the opposite: leave drawings and delete text, it works almost fine. I.e. it is only a bit slower.

jirislaby:
I am on openSUSE with librecad 2.0.10.

jirislaby:
I built it with -pg -mfentry and the profiler says (cropped to top 10):
Each sample counts as 0.01 seconds.
  %   cumulative   self              self     total           
 time   seconds   seconds    calls  ms/call  ms/call  name    
 12.54      0.38     0.38 253630493     0.00     0.00  RS_Entity::getLayer(bool) const
 10.37      0.69     0.31  5654719     0.00     0.00  RS_Entity::getPen(bool) const
  3.68      0.80     0.11  2411481     0.00     0.00  RS_Line::draw(RS_Painter*, RS_GraphicView*, double&)
  2.68      0.88     0.08 12087377     0.00     0.00  RS_Entity::RS_Entity(RS_EntityContainer*)
  2.51      0.95     0.08 88603338     0.00     0.00  RS_Entity::isUndone() const
  2.01      1.01     0.06  6495686     0.00     0.00  RS_GraphicView::drawEntityPlain(RS_Painter*, RS_Entity*, double&)
  2.01      1.07     0.06  2766696     0.00     0.00  RS_GraphicView::drawEntity(RS_Painter*, RS_Entity*, double&)
  1.84      1.13     0.06 19154983     0.00     0.00  RS_Entity::isVisible() const
  1.67      1.18     0.05  3369092     0.00     0.00  void std::vector<RS_Vector, std::allocator<RS_Vector> >::_M_emplace_back_aux<RS_Vector const&>(RS_Vector const&)
  1.67      1.23     0.05  2770510     0.00     0.00  RS_PainterQt::setPen(RS_Pen const&)

Whole profile at:
http://www.fi.muni.cz/~xslaby/sklad/profile.txt

r-a-v-a-s:
Does enabling draft mode help?

---

It seems to go from `RS_GraphicView::drawEntity` to `RS_GraphicView::drawEntityPlain`
but then I don't understand the call to `e->draw(painter, this, patternOffset);`
[...]

PR #772, opened 2016-06-20
Improving rendering performance w.r.t. #727

rubdos:
As stated in issue #727, using a lot of entities makes LibreCAD slow. I'm trying to figure out what things can get optimized. I already have one commit which makes it noticeably faster.

~~**Please do not merge yet, I'm going to find more optimizations.**~~

### DIFF
--- a/librecad/src/ui/qg_scrollbar.h
+++ b/librecad/src/ui/qg_scrollbar.h
@@ -39,15 +39,34 @@ class QG_ScrollBar: public QScrollBar {
 
 public:
     QG_ScrollBar(QWidget* parent=0)
-            : QScrollBar(parent) {}
+            : QScrollBar(parent) {
+        sizeHintCache = QScrollBar::sizeHint();
+    }
     QG_ScrollBar(Qt::Orientation orientation,
                  QWidget* parent=0)
-            : QScrollBar(orientation, parent) {}
+            : QScrollBar(orientation, parent) {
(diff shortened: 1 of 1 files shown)


rubdos:
Let's say that I cannot find more bottlenecks that are easy to tackle. There's still one that could use investigation: 6% of the drawing time of the dxf file in #727 goes to `RS_EntityContainers::clear()`, which is called by `RS_EntityContainer::~RS_EntityContainer()`.

This sits so deep that I do not dare to touch it though, so perhaps someone else could look at that. 

The EntityContainer's in question are those of `RS_Line::draw()`, which uses one to calculate some intersections before getting disposed. I assume that that part can be optimized easily, although someone could have a look at the mentioned clear() method too.

rubdos:
@r-a-v-a-s, why do you think my git email isn't set right?

```
git config --global -l
user.email=ruben.de.smet [at] rubdos.be
user.name=Ruben De Smet
```
(the [at] is against crappy spambots, but it's right in my config. Just replaced it here)

Iirc, github has that e-mail address registered. Perhaps not as primary, but they know about it :P

r-a-v-a-s:
See the '?' next to your name in the commit.

You have merged commits, yet you are not in the contributors list.

rubdos:
Oooh, okay, that's more logical now. GitHub knows about that e-mail address now. I recently bought that domain, but I only added it on GitLab, not here.

Issue #774, opened 2016-06-24
copied objects are not hidden with other layer objects


r-a-v-a-s:
1. create a circle
2. copy and paste
3. toggle layer visibility

The original circle is hidden, but the copy is not.

r-a-v-a-s:
It works for Move / Copy,
but not for Ctrl+C / Ctrl+V.

Equoda:
OK, there is a question.

While catching this bug some weakly related changes in code have been made, like: ```for (...;...;...)``` -> ```for (auto ... : ...)``` or additional ```if (!ptr)``` checks. Nothing critical, I think. Should it be a part of different pull request?

r-a-v-a-s:
You can include those changes.

Cantar4:
it seems that v2.2.1 still suffers from the same problem :
Ctrl+C/V now correctly transfer propperties but objects pasted that way are not hidden 
with the layer like the "move-copied' or the 'layer to layer' transfered ones are.

Equoda:
Pull request #777 merged

r-a-v-a-s:
@Cantar4 
There were no changes associated with this in 2.1.1
A full change log was included https://github.com/LibreCAD/LibreCAD/releases/tag/2.1.1

PR #777, opened 2016-06-24
fix: copied objects are not hidden with other layer objects; #774

Equoda:
Wrong logic for layers copying in ```RS_Modification::paste``` corected.

```RS_Polyline::setLayer``` function implemented to change sub-entities layers to the same value.

Some changes in code for better readability and stability


r-a-v-a-s:
Note my comment :point_right: https://github.com/LibreCAD/LibreCAD/commit/8886a959c0bd13bc114f6290d7c3fa89ce9db2a1

Equoda:
Oh, yes. I have read a post about using ```foreach``` but forgot. Sorry. Will make changes tomorrow.

Equoda:
Yes, this should be changed accordingly. There are many ```for (each)``` loops in the code.

I probably will be absent for a few days, so check it later.

PR #789, opened 2016-07-07
Fix compile with gcc 5.4

rezso:
lib/engine/rs_spline.cpp:592:2: error: 'iota' is not a member of 'std'
include numeric c++ header solves this.


lordofbikes:
Thanks for fixing this.

PR #800, opened 2016-08-05
Don't add /usr/include to INCLUDEPATH

maqifrnswa:
You get a build error with gcc 6 if /usr/include is in the INCLUDEPATH
See:
https://bugreports.qt.io/browse/QTBUG-53367
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=831189


r-a-v-a-s:
:sunglasses: :+1: 

Will this fix the daily ppa?
Was/is that an automated service?

Is it possible to update the stable ppa to 2.1.1? 

Thanks :-]

PR #848, opened 2016-11-30
Changing the globally declared array to non-global

coderaashir:
Any reference to dxfColors would need to be accessed through the class after this commit


r-a-v-a-s:
What is the benefit of this?

Always build your changes before submitting a pull request.
You broke the comment block.

coderaashir:
Sorry and I'll keep that in mind. I'm working on it

PR #851, opened 2016-12-24
Plugin Format & Translation Update

Xia-Weiwen:
1. Plugins
Use Qt standard plugin interface format now (Qt 5.7.0), according to this Qt official documentation: http://doc.qt.io/qt-5/qtwidgets-tools-plugandpaint-app-example.html
(1) All plugins use one single interface ID (iid)
(2) ```.pro``` file Modified for each plugin

2. Translation Update
Translations for Chinese Simplified are updated.

Thank you!


lordofbikes:
Xia-Weiwen, many thanks for your efforts!

I see two issues we have to check before merging.

1. Translation - please see https://github.com/LibreCAD/LibreCAD/blob/master/librecad/ts/README.md which is contained in both `ts` folders. 
As you have put much effort into this, I'll check if I can manually merge your changes to our translation server.

2. Qt 5.7.0 - on Linux, specifically the long term support distributions, don't have an up to date release packed. Thus we have lot of Linux users who compile LibreCAD themselves. On that systems Qt is not up to date too, so I have to check first what happens with older Qt. Maybe we have to add a preprocessor condition to use old code with Qt < 5.7 and your patched code on Qt >= 5.7

As the Christmas season starts today, this will take couple of days.

Xia-Weiwen:
@lordofbikes Thank you!

Xia-Weiwen:
New changes:
(1) @polarbear0330 added a method to draw regular polygons by selecting the center and the tangent point on its inscribed circle.
(2) Only load *.dll files when loading plugins on Windows.
(3) Translations updated for Chinese simplified.
Thank you!

Please test the plugins on MacOs and other platforms for changes in .pro files. I've tested them on Ubuntu 16.04 and Windows 7/10.

rvt:
Can we merge this?? If so, are there any implications?

Xia-Weiwen:
We have tested this ourselves on Ubuntu 16.04, Windows 7/10. If you merge this, please do some tests on MacOS and other platforms, and please pay attention to the translation files. I only did some translation for simplified Chinese, but .ts files for other languages may need updating as well.

lordofbikes:
I've manually merged your translation updates on the translation server and pulled in the changes. As pointed out in the [translation folders README.md](https://github.com/LibreCAD/LibreCAD/blob/master/librecad/ts/README.md) merging TS files on github will result in conflicts in the translation servers repo.

Thus the other changes have to be merged manually too. I schedule this for next week.

lordofbikes:
I have cherry-picked your commits and thus pulled all your changes into the main repo now.

I had to split your commit 3afd004 as this commit handled 3 issues. The translation part was previously merged on the translation server and the other two parts are separate commits now

As you have committed all into your master branch, you have to do a forced update of your fork now to sync with the main repo here again.

That's the reason why it is recommended to create separate feature branches, one for each feature. Thus one can keep the forks master branch synced with the main repo. 
After merging a feature branch of the fork by pull request the branch in the forked repo can be removed and the master branch syncs without conflicts then.

Many thanks again for your efforts and patience.

Xia-Weiwen:
Thank you!

PR #857, opened 2017-01-11
build-osx.sh: temporary dmg -> overwrite dmg & fix volname

rbuj:


lordofbikes:
I can't test this myself, but it reads plausible thus I'll merge it.

Many thanks for your efforts.

Issue #843, opened 2016-11-24
Issues with drawing circles


froebelina:
Hello everyone, 

I tried drawing circles and I observed unexpected behavior
### Expected behavior
1. I draw a circle with the Center, Point command, the command input should ask me to enter two points
2. I draw a circle with the Center, Radius command, the command input should ask me for a point and a length
### Observed behavior
1. The command input asks me to enter a point and a radius
2. The command input asks me for a point and afterwards draws the circle without asking me for a radius
### Steps to reproduce or sample file
See above
### Operating System and LibreCAD version info
Windows 7 Pro and same issues on Debian stretch, LibreCAD Version 2.1.3

lordofbikes:
Both tools work as designed. This is a matter of missing or imperfect documentation I think. 
Both issues are discussed in our forum already.

Concerning 1. I agree, that *Specify radius* is a misleading wording. But in fact, the point to specify is the radius of the created circle. One can enter the relative distance e.g. `@50,0` when asked for the radius, what creates a circle with a radius of 50, independent from the center point.
This becomes also clearer by drawing with the mouse, where the mouse click represents the input of a X,Y coordinate. After the center point is set, the circles radius is following the mouse movement.

Concerning 2. you already found out and stated in pr #860 that it works as designed. The problem is the input box for the radius in the tool options bar, which could be overseen or even maybe disabled by mistake. There are several other tools which use input boxes in the tool options bar.

This is not selling bugs as features :wink: 
These tools were not modified significant since QCAD was forked for LibreCAD. See the tool documentation of [QCAD 2.2](http://qcad.org/doc/qcad/2.2/reference/en/chapter19.html#1)

PR #860, opened 2017-01-21
Circle defined by center and point on circle now asks for center and …

froebelina:
…another point (instead of center and radius) fix for issue #843

### DIFF
--- a/librecad/src/actions/rs_actiondrawcircle.cpp
+++ b/librecad/src/actions/rs_actiondrawcircle.cpp
@@ -101,7 +101,7 @@ void RS_ActionDrawCircle::mouseMoveEvent(QMouseEvent* e) {
 		data->center = mouse;
         break;
 
-    case SetRadius:
+    case SetPoint:
 		if (data->center.valid) {
 			data->radius = data->center.distanceTo(mouse);
             deletePreview();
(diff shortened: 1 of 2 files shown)


FabianInostroza:
I also noticed this recently.

A google search pointed me to https://sourceforge.net/p/librecad/bugs/313/, but I could not understand why the problem was "solved", why "circle with center and point" was the right option.

froebelina:
Hi Fabian, thanks for providing the link. As I understand it the bug from the link is no bug. When you try to enter a circle defined by center and radius the command line does not ask for the radius. I thought that was a bug, too, but while I was fixing it I came across a textbox, that appears when you select the center, radius - circle and there you can enter a radius. So this is not a bug, it is just well hidden functionality ;)
But in the conversation you send me someone points out that the center, point-circle behaves as you would expect from the center,radius - circle, which seems to me as if he is selling a bug as a feature. A center, point circle should ask for a center and a point and not for a radius.

lordofbikes:
As I replied in Issue #843 I see no bug to fix.
I'll check modifying your wording:
`RS_DIALOGFACTORY->updateMouseWidget(tr("Specify point"), tr("Back"));`
or better 
`RS_DIALOGFACTORY->updateMouseWidget(tr("Specify point on circle"), tr("Back"));`
But this affects all translations and maybe well hidden documentation :wink:
Thus I'll sleep over it and keep the PR open to keep in sight.

Issue #884, opened 2017-05-08
Improve Current Drawing Preferences - Units tab


lordofbikes:
### Expected behavior
Changing the *Angle Format* combobox should update the *Angle Precision* combobox

### Observed behavior
Changing the *Angle Format* combobox modifies the *Length Precision* combobox.
The *Angle Precision* combobox keeps unchanged.

### Steps to reproduce or sample file
Open the Current Drawing Preferences dialog and select the *Units* tab.
Modify the *Angle Precision* and watch the *Length precision* combobox.

### Operating System and LibreCAD version info
Version: 2.2.0-alpha-179-g3d4adab
Compiler: GNU GCC 5.4.0
Compiled on: May  5 2017
Qt Version: 5.5.1
Boost Version: 1.58.0
System: Ubuntu 16.04.2 LTS

lordofbikes:
This is not related to #883 but the opportunity arises to handle them together.

PR #886, opened 2017-05-11
Fix for issue #884 -- Improve Current Drawing Preferences - Units tab

webmite:
updateAnglePrecision referenced wrong comboBox


lordofbikes:
Many thanks for fixing this.

Issue #883, opened 2017-05-07
Document Surveyor's Units in Current Drawing Preferences


elil:
### Expected behavior

Navigating through the interface Options-->Current Drawing Preferences-->unit(tab)-->Angle(section) and selecting "Surveyor's units" does not give an example or preview.  Like the other choices, an example is expected in the preview section.  

### Observed behavior

The preview is blank.  

### Steps to reproduce or sample file

Open program, then navigate to Options-->Current Drawing Preferences-->unit(tab)-->Angle(section) and selecting "Surveyor's units".  Watch the Angle preview while doing this and change it between a few things to see the presence and absence of the preview.  

### Operating System and LibreCAD version info

MacOS, 10.12.4
 
Version: 2.1.3
Compiler: Clang 8.0.0
Compiled on: Sep 22 2016
Qt Version: 5.6.1
Boost Version: 1.60.0
System: OS X 10.12

### Additional information

I looked [here](https://github.com/LibreCAD/LibreCAD/blob/3d4adab1081e683e34d7bd519f1d9fa5b8f5cd2b/librecad/src/lib/engine/rs.h#L511) and [here](https://github.com/LibreCAD/LibreCAD/blob/3d4adab1081e683e34d7bd519f1d9fa5b8f5cd2b/librecad/src/ui/forms/qg_dlgoptionsdrawing.cpp#L687) trying to figure it out and am still not quite sure.  

It might be quadrant bearing like N30d31'32"E (which is east or west bearing from north or south see also [#5 in wikipedia entry](https://en.wikipedia.org/wiki/Bearing_(navigation)#Bearing_measurement) or any search for 'convert azimuths and bearings' like [this](http://academic.brooklyn.cuny.edu/geology/leveson/core/linksa/comp.html)).  N30d31'32"E is pretty close to the number being used in the other examples, however, it is in the NE quadrant which makes it less obviously meaningful.  A better overall example might be in a different quadrant like (in Deg/min/sec) 149(0xB0) 28' 28" which in bearings would be S30d31'32"E.  

The other thing it could be (either instead of or in addition to a quadrant bearing) is that many surveying calculators and programs use a notation like 149.2828 which "obviously" (not obvious at all and very confusing) means 149(0xB0) 28' 28" and not 149.2828 or (in Deg/min/sec) 149(0xB0) 47'44"!

lordofbikes:
[...]

Issue #864, opened 2017-02-03
custom page size - incorrect unit handling


pacew:
### Expected behavior
Using Options -> Paper -> Custom, the desired Paper Width and Height should stick to the supplied values, and should appear in the PDF file when exporting

### Observed behavior
If you set Units to Inches and Paper Size to Custom, and enter, say 10 and 20 for the width and height. Then close the Drawing Preferences dialog, then re-open it, the values have changed to .3937 and .7874.
(these values are 10/25.4 and 20/25.4 - they have been scaled as if the previous numbers were millimeters).  Each time you re-open the dialog, the values have been scaled down again by another factor of 25.4.

### Steps to reproduce or sample file
Launch librecad so it shows a blank new drawing.  
Select Options -> Current Drawing Preferences
Select Units tab
Units -> Inches
Select Paper tab
Paper Format -> Custom
Paper Width -> 10
Paper Height -> 20
click OK
re-open dialog with Options -> Current Drawing Preferences
observe that Paper Width is now .3937 and Paper Height is .7874
close and re-open the dialog and observe that the Paper Width and Height are even smaller.

Also, if you export a PDF file, then do something like view it with evince and examine the properties, the page size has been scaled down improperly.

### Operating System and LibreCAD version info
Ubuntu 16.04
LibreCAD compiled from git commit 3ab980c of 2017-01-10 but also observed in librecad 2.0.9 from standard Ubuntu repository

After experimenting this afternoon, I believe the attached patch file is on the right track to fix the problem, but I have never looked at librecad before and I don't know if there are other parts of the program that may need similar attention.  It does make the size parameters sticky, and they do flow through to the exported PDF.

Thanks for a great tool!

[custom-units-fix.txt](https://github.com/LibreCAD/LibreCAD/files/751738/custom-units-fix.txt)

lordofbikes:
The fix contributed by @webmite works fine for me.
If there are any further issues, please feel free to reopen it.

PR #887, opened 2017-05-13
fix for Issue #864, #883 -- custom page size - incorrect unit handling

webmite:
This gets the unit handling corrected for custom paper size.
It appears to generate correct dimensional pages for 'File | Export | pdf'
but it does not complete printing if using something like pdf creator if
just doing a 'File | Print'.


webmite:
Having found that I have commits for 2 distinct issues and realizing this is likely to cause you grief I am going to close off this pull request and segregate the 2 commits into separate branches.

PR #888, opened 2017-05-15
fix for Issue #864 -- custom page size - incorrect unit handling

webmite:
This fix allows for the custom page size values to be entered and stick after clicking 'OK' and returning
to the drawing options page.  It also correctly swaps width and height when toggling from landscape to portrait.  
Selecting 'File | Export | Export as Pdf' will create a document with the correct dimensions.
Selecting 'File | Print' and selecting a tool like pdfcreator does not generate a document.
It has not been tested on a physical printer either.


webmite:
I am postponing this commit until I clean up the branches in my fork.

webmite:
cleaned up my fork and separated revisions into separate branches. Now OK to share the changes.

lordofbikes:
Many thanks for that fix @webmite!

PR #889, opened 2017-05-18
fix for issue #883 -- Document Surveyor's Units in Current Drawing Pr…

webmite:
…eferences

AngleFormat now has case specifier for Surveyors units.


webmite:
This update assumes that the stored angle is relative to 0 degrees as North and a counter-clockwise angle rotation being the positive direction.  If this is not the case I will have to introduce a translation function at the start of the code.

lordofbikes:
Thanks for this fix @webmite!

Issue #853, opened 2016-12-29
Feature Request:  Provide Mapping Angular Layout


DocInvent:
### Ubuntu 16.04 with 
LibreCAD
Version: master
SCM Revision: 2.0.9
Compiler: GNU GCC 5.3.1
Qt Version: 5.5.1
Compiled on: Mar 24 2016

ngnmnt:
As above, this would be highly beneficial. 0 degrees UP, clockwise angles. Similar to a compass.
Hopefully easy to implement. Offset the base angle 90 degrees and change the +/- sign.

Essential to people working with charts, maps, GIS systems. This is the only reason I don't use LibreCAD so much. It would be great to see this as a new feature.

PR #890, opened 2017-05-19
Basic Handling of ANGBASE and ANGDIR DXF variables. 

webmite:
Required to provide fix for issue #883 and is a requirement for pull #889.
These variables were already being read and written at the filesystem level.
Coordinates widget on main page
adjusted to compensate for new base and angle direction. Access to
view and modify within drawing is provided by entries on the
'Units' Tab of the drawing preferences dialog, under the
'Angle' area of the page. 

It allows user to specify a new 'base' direction for 0 degrees and
to specify which direction the positive degrees will proceed in...
clockwise or counter-clockwise. 

Note: It is NOT complete. It has to have appropriate modifications applied
to all 'angle' related creation/preview/modification code.
probably should be done against a separate development branch until
it is consistent.

Comments and suggestions are welcome.

### DIFF
--- a/librecad/src/lib/engine/rs.h
+++ b/librecad/src/lib/engine/rs.h
@@ -512,6 +512,11 @@ public:
         Surveyors
     };
 
+    enum AngleDirection {
+        CounterClockwise,
+        Clockwise
+    };
+
     /**
      * Enum of levels of resolving when iterating through an entity tree.
      */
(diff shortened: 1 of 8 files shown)


webmite:
Also fulfills feature request in issue #853

webmite:
I am going to take this offline until I can make a more complete submission

Issue #813, opened 2016-09-08
Text generation falgs not interpreted correctly (mirrored text)


NjamNjamNjam:
Dear LibreCAD Team,

I noticed that LibreCAD does not correctly interpret mirrored text.

See specification:
[http://www.autodesk.com/techpubs/autocad/acad2000/dxf/text_dxf_06.htm](http://www.autodesk.com/techpubs/autocad/acad2000/dxf/text_dxf_06.htm)

71
Text generation flags (optional, default = 0):
2 = Text is backward (mirrored in X).
4 = Text is upside down (mirrored in Y).

I attached an example DXF file that contains text that has the text generation flag "2" and screenshots that show how LibreCAD interprets the text and for comparison, how another toll interprets the text.

Here is the example DXF file:
[Example_DXF.zip](https://github.com/LibreCAD/LibreCAD/files/461438/Example_DXF.zip)

This is how the DXF file looks in LibreCAD:
![example_project_bot_librecad](https://cloud.githubusercontent.com/assets/22073992/18347147/40191600-75c4-11e6-9dd5-76356987864c.PNG)

This is how the DXF file looks in another CAD tool:
![example_project_bot_othertool](https://cloud.githubusercontent.com/assets/22073992/18347150/432a6c2c-75c4-11e6-958c-42696e696507.PNG)

jhansonxi:
Also encountering this with LibreCAD 2.0.9 on Xubuntu 16.04 x86_64.

PR #893, opened 2017-05-25
fix for issue #813 -- Text generation flags handling

webmite:
This code change maintains the type-safe standard of RS_TextData by adding an additional
type of MirroredXnY equivalent to the Backward and UpsideDown flags
both being set.  Changed the type of None to MirroredNone to avoid
confusion.  Added checkboxes to the Text Dialog to allow viewing/setting
these flags at time of creation or modification.

Fix for issue #813.

Added some additional test files to test each Mirrored flag in a different orientation.
first file is original test file from issue #813 .
[Example_Project_Bot.zip](https://github.com/LibreCAD/LibreCAD/files/1029087/Example_Project_Bot.zip)
[Example_Project_Bot-1.zip](https://github.com/LibreCAD/LibreCAD/files/1029088/Example_Project_Bot-1.zip)
[Example_Project_Bot-2.zip](https://github.com/LibreCAD/LibreCAD/files/1029089/Example_Project_Bot-2.zip)

### DIFF
--- a/libraries/jwwlib/src/dl_entities.h
+++ b/libraries/jwwlib/src/dl_entities.h
@@ -708,7 +708,7 @@ struct DL_TextData {
     double height;
     /*! Relative X scale factor. */
     double xScaleFactor;
-    /*! 0 = default, 2 = Backwards, 4 = Upside down */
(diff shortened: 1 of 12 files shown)


webmite:
I have identified that mirror and rotation behavior have to be corrected before this can be merged..so I am taking it offline until I can resolve.

Issue #872, opened 2017-03-01
cannot assign Light Grey to a layer using Edit Layer Attributes


andykmv:
### Expected behavior
can assign Light Grey to a layer using Edit Layer Attributes

### Observed behavior
in my drawing, i noticed that when trying to assign Light Grey to a layer using Edit Layer Attributes, i can go through the motions of assigning Light Grey - but the change does not stick.   

from dellus:
Confirmed. 
Light grey seems not to exist. Even if you draw something directly as light grey (not via layer property) or try to change to it as general property. A bug!

### Steps to reproduce or sample file
to test this was not an issue in my drawing, i created a new drawing (no layers to start with).  i created a test 1 layer and a test 2 layer. 

if i edit layer attributes select test 1 and change the layer colour to GREEN i see the change in the layer list straight away, and, if i edit the layer attributes for layer test 1 i also see the layer colour is green. editing test 2 layer colour is the same. i did the same test with 6 or so diff colours no probs. 

BUT when i edit a layer's attributes to change the colour, and if i select light grey, the colour does not change either in the layer list or in edit attributes list. 

### Operating System and LibreCAD version info
I am using Windows 7 pro, with LC 2.1.3

PR #895, opened 2017-05-29
fix for issue #872 -- Allow selection of last color in selection list

webmite:
This code now allows for selection of last color in the list
for any layer.
Meant to resolve issue #872 .


lordofbikes:
May thanks tor this fix and sorry for delay!

Issue #824, opened 2016-10-02
Angular dimension can't do outer angles


DiGitHubCap:
When I use the angular dimension tool, I can only dimension inner angles, not outer ones.

What currently happens:
![2016-10-02-181839_848x813_scrot](https://cloud.githubusercontent.com/assets/905317/19024235/1e17e7d6-88cd-11e6-8548-c32737716f02.png)
What I need:
![2016-10-02-181931_814x748_scrot](https://cloud.githubusercontent.com/assets/905317/19024243/285dbf2c-88cd-11e6-8833-d97152290582.png)

PR #896, opened 2017-06-23
adjusted correctAngle functions to be more consistent. Fix for #824 request.

webmite:
Added fmod call in correctAngle to ensure 360 degree results wrap back to 0 degrees.
Added comments to document expected output range of functions.
Added correctAngle2 function with output range of -PI to +PI.
Adjusted Angular dimension code to provide inside/outside angles
rather than inside/supplementary angles. 
Solves issue #824 requested feature of being able to represent outside angles.


lordofbikes:
Many thanks for this contribution.

Issue #903, opened 2017-08-11
Program not responding


migial:
[Boden-1150-01.zip](https://github.com/LibreCAD/LibreCAD/files/1218514/Boden-1150-01.zip)

### Expected behavior
zooming per mouse wheel in the right region of the drawing with activated construction layer for helplines
### Observed behavior
few zoom events per mouse wheel will be executed and after that program not responding
### Steps to reproduce or sample file
Sample file is uploaded

### Operating System and LibreCAD version info
Windows 7 professional, 
LibreCAD Version: 2.1.3
Compiler: GNU GCC 4.9.1
Compiled on: Sep 23 2016
Qt Version: 5.4.1
Boost Version: 1.60.0
System: Windows 7

Issue #891, opened 2017-05-21
LibreCAD freezes opening a dxf previously saved


luislfc:
### Expected behavior
It should open the file normally, show all lines and not freeze while we "navigate" the file.

### Observed behavior
LibreCAD freezes and the CPU jumps to 100%. It will never recover from this state

### Steps to reproduce or sample file
1- Open the sample dxf file,
2- LibreCAD will not show all lines but it still is not frozen
3- select "view -> auto zoom" and it will freeze

There are other ways to replicate, like zooming out with the mouse, but the easiest way to replicate is the described one.

### Operating System and LibreCAD version info
Linux version 4.8.0-1-amd64
CPU: Intel(R) Core(TM) i7-3630QM CPU @ 2.40GHz
Version: 2.1.2
Compiler: GNU GCC 6.3.0
Compiled on: Sep 17 2016
Qt Version: 5.7.1
Boost Version: 1.62.0
System: Debian GNU/Linux 9 (stretch)

### Notes
The dxf file was zipped as the issue tracker doesn't accept dxf files to be uploaded.

[Transistors.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/1017101/Transistors.dxf.zip)

metrologymaster:
I have experienced the same issue while using the Windows version running under both Win7 and Win10.

Version: 2.1.3
Compiler: GNU GCC 4.9.1
Compiled on: Sep 23 2016
Qt Version: 5.4.1
Boost Version: 1.60.0
System: Windows 7

gratefulfrog:
I have the same problem on Ubuntu Linux 64it when zooming on a previously created dxf.

Version: 2.2.0-alpha
Compiler: GNU GCC 5.4.0
Compiled on: May 31 2017
Qt Version: 5.5.1
Boost Version: 1.58.0
System: Ubuntu 16.04.2 LTS

dxli:
The file Transistor.dxf  contains a polyline bulge of positive inf bulge.

dxli:
The infinity(saved as -1.633123935319537e+16, line 2716 of the file Transistor.dxf)
 bulge is found for the polyline bulge field (code=42).

The definition of bulge is $\tan(\frac{\alpha}{4}$, where $\alpha$ is the angular size of the arc. Conceputally, zero angular size is equivalent to $2 \pi$, but a bulge size of $\pm\infinity$ cannot be used to construct an arc.

[...]

Issue #868, opened 2017-02-15
Infinite loop while manipulating construction lines (quadratic form)


damienflament:
### Expected behavior
Select the line.

### Observed behavior
 * Infinite loop (100% CPU usage),
 * unusable UI (need to kill the _librecad_ process.

### Steps to reproduce or sample file
 * Create a new file,
 * make a vertical line at position (0,0),
 * set the current layer as a _construction_ layer,
 * pan the view to the up-right to hide the previously constructed line,
   - **note**: after this step, a line appears starting at the upper-right corner of the viewport and ending
      near the lower-left corner.
 * try to select all objects (_Ctrl-A_).

**N.B:** With a previous LibreCAD version (maybe 6 month ago), panning the view containing construction lines always triggers the same issue. Now (with current stable or the Git version), it seems to be more unpredictable. The steps described above definitely "work".

### Operating System and LibreCAD version info
 * ArchLinux `_uname -a`: 
   > Linux _<hostname>_ 4.9.8-1-ARCH SMP PREEMPT Mon Feb 6 12:59:40 CET 2017 x86_64 GNU/Linux
 * LibreCAD:
   - version 2.2.0 alpha (from current ArchLinux User Repository package _librecad-git_),
   - compiled for x86_64 architecture.

### Debug log
Logs created using command `librecad --debug 6 &>librecad.txt`: [librecad.txt](https://github.com/LibreCAD/LibreCAD/files/776303/librecad.txt)

LibreCAD seems to loop on computing _quadratic forms_. But the logs are not populated anymore when it starts infinite looping.

r-a-v-a-s:
Confirmed... odd behavior,

For anyone interested in solving this,
I suggest looking at 
`RS_GraphicView::drawEntity(RS_Painter *painter, RS_Entity* e, double& patternOffset)`
and
`RS_Line::draw(RS_Painter* painter, RS_GraphicView* view, double& patternOffset)`
and
`RS_Selection::selectAll(bool select)`

jamesruan:
confirmed.
I lost half an hour's work of this odd line, go through the top-right to bottom-left.

fa201:
I cannot reproduce this with
Version: 2.2.0.2
Compiler: GNU GCC 9.4.0
Compiled on: Jul 29 2023
Qt Version: 5.12.9
Boost Version: 1.75.0
System: Ubuntu 20.04.6 LTS

PR #915, opened 2017-10-24
Check viewport intersection only through points along the line

Akranen:
I've noticed the code to draw the line was creating an intersection on a point outside the line and the viewport boundaries and that was resulting in an almost infinite loop when trying to draw the pattern. I just changed the parameter on getIntersection to consider only points on the entities.

### DIFF
--- a/librecad/src/lib/engine/rs_line.cpp
+++ b/librecad/src/lib/engine/rs_line.cpp
@@ -575,7 +575,9 @@ void RS_Line::draw(RS_Painter* painter, RS_GraphicView* view, double& patternOff
 
     if ((endPoints[0] - getStartpoint()).squared() >
             (endPoints[1] - getStartpoint()).squared() )
+    {
         std::swap(endPoints[0],endPoints[1]);
+    }
 
 	RS_Vector pStart{view->toGui(endPoints.at(0))};
 	RS_Vector pEnd{view->toGui(endPoints.at(1))};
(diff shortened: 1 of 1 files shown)


Akranen:
Also, that should solve problems from issues 868, 891 and 903

lordofbikes:
Many thanks for your efforts!

* I couldn't reproduce #868 first, it works for me with and without your fix. After finding the line style thing with the #903 file, LibreCAD freezes when I used a dashed construction line.
* With the file from #891 LibreCAD freezes with and without your fix, seems to be another problem. Also there seems to be no construction line involved.
* The file from #903 worked for me without your fix when I changed the construction layer line style to continuous.

With these observations I'm not sure if the intersection point computing is the real cause for infinite looping. I need more time to understand the issues and your fix.

Maybe you could give some more details from challenging this. 
What are your conclusions and how did you come to this solution?

Akranen:
I just assumed the three were the same problem. But for 903 it takes two lines that are parallel within RS_Tolerance, however getIntersection is not considering that tolerance and is returning a huge point that is not on either lines.

It works for continuous lines because the infinite loop happens at the end of the function, when trying to draw the pattern.

line 685 of the original file: for (int j=0; total<length; j=(j+1)%pat->num)

In my case length was 1e49 

Does that make sense?

lordofbikes:
Still haven't found a better solution, so I merge it now and we will see if there are any drawbacks.

Many thanks for your efforts!

PR #916, opened 2017-10-27
Luiscoloradourcola@gmail.com gears

mojadita:
On my humble opinion, this patch has improved gear drawing.  At present, the modulus is used instead of the pitch radius, as for a gear set the modulus has to be the same for all the gears, so I think it is better to conserve moduli and make the program to calculate the gear overal dimension.

I plan to eliminate the gear type, as the inner/outer mechanism only affects the addendum dedendum values (which are now configurable, as a factor of the modulus ---again--- to make them easily eligible)

I also plan to include a layer selection field to put the auxiliary entities (pitch circle, base circle, addendum  and root circle) there.


lordofbikes:
Is this still work in progress or is it ready for review?

mojadita:
I'll have it ready probably during this weekend.... It can be merged, as it works, but the patch will finally bring some cleaning and the interference calculation lacks of a proper routine to detect where both curves hit each other, so it can be probed, but the interference part makes some undesirable glitches in the graph.  I prefer you to wait to next week to see the final result.

plugins/gear/gear.h
```diff
@@ -15,11 +15,11 @@
 
 #include "qc_plugininterface.h"
 #include <QDialog>
-
-class QPointF;
-class QSpinBox;
-class QDoubleSpinBox;
-class QComboBox;
```
lordofbikes: The previous code is preferred here, use forward class declaration instead of includes for pointer variables.
The correct place for includes of member pointer classes is the cpp file (where they are already).

Please restore the previous `class Q...;` lines and add `class QCheckBox;` for the new member pointer variables.

plugins/gear/gear.h
```diff
@@ -57,12 +56,25 @@ public slots:
...
 
     QPointF *center;
+    QDoubleSpinBox *rotateBox;
     QSpinBox *nteethBox;
-    QDoubleSpinBox *pitchBox;
+    QDoubleSpinBox *modulusBox;
     QDoubleSpinBox *pressureBox;
     QDoubleSpinBox *addendumBox;
     QDoubleSpinBox *dedendumBox;
-    QComboBox *typeBox;
```
lordofbikes: With the use of C++11 features, it's easy to initialize pointers inside class declaration.
For new code we'd like to see this:
```    QSpinBox  *n1Box {nullptr};```
Please use this initialization for all your new pointer members and add it to existing pointers too.

plugins/gear/gear.h
mojadita: I used that because I had problems compiling the moc files, but I'll try to put them as they were in the previous code.

mojadita:
Now I think the patch can be merged to master.  Please review and make any comments.
Thank you in advance, is finally you merge this patch :)

mojadita:
Now everything should be ok.

lordofbikes:
Many thanks for your great efforts!

mojadita:
You are welcome!!!  It was delicious to receive such a great welcome to the project. I expect to be able to collaborate more in this software project.

Issue #479, opened 2015-01-18
Scrolling with mouse doesn't sync with scrollbars.


baranovskiykonstantin:
Scrollbars works fine, but when I scroll with mouse roller of the scrollbars doesn't moves. So, for example, when I scroll screen to top with mouse, roller of the scrollbars stay on this position (middle by default) and if I will try to scroll screen back using scrollbars - I can't.

LibreCAD:
Version: master
SCM Revision: 2.0.7
Compiler: GNU GCC 4.9.2
Qt Version: 5.4.0
Compiled on: Jan 18 2015

OS:
ArchLinux
Linux 3.17.7-1-ck x86_64

Installed from AUR 
https://aur.archlinux.org/packages/librecad/

r-a-v-a-s:
If the drawing area is infinite, then there are no boundaries for the scrollbar to sync with.
What about a fixed size mode, where the user can declare the size of the drawing area?

baranovskiykonstantin:
Paper size is set to A4.

dxli:
currently, if you scroll by the scroll bars, the area is bound to the visible area plus buffer with a fixed size. Therefore, scrolling is limited here.

when scrolling with mouse/keyboard, scrolling is unbound, making the sliders of the scroll bars always centered.

Proposal:

make mouse/keyboard scrolling behavior identical to scroll bars: limited with some boundary.

1. The current unlimited scrolling is not useful to me, because I mostly scroll too much, and get lost in a few steps;
2. If a user does want to scroll beyond the limit, zooming out would do it. Just like on google maps, you do want to zoom out first, if panning by large distance is needed;
3. we do want to implement a "viewAt" method to allow centering the drawing area around some coordinates.

dxli:
wheel scrolling is limited to scroll bar ranges

When zoomed in, the scrolling range won't cover the whole drawing

need to see whether this limitation is acceptable.

if not, need to revert this commit

atorrey:
It sounds to me like this is sort of related to the bug I filed a few days ago about the scroll bars being much smaller than the drawing area...  IMHO the scroll bar ranges should be about the same as the drawing size you'd get if you did a Zoom Auto to get the drawing all on the screen, or maybe a SMALL amount bigger...  Thus the scroll bars would change the percent of the range that they occupied in proportion to how much one is zoomed in / out, and the range would always allow scrolling to any part of the drawing.

[...]

PR #919, opened 2017-11-10
Fixes

xzcvczx:
Just a couple of fixes that i have found necessary for using and building LibreCAD on macos.

The issue with trackpad scroll not working with scrollbars disabled was from #479 where the scroll was controlled by the scrollbars so they were realigned correctly


lordofbikes:
Many thanks!

PR #922, opened 2017-11-12
fix a couple more issues on macOS

xzcvczx:
the PlistBuddy line requires the type, misunderstood the issue as it was working sometimes, however the one here works on the command line all the time

The postprocess-osx script was very specific and so now it will use the information gained from qmake to find the lrelease executable, the choice of qmakes to use is able to be done as well so will work for non-path qmake/lrelease.

Added the generated dmg file to .gitignore just in case

sorry about the noise, closed the last pull request and redid it as this as it got ugly

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -196,6 +196,7 @@ QC_ApplicationWindow::QC_ApplicationWindow()
     setCentralWidget(central);
 
     mdiAreaCAD = central->getMdiArea();
+    mdiAreaCAD->setDocumentMode(true);
 
     settings.beginGroup("Startup");
     if (settings.value("TabMode", 0).toBool())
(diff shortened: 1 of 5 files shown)


librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -196,6 +196,7 @@ QC_ApplicationWindow::QC_ApplicationWindow()
     setCentralWidget(central);
 
     mdiAreaCAD = central->getMdiArea();
+    mdiAreaCAD->setDocumentMode(true);
```
lordofbikes: This seems to do nothing in Windows and Linux, hope it has some value in OSX

librecad/src/ui/qg_layerwidget.cpp
```diff
@@ -246,7 +246,8 @@ QG_LayerWidget::QG_LayerWidget(QG_ActionHandler* ah, QWidget* parent,
     // lineEdit to filter layer list with RegEx
     matchLayerName = new QLineEdit(this);
     matchLayerName->setReadOnly(false);
-    //matchLayerName->setText("*");
+    matchLayerName->setPlaceholderText("Filter");
+    matchLayerName->setClearButtonEnabled(true);
```
lordofbikes: much better than before, thanks

lordofbikes:
many thanks :+1:

PR #924, opened 2017-11-14
fixes #877

### LINKED ISSUE #877: "open with" LibreCAD doesn't seem to work on a mac. Opens an unnamed document.
### Expected behavior
From the finder opening a DXF or DWG should open the file directly.

### Observed behavior
LibreCAD opens with an unnamed document.

Any DXF or DWG will do.

macosx 10.11 El Capitan, LIbreCAD 2.1.3

xzcvczx:
- Adds required information for open with to Info.plist
- Uses a QApplication subclass to catch the file open events prior to the event filter being in place
- Creates an Event filter in the main window for QApplication for fileopen events that are received

### DIFF
--- a/librecad/src/main/lc_application.cpp
+++ b/librecad/src/main/lc_application.cpp
@@ -1,0 +1,24 @@
+#include <QStringList>
+#include <QFileOpenEvent>
+
+#include "lc_application.h"
+
+LC_Application::LC_Application(int &argc, char **argv)
+    : QApplication(argc, argv)
+{
+    QStringList files = QStringList();
+}
+
+// This is only used until the event filter is in place in mainwindow
+bool LC_Application::event(QEvent *event)
+{
+    if (event->type() == QEvent::FileOpen) {
+        QFileOpenEvent *openEvent = static_cast<QFileOpenEvent *>(event);
+        files.append(openEvent->file());
+    }
+    return QApplication::event(event);
+}
+
+QStringList const& LC_Application::fileList() const {
+    return files;
+}
--- a/librecad/src/main/main.cpp
+++ b/librecad/src/main/main.cpp
@@ -31,14 +31,15 @@
 #include <QSplashScreen>
 #include <QSettings>
 #include <QMessageBox>
+#include <QFileInfo>
 
 #include "rs_fontlist.h"
 #include "rs_patternlist.h"
 #include "rs_settings.h"
 #include "rs_system.h"
-#include "rs_fileio.h"
 #include "qg_dlginitial.h"
 
+#include "lc_application.h"
 #include "qc_applicationwindow.h"
 #include "rs_debug.h"
 
@@ -51,7 +52,7 @@ int main(int argc, char** argv)
 
     RS_DEBUG->setLevel(RS_Debug::D_WARNING);
 
-    QApplication app(argc, argv);
+    LC_Application app(argc, argv);
     QCoreApplication::setOrganizationName("LibreCAD");
     QCoreApplication::setApplicationName("LibreCAD");
     QCoreApplication::setApplicationVersion(XSTR(LC_VERSION));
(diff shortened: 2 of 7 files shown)


lordofbikes:
From code review this looks a bit suspect to me, I'll add some comments to clarify what I mean.
I can't test it now, try to reactivate my Mac Mini to give it a try.
Maybe the changes to `librecad/src/Info.plist.app` are sufficient to fix #877?

xzcvczx:
while the changes to Info.plist do allow the files to be opened with the application, the application still needs to be able to handle them, both on initial load of the application (if double clicking the file opens the application) and if the application are already open

librecad/src/main/lc_application.cpp
```diff
@@ -0,0 +1,24 @@
+#include <QStringList>
+#include <QFileOpenEvent>
+
+#include "lc_application.h"
+
+LC_Application::LC_Application(int &argc, char **argv)
+    : QApplication(argc, argv)
+{
+    QStringList files = QStringList();
```
lordofbikes: This is a local instance named `files`, this is not the `LC_Application::files` member variable!
So this line is useless.
`QStringList `'s constructor is called on creation of LC_Application, so initializing is not needed here.

librecad/src/main/main.cpp
```diff
@@ -294,6 +296,8 @@ int main(int argc, char** argv)
     setlocale(LC_NUMERIC, "C");
 
     RS_DEBUG->print("main: loading files..");
+    // get the file list from LC_Application
+    fileList << app.fileList();
```
lordofbikes: This is the only place where `fileList()` is called.
At this time `LC_Application` is instanced and initialized, but the event loop is not running and so `LC_Application::files` is still empty.

librecad/src/main/lc_application.cpp
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// This is only used until the event filter is in place in mainwindow
+bool LC_Application::event(QEvent *event)
+{
+    if (event->type() == QEvent::FileOpen) {
```
lordofbikes: This event is OSX specific, it is not fired on Windows or Linux, so I couldn't test this yet.
Even when `files` is populated here, `filesList()` is never called after the event loop starts.
Seems to be a WOM (write only memory) :wink:
I haven't got yet, what OSX want to signal with `QEvent::FileOpen`. On Windows and Linux, when I double click a DXF file, a new instance of LibreCAD opens with this file, as the filename is pushed into `main()` and handled by `handleArgs()` there.

xzcvczx: anything that is opened by double clicking the file in finder unfortunately doesn't get pushed into main but does get sent as a fileopenevent, unfortunately one can't make it wait until the eventfilter is installed, as such the LC_Application is needed to catch any early events and then main.cpp uses that to send to the window.

any subsequent fileopenevents are handled by the eventfilter

lordofbikes:
Don't be cross with me, that I disputed your enhancements :flushed:
It's still hard to believe for me, feels too complicated, but thats really how it works on OS X, well done!

I've added copyright header to your new files and did some style cosmetics. Also added `#ifdef Q_OS_MAC` to extract parts from non OS X builds and mark them as OS specific.

You can apply my changes to your branch and push them to your github repo, then I can merge it.
`git apply <`[lcapp.patch.txt](https://github.com/LibreCAD/LibreCAD/files/1674628/lcapp.patch.txt)

I added your copyright notice from git log info, if you don't want this, feel free to remove or change it at your pleasure. And please, before you commit the changes, check if my changes build well on your system.

lordofbikes:
Any issues with my patch? Shall I merge and patch it myself?

lordofbikes:
Many thanks for your efforts!

PR #928, opened 2017-11-24
replaced-auto_ptr-with-unique_ptr

webmite:
direct swap of unique_ptr (supported) for auto_ptr (deprecated)


lordofbikes:
Many thanks for fixing this.

PR #934, opened 2017-12-04
fix-new-layer-auto-numbering for issue #935

webmite:
This allows for any base layer name with a suffix number to be auto
incremented. If no name is supplied it reverts to the default base layer
name of 'noname' and with any additional layers being provided a suffix
number automatically. It is also possible to just use numbered layers
with no base layer name and it will still auto increment correctly...the
only exception being layer '0'. So if you want simple numbered layers
click the + button then replace the default 'noname' layer name in the
dialog with '1' and click 'OK'. From then on highlighting the layer you
want to increment and clicking the '+' button will give you a new layer
with the next available number above the one you just selected.

### DIFF
--- a/librecad/src/ui/qg_dialogfactory.cpp
+++ b/librecad/src/ui/qg_dialogfactory.cpp
@@ -187,9 +187,25 @@ RS_Layer* QG_DialogFactory::requestNewLayerDialog(RS_LayerList* layerList) {
             layer_name = "noname";
         }
         newLayerName = QString(layer_name);
-		int i = 2;
+        QString sBaseLayerName = layer_name;
+        int ndx = newLayerName.length();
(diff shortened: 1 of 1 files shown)


webmite:
Just to give you a feel for the difference this behaviour produces look at the below table...
<table>
<row>
<td>old behaviour</td>
<td>
screenshot from 2017-12-04 18-07-37
</td>
</row>
</table>
<table>
<row>
<td>new behaviour</td>
<td>
screenshot from 2017-12-04 17-20-10
</td>
</row>
</table>

lordofbikes:
Well done, many thanks!

A smarter solution would have been the use of `QRegularExpression`, maybe you give it a try sometime :wink:

PR #947, opened 2018-01-28
updated-AddLayer-routine-to-use-QRegularExpression

webmite:
This routine uses QRegularExpression in the AddLayer routine to identify the base layer name and numeric component and also extends the existing routine by allowing zero padded numbers to be correctly incremented

### DIFF
--- a/librecad/src/ui/qg_dialogfactory.cpp
+++ b/librecad/src/ui/qg_dialogfactory.cpp
@@ -32,6 +32,7 @@
 #include <QString>
 #include <QFileDialog>
 #include <QToolBar>
+#include <QRegularExpression>
 
 #include "rs_patternlist.h"
 #include "rs_settings.h"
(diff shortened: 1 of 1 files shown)


lordofbikes:
My comment on #934 wasn't meant as an urgent request :wink:
Many thanks for challenging this.

I modified the regex, as it fails on some extraordinary layer names like _Floor 17/Room 45.02/Electrical 1.05_, that means, when one or more numbers are included in the middle of the base name.
And I did some C++11 style cosmetic.
Then I added copyright notice from your git log info, if you don't want this, feel free to delete or modify it at your pleasure.

You can apply and test my changes and then commit and push to your branch, so I can merge it.
`git apply <`[qregex.patch.txt](https://github.com/LibreCAD/LibreCAD/files/1675204/qregex.patch.txt)

Also, there seems to be a bug in `QRegularExpression`, as the regex failed as `^(.*\D+)|(\d*)$` but matched as `^(.*\D+|)(\d*)$`. It shouldn't matter where the `|` is, it looks more correct in the first version, but it doesn't match :confused:

webmite:
Thanks for the suggestion...that seems to have done the trick...I wanted a regex that would handle the example you gave but did not come up with the solution.  I have found the site 

https://regex101.com/

to be useful in verifying the behaviour of a regex pattern prior to testing in the application.

I will try to get the patch applied and re-submitted over the next few days.

lordofbikes:
Well done, many thanks!

PR #952, opened 2018-03-09
Fix VS error C2600 by removing ambiguity in the default constructor.

kkouptsov:
When compiling the code using VS2017, I get C2600 errors:

error C2600: 'RS_BlockList::RS_BlockList': cannot define a compiler-generated special member function (must be declared in the class first)
error C2600: 'RS_Entity::RS_Entity': cannot define a compiler-generated special member function (must be declared in the class first)

which eventually occur due to ambiguity of the default constructors (what should RS_BlockList() do?)

This change deletes unnecessary definitions and eliminates the error.


lordofbikes:
Many thanks for your efforts!

PR #958, opened 2018-03-17
Angular dimensions now respect user's custom label text

pwnage101:
dimangular objects should reference getLabel instead of getMeasuredLabel
in order to respect any optionally provided custom label text.

This bug was introduced in #896, and no explanation for the change was given in the commit message, so I assume it was just left over from testing the PR.


lordofbikes:
Many thanks for your efforts!

PR #964, opened 2018-03-27
Create meulière.dxf

Lamy78:
I propose a new hatching patern ....


lordofbikes:
Many thanks for sharing your pattern.

The filename is used in the pattern drop down box, thus I recommend to rename the file and give it an appropriate english name.
This will also avoid possible issues with non UTF file systems.

Lamy78:
Hello, 
So could you rename it "millstone.dxf" ? 
(and show it as "millstone" ?)
Thanks !
F

PR #967, opened 2018-03-30
Better inside-horizontal dimensions

pwnage101:
I split this into two commits because I didn't want a bunch of style fixes to be mixed in with the actual code changes for the reviewer to parse.  Reviewer, it's best to review each commit individually.

The commit message should explain most of the changes:

    Improve appearance of inside-horizontal dimensions

    * The monolithic method to calculate both inside-horizontal and
      aligned dimensions were broken into two separate methods since they
      became too different.  As a result, some calculations are duplicated,
      so there is still room for improvement.

    * Now, dimension text is never offset from the insertion point for
      inside-horizontal mode.  This would otherwise be a breaking change,
      but inside-horizontal was broken to begin with, so I would rather take
      this oppportunity to fix what I believe to be a bug.

    * getAlignText() was renamed to getInsideHorizontalText() because the
      method returns true when inside-horizontal mode is enabled, not when
      aligned mode is enabled.

    * A new helper function was added for calculating intersections between
      dimensionLine and text boxes.  This de-dupes some code.

    * A new method was added to RS_Line (getProjectionValueAlongLine) to get
      the projected value of a point along the line.  This is only used
      currently for sorting vector solutions of intersections.

Here is a preview of inside-horizontal dimensions BEFORE this PR:

test_librecad_dimensions_before-2

and AFTER:

test_librecad_dimensions_after-2

I feel that most of the changes are fairly self explanatory, so I will not go into detail.  However, please feel free to ask questions.

Regarding the change to text insertion point, here is an example of the offset insertion point BEFORE this commit:

test_librecad_dimensions_before-2_middle_of_text

and AFTER:

test_librecad_dimensions_after-2_middle_of_text

Note that the blue dot (reference point) for the dimension text is in the center of the text.  This will always be true after this commit, for inside-horizontal dimensions.

Finally, a comprehensive screencast demonstrating all the angles:

out

### DIFF
--- a/librecad/src/lib/engine/rs_dimension.cpp
+++ b/librecad/src/lib/engine/rs_dimension.cpp
@@ -24,6 +24,8 @@
 **
 **********************************************************************/
 #include<iostream>
+#include<cmath>
+#include<string>
 #include "rs_information.h"
 #include "rs_line.h"
 #include "rs_dimension.h"
@@ -31,6 +33,7 @@
 #include "rs_units.h"
 #include "rs_math.h"
 #include "rs_filterdxfrw.h" //for int <-> rs_color conversion
+#include "rs_debug.h"
 
 RS_DimensionData::RS_DimensionData():
 	definitionPoint(false),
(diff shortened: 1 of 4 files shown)


pwnage101:
See also my forum post announcing this change: http://forum.librecad.org/Inside-horizontal-dimensions-td5715827.html

lordofbikes:
Great improvement and perfect documented pull request!
Many thanks for your efforts.

PR #975, opened 2018-04-19
fixed length extension lines should scale according to the General Scale

pwnage101:
This was reported by dellus on the forum:

http://forum.librecad.org/Inside-horizontal-dimensions-tp5715827p5715872.html

I'm not *certain* what the correct behavior should be, but this seems
right to me (scaling with General Scale).

Standard general scale of 1:
2018-04-19-184110_526x457_scrot

After setting general scale to 2:
2018-04-19-184204_531x452_scrot


lordofbikes:
Good job again, many thanks for fixing this!

I found no clear reference that fixed extensions must be scaled with $DIMSCALE but also no opposing view. Dellus and you are right, it makes more sense this way.
I tested a LibreCAD DXF for reference with DWG TrueView and got same results, so no reason to decline your work.

Issue #930, opened 2017-11-27
LibreCAD source code compile problem


qium06:
My OS is win7. I use Qt Creator 4.4.1(based on Qt5.9.3). When compiling, I met with the following error.

\LibreCAD\librecad\src\lib\math\rs_math.cpp:305: error: no matching function for call to 'mu::Parser::DefineConst(const char [3], double)' 
         p.DefineConst("pi",M_PI); 
                                ^ 

\LibreCAD\librecad\src\lib\math\rs_math.cpp:306: error: no matching function for call to 'mu::Parser::SetExpr(std::__cxx11::string)' 
         p.SetExpr(expr_copy.toStdString()); 
                                          ^ 

\LibreCAD\librecad\src\lib\math\rs_math.cpp:312: error: no match for 'operator<<' (operand types are 'std::ostream {aka std::basic_ostream<char>}' and 'const string_type {aka const std::__cxx11::basic_string<wchar_t>}') 
         std::cout << e.GetMsg() << std::endl;

qium06:
Has anyone successfully compile these source code before? It seems to be the problem of type conversion. The compiler can not automatically transfer type from "const char *" to "std::wstring" or from  "std::string" to "std::wstring" or from "std::wstring" to type suitable for "cout".

qium06:
These problems involve two cpp files. After I do changes to a few lines in these two files. The compilation is successful. I think there must be better ways to solve the problem between "std::string" and "std::wstring".

In the file  \LibreCAD\librecad\src\lib\math\rs_math.cpp:
rs_math.cpp
Line:305
p.DefineConst(L"pi",M_PI); //   <--  p.DefineConst("pi",M_PI);
Line:306
p.SetExpr(expr_copy.toStdWString());  //   <--p.SetExpr(expr_copy.toStdString())
Line:312
std::cout << QString::fromStdWString(e.GetMsg()).toStdString() << std::endl;  // <-- std::cout << e.GetMsg() << std::endl;

In the file   \libreCAD\plugins\plotequation\plot.cpp:
Line:60~64
            p.DefineConst(L"pi",M_PI);                   //<-- p.DefineConst("pi",M_PI);
            p.DefineConst(L"e",M_E);                     //<-- p.DefineConst("e",M_E);
            p.DefineVar(L"x", &equationVariable);        //<-- p.DefineVar("x", &equationVariable);
[...]

PR #977, opened 2018-04-23
SVG export changes/updates

ayaromenok:
- renamed to Export as CAM/plain SVG;
 - MakerCAM option description, which originally was implemented as QLabel, replaced with ToolTips - to make it more similar to other UI/save space in dialog;
 - all options by default is the same, as was before;
 - added option to set line width (it was hardcoded to 1mm, so default value is the same)
 - added export of image and corresponding option to dialog (absolute path used)
 - added baking of lines of all supported types (dot, dash, dashdot, etc) to used it in CAM's

related forum thread - http://forum.librecad.org/SVG-Import-Export-td5715881.html

### DIFF
--- a/librecad/src/actions/lc_actionfileexportmakercam.cpp
+++ b/librecad/src/actions/lc_actionfileexportmakercam.cpp
@@ -3,6 +3,7 @@
 ** This file is part of the LibreCAD project, a 2D CAD program
 **
 ** Copyright (C) 2014 Christian Luginbühl (dinkel@pimprecords.com)
+** Copyright (C) 2018 Andrey Yaromenok (ayaromenok@gmail.com)
 **
 **
 ** This program is free software; you can redistribute it and/or modify
@@ -36,7 +37,7 @@
 
 LC_ActionFileExportMakerCam::LC_ActionFileExportMakerCam(RS_EntityContainer& container,
                                                          RS_GraphicView& graphicView)
-    : RS_ActionInterface("Export as MakerCAM SVG...", container, graphicView) {}
+    : RS_ActionInterface("Export as CAM/plain SVG...", container, graphicView) {}
 
 
 void LC_ActionFileExportMakerCam::init(int status) {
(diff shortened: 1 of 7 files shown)


lordofbikes:
Thanks for your efforts!
But it will take a couple of days to find a free slot for review.

lordofbikes:
LibreCAD crashes on Windows with a runtime error in QXmlStream, on export with an image.
This may depend on platform, I have to do more debugging.

ayaromenok:
Thank you. I will try to build windows version soon to fix it

ayaromenok:
can you give a exact configuration of Windows build and path to image? 
(it's looks like a string conversion on certain configuration(compiler\qt version\etc) and\or path to file)

ayaromenok:
sorry for delay with Win test - I can confirm, that this is Windows only issue, related to usage ':' in path (like "C:\my.img") and it's my fault, that I didn't test patch on Windows :( 

In fact, a ASSERT: "!name.contains(QLatin1Char(':'))" came from from Qt itself: qtbase\tree\src\corelib\xml\qxmlstream.cpp, line 3448 - for some unclear for me reason it's strictly forbidden to use ':' in attributes in QXmlStream, so any correct XML/SVG with standard windows path will crash. (and tools like Inkscape generate\required that path).  

I will evaluate a possibility to workaround this QXmlStream issue. If it's not possible without significal changes, I will remove ImageExprort in SVG - so it will exist as a *.diff file for people, who use Unix\OSX only SVG pipeline to manually build LibreCad.

Also I had a #930 issue during compilation with QtSDK\mingw530_32bit\Qt 5.9.5 on Win10\64bit

lordofbikes:
The ASSERT is about the attribute name `xlink:href`, not the path containing a colon.
```
void LC_MakerCamSVG::writeImage(RS_Image* image)
{
    ....
    xmlWriter->addAttribute("xlink:href", image->getData().file.toStdString());
    ....
}
````
I found correct usage in `void LC_MakerCamSVG::writeInsert(RS_Insert* insert);`

So this modification works for me:
```
xmlWriter->addAttribute("href", image->getData().file.toStdString(), NAMESPACE_URI_XLINK);
```
`NAMESPACE_URI_XLINK` was already defined because of its usage in `writeInsert(()`

Now I wonder how this worked on other OS, I haven't tried yet.

**Comment by lordofbikes** [comment]
No problem, you're welcome.

I had a look at Qt source code to learn this lesson, there are no OS defines explaining different working on varying systems. So maybe this is related to different encoding on Linux/Windows.
Anyhow, when it works this way everything is fine. 
Then I will test it on Mac OS too, to be sure.

**Comment by lordofbikes** [comment]
Concerning the #930 issue, what was your solution to solve this?

**Comment by ayaromenok** [comment]
solution from #930 issue comment N2 works for me:
```
p.DefineConst(L"pi",M_PI); // <-- p.DefineConst("pi",M_PI);
p.SetExpr(expr_copy.toStdWString()); // <--p.SetExpr(expr_copy.toStdString())
```

ayaromenok:
it's a issue to build LibreCad\master with Qt 5.11rc(which had some code internal structure changes) on Linux\Windows - few includes missing (QAction & QActionGroup). I will make it as separate issue\pull request.

lordofbikes:
Many thanks for this great improvement and sorry for merge delay.

PR #986, opened 2018-06-05
Fix warnings made by gcc 7

Daniel-Trevitz:
gcc7 adds fall-through as default check. Pull request fixes most of the errors (not the ones in the included libraries)


lordofbikes:
Many thanks for fixing this :+1:

Daniel-Trevitz:
Easy peasy :)

PR #990, opened 2018-06-11
fix build with Qt 5.11

jirislaby:
The new Qt removed some implicit inclusions of headers. To avoid build
errors, add explicit includes of those we use in the sources.


lordofbikes:
Many thanks for fixing this :+1:

PR #991, opened 2018-06-21
C++11-ify code base

Daniel-Trevitz:
4 patches, 1 particularly large.
Removes NULL and 0 from setting pointers.
Use a single approach for singletons - the same approach as in qtcreator.
Reformat line indents of rs_actionpolylinetrim.cpp

### DIFF
--- a/librecad/src/ui/forms/qg_commandwidget.cpp
+++ b/librecad/src/ui/forms/qg_commandwidget.cpp
@@ -27,6 +27,7 @@
 
 #include <algorithm>
 
+#include <QAction>
 #include <QKeyEvent>
 #include <QFileDialog>
 #include <QSettings>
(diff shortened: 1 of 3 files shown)


Daniel-Trevitz:
I'm not sure why there are3 commits from others included. Did I mess up my git tree?

ArmanHayots:
The main point of nullptr is not to manually replace all zero's and NULL's with it, but just replace one macros
`#define NULL 0 //or (void*)0`
with another
`#define NULL nullptr`

That's how it must be done, I think.

Daniel-Trevitz:
#defing' NULL is not a recomeneded approach: https://stackoverflow.com/questions/9003425/is-it-safe-to-define-null-nullptr

ArmanHayots:
Follow all recommendations, patterns&practices and we'll see where it will bring you. 

> That said, if NULL was correctly used as a null pointer constant, then nothing should break.

It means that good code will work and bad code will not. No need to dig a hole to walk around it.

Daniel-Trevitz:
Overwriting system definitions breaks portability while increasing build system complexity. It should not be done.
Feel free to fork and send pull requests with alternative solutions.

PR #995, opened 2018-07-11
Cpp modernize

Daniel-Trevitz:
Begin effort of modernizing the code base by using nullptr.

### DIFF
--- a/librecad/src/actions/lc_actionfileexportmakercam.cpp
+++ b/librecad/src/actions/lc_actionfileexportmakercam.cpp
@@ -50,7 +50,7 @@ void LC_ActionFileExportMakerCam::trigger() {
 
 	RS_DEBUG->print("LC_ActionFileExportMakerCam::trigger()");
 
-    if (graphic != NULL) {
+    if (graphic != nullptr) {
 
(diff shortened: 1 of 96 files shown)


Daniel-Trevitz:
This is exactly the same as before, but this time I have fixed my git tree... that took a while... Now I know to use branches since that makes pull requests cleaner.

lordofbikes:
This looks better than #991!

But there is still an issue with your forked repo. The `master` branch still contains the 4 commits from your `cpp-modernize` branch and is ahead of `origin master`. You should `reset --hard` your master branch to commit ac524dbd17008f08e6995e6fe8b27094dd321d15, which is the latest commit at `origin master` and then again forced push your `master` branch.

I already viewed #991 couple times, but I'm still not sure how to handle this PR. There are some issues, too much for now, we should discuss before you put more efforts in C++11-ifying the code base.
Maybe you want to join one of our chat channels, [IRC #librecad](https://webchat.freenode.net/) or [Zulip](https://librecad.zulipchat.com/), what you prefer. If you don't have any preferences, go with Zulip, this is more comfortable.

PR #996, opened 2018-07-12
Fix copy/paste/undo

Daniel-Trevitz:
Fix simple copy/paste/undo by removing deleted entities from the undo cycle.
This probably introduces glitches in complicated undo chains,
but this is preferable to a segfault.


lordofbikes:
Many thanks and sorry for the delay

PR #1018, opened 2018-10-09
Lock/unlock all layers

xanderdin:
My LibreCAD projects usually contain dozens of layers. It's a pain to lock/unlock them one by one. This code adds corresponding action and makes it possible to lock/unlock them all at once, just like the hide/show layers code does.


lordofbikes:
Many thanks for your efforts

PR #1023, opened 2018-10-19
Console dxf2pdf

xanderdin:
Making librecad run in console and print a bunch of dxf files to pdf file(s).

`./unix/librecad dxf2pdf [options] <dxf_files>`

**Examples:**

  `./unix/librecad dxf2pdf *.dxf`    -- print all dxf files to pdf files with the same names.

 ` ./unix/librecad dxf2pdf -o some.pdf *.dxf`    -- print all dxf files to 'some.pdf' file.

**Options:**
  **-h, --help**                  Displays this help.
  **-v, --version**               Displays version information.
  **-a, --fit**                   Auto fit and center drawing to page.
  **-c, --center**                Auto center drawing on page.
  **-k, --grayscale**             Print grayscale.
  **-m, --monochrome**            Print monochrome (black/white).
  **-p, --paper <WxH>**           Paper size (Width x Height) in mm.
  **-r, --resolution <integer>**  Output resolution (DPI).
  **-s, --scale <double>**        Output scale. E.g.: 0.01 (for 1:100 scale).
  **-o, --outfile <file>**        Output PDF file.
  **-t, --directory <path>**      Target output directory.

**Arguments:**
  **<dxf_files>**                 Input DXF file(s)

On Linux we can also create a link to librecad executable named **dxf2pdf** and run like this:
`dxf2pdf [options] <dxf_files>`


xanderdin:
No, it's not related to that topic. My projects usually contain 50-60 lists of architectural drawings. It was pretty time consuming and dumb to open it one by one and print from LibreCAD GUI. After some searches I found dwg2pdf of QCAD project, but it is not available without a pro license. So I decided to try and write a tool that can automate my boring stuff. After a week of my tryings that tool now exists and do what I want. And why not to share it with the project which is the core dependency for that tool?

Initially I was trying not to touch LibreCAD code at all and do a standalone tool. But dependencies are huge and I failed. After that I was trying to compile that tool using all LibreCAD code and got two binaries - librecad and dxf2pdf which is actually yet another instance of librecad plus my code. Considered that stupid, and finally decided to do a little injection of my code in the beginning of LibreCAD's main.cpp, making it just a [command] on librecad's arguments list.

I'm not well neither in C++ not in Qt, and barely understand much of LibreCAD code yet. But I hope there is not too much bad code in my pull request as far as this code just works as I need it. Hope it can be useful and for others who require similar functionality.

lordofbikes:
That's great! This is what keeps the open source world alive.
Your efforts are very welcome and hopefully inspire others to follow your lead.

lordofbikes:
Many thanks again for sharing this!
This functionality was asked for several times here and in the forum.

There is some room for improvement, e.g. on Windows is no output at all.
Windows GUI applications have no access to command windows, even if the executable was started from there. But basically the conversion itself works.

VictorLamoine:
I'm interested in doing the same for DXF > SVG, I need a command line tool that converts DXF files to SVG files.

There is no `QPrinter::SvgFormat` [enum](http://doc.qt.io/qt-5/qprinter.html#OutputFormat-enum). How are files exported to SVG in LibreCAD?

lordofbikes:
The classes for SVG export are located in files at https://github.com/LibreCAD/LibreCAD/tree/master/librecad/src/lib/generators

xanderdin:
Just a notice to the community.

It's turned out that just running LibreCAD in console mode does not make it a "true" console application. It still requires a running X server. For most use cases it is probably OK, but sometimes it is inconvenient, for example, if you want to call dxf2pdf on a remote server or create a docker container or automate a pipeline, etc. For those advanced use cases I created a set of patches that builds upon LibreCAD code a standalone version of dxf2pdf that does not require a running X server (X libraries are still required though).

https://github.com/xanderdin/LibreCAD_dxf2pdf_standalone

disarticulate:
was this tested under windows, or is this a unix only thing? I'm getting an opening aborted message after librecad opens.

xanderdin:
As @lordofbikes mentioned, there are some issues on Windows. I don't have Windows, but I managed to test it once on Windows 10 and yes, there were some issues with console messages output, but resulting PDF files were OK. As far as I remember, also failed run it in cmd.exe and PowerShell because it was constantly crashing, no matter was it LibreCAD.exe with dxf2pdf command or not. But in **Bash** shell shipped with **Git For Windows** PDF files were generated successfully. It was a few months ago and I don't remember more details already.

PR #1028, opened 2018-11-01
Fix: RowSpacing and ColSpacing values were swapped in Insert dialog

xanderdin:
In the 'Insert' properties dialog the Row Spacing field value is actually displaying and saving column spacing, and the Column Spacing field value is actually displaying and saving row spacing. Fixing that.


lordofbikes:
Nice catch. Many thanks for fixing this.

PR #1029, opened 2018-11-02
Fix: Factor value was ignored on block import

xanderdin:
Factor value was ignored when importing block from file or library. Fixing that issue.

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -489,7 +489,7 @@ void RS_Modification::paste(const RS_PasteData& data, RS_Graphic* source) {
     }
 
     // adjust scaling factor for units conversion in case of clipboard paste
-    double factor = 1.0;
+    double factor = data.factor;
     // scale factor as vector
     RS_Vector vfactor = RS_Vector(factor, factor);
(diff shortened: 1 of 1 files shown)


lordofbikes:
Many thanks for this fix.
It's straightforward and seems logical.

But because of the comment in the line above the fix I hesitate to merge it now.
The comment implies that there possibly was a copy/cut/paste issue.
The paste() method is used for clipboard actions too.
So we have to be sure that the fix doesn't make other things worse again.
I haven't used this before, so I can't say how it worked before.

We have to check a couple of test cases before.
- import mm block into inch drawing and vice versa
- cut/copy entities in mm drawing and paste to inch drawing and vice versa
- all this between drawings with same units mm and inch too

Have you already checked some of these cases?
Can you support us here and report results?

xanderdin:
There is an 'if' statement a couple of lines below that reassigns the 'factor' and 'vfactor' anyway for the clipboard actions. So I believe my fix shouldn't harm.  But I'll check the proposed test cases and report.

xanderdin:
Edge cases! I forgot about "Don't trust user input" principle. When user inputs 0 for the factor, the behavior is a little bit wired. I'll fix it.

PR #1030, opened 2018-11-03
Possible fix: import block to active layer

xanderdin:
There is an issue for me, when importing a block from file or library it is almost always inserted to layer 0 no matter what layer is selected. For example, in LibreCAD 2.1.3 that wasn't the case. In this pull request I try to fix that issue. But there is a block of code which tries to 'select the same layer in graphic as in source' (I just commented it out). I don't really understand why it is needed. Please feel free to reject this pull request if it breaks anything vital.

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -509,6 +509,9 @@ void RS_Modification::paste(const RS_PasteData& data, RS_Graphic* source) {
     // default insertion point for container
     RS_Vector ip = data.insertionPoint;
 
+    // remember active layer before inserting absent layers
+    RS_Layer *l = graphic->getActiveLayer();
+
     // insert absent layers from source to graphic
     if (!pasteLayers(source)) {
         RS_DEBUG->print(RS_Debug::D_ERROR, "RS_Modification::paste: unable to copy due to absence of needed layers");
@@ -516,6 +519,7 @@ void RS_Modification::paste(const RS_PasteData& data, RS_Graphic* source) {
     }
 
     // select the same layer in graphic as in source
+    /*
     auto a_layer = source->getActiveLayer();
     if (!a_layer)
     {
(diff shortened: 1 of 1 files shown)


lordofbikes:
So you detected changed behavior in latest code, that's good to know, any thanks.

I've checked diffs between 2.1.3 and master and found a bunch of modifications.
So let me check first how it worked in 2.1.3 and what has changed.
The goal must be, if it wasn't absolutely worse in 2.1.3, to keep functionality. Otherwise we will trigger  a support request wave with next release.

Concerning your fix, is this how it works for you or is it the 2.1.3 behavior?

If the fix is different, we can consider a checkbox in import tool options widget or add an option in application preferences. So the next release can behave like 2.1.3, but gives the chance to influence behavior for anyone who need it.

xanderdin:
My fix is not the exact 2.1.3 behavior, just partially restores it to its bare minimum convenient level. For example, in 2.1.3 when you try to insert a block into active layer whose color is red, the preview also turns red (not always though, and I don't know on what exactly it is dependent). My fix doesn't affect preview color at all, just lets me insert a block to the selected active layer without further need of doing additional actions such as changing properties of every insertion made.

I also believe restoring of 2.1.3 functionality is preferred. Possibly, checkbox is also an option. And this my fix should be rather considered as a quick temporary fix, both drawing attention to the issue and allowing me do my work with the bleeding edge version of LibreCAD. Probably I should had to file a bug report fist, but sometimes for me it's easier and more interesting to delve into code myself than find right words to correctly describe the issue. Sorry for that and thank you for all your hard work you're doing for the project.

lordofbikes:
There's no reason to feel sorry, everything is fine.
We appreciate your efforts and it doesn't matter if you choose the pull request or issue way.
Especially when you already have a branch for battling an issue, pull request is fine.
Important is, that you report the issue and we are aware of it.
So we keep this open to track it until it's solved finally somehow.

lordofbikes:
Many thanks, sorry for the delay

PR #1031, opened 2018-11-03
Fix: Factor value was ignored on block import

xanderdin:
This is reopening of #1029 pull request with check for user inputs 0 as factor value. Substituting 0 with 1.0 for preventing empty blocks inserts.

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -489,7 +489,7 @@ void RS_Modification::paste(const RS_PasteData& data, RS_Graphic* source) {
     }
 
     // adjust scaling factor for units conversion in case of clipboard paste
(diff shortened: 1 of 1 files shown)


xanderdin:
After suggested modification when user inputs values less than zero the preview is set according to factor but mirrored. But after mouse click block is inserted with factor 1.0 as intended by the change. When we were comparing data.factor to 0 the insertion was according to factor and mirrored, which in my opinion is more consistent with the preview.

lordofbikes:
I was aware of the mirror effect, but assumed that negative values are invalid.
With my investigations I come to realize, that negative values are quite valid.
So what do you think about:
`double factor = (RS_TOLERANCE < fabs( data.factor)) ? data.factor : 1.0;`

lordofbikes:
Many thanks for your efforts! :+1: 
And sorry for the extra work.

PR #1032, opened 2018-11-05
Fixed 2 segfaults causes

xanderdin:
When all MDI windows are closed there are a bunch of document-dependent actions available through **File** menu (which obviously should be disabled, but they are not). Most of those actions don't cause any segfaults. But two of them do:

1. **File->Import->Block**
2. **File->Export->Export as CAM/plain SVG**

The first is pretty easy to fix by checking for MDI window in `QC_ApplicationWindow::slotImportBlock()` as most of the similar code does.

The second is trickier and is rooted down to `QC_ActionHandler::setCurrentAction()`. Although inside that method there is a check for if _view_ is NULL or if _document_ is NULL, neither _view_ nor _document_ is set to NULL after closing document window.

Fix for the second segfault cause probably should be thoroughly tested for any side effects, though I didn't encountered any during my tests.


lordofbikes:
Once again, many thanks for your efforts :+1:

PR #1041, opened 2018-11-10
Fix a repeated typo in comments

KimK:
Fix a minor misspelling in the comments that was
(probably) picked up by copy-and-paste and repeated.


lordofbikes:
Many thanks for your contribution!

PR #1044, opened 2018-11-10
Added filter field to Block List widget

xanderdin:
Similar as filter in Layer List widget.


lordofbikes:
Good job, many thanks!

Issue #819, opened 2016-09-17
"Explode" messing with colors of rectangle


Zmuogs:
Steps to reproduce:
- start a new drawing,
- create second layer,
- color of second layer change to Green,
- while second layer is "current" draw a Rectangle:![explode_before](https://cloud.githubusercontent.com/assets/20404470/18608576/f2454a50-7cf5-11e6-850b-c33f13459b21.png)
- activate command "Explode" and explode mentioned Rectangle: you will get lines with various colors:![explode_after](https://cloud.githubusercontent.com/assets/20404470/18608589/3bc639e6-7cf6-11e6-9ae7-b84dcb4198e6.png)

Version: 2.2.0-alpha
Compiler: GNU GCC 5.4.0
Compiled on: Sep 15 2016
Qt Version: 5.5.1
Boost Version: 1.58.0
System: Ubuntu 16.04.1 LTS

lordofbikes:
I can't test it at the moment, but I think the problem isn't the color. My assumption is, that the white lines are created on layer 0, which default color is white.
If so, the problem is related to losing the layer information for the new generated entities.
Can you confirm this?

Zmuogs:
I used "Modify" -> "Properties" on Lines: Layer stays same, Color changes:
![line_properties_01](https://cloud.githubusercontent.com/assets/20404470/18614400/91addc74-7d96-11e6-81c3-46f507f1f668.png)
![line_properties_02](https://cloud.githubusercontent.com/assets/20404470/18614401/91b07cfe-7d96-11e6-8e28-3ecc70b43d6a.png)

Equoda:
Yes, I confirm this, but can't investigate now due to the debugger problems.
This may be related to #755 because a rectangle is a some kind of container, like a block.

FelixNotTheCat:
@lordofbikes @r-a-v-a-s For RS2::EntityPolylines, RS_Modification::explode() (in rs_modification.cpp) resolves layers (resolveLayer = true;) but not colours (resolvePen = false;). resolvePen = true; then resolves colours of all entities in a (polylines) container. Pls say whether I can commit the change.
(PS the containers first entity gets its colour from the container as opposed to from the preceding entity in the loop.)

PR #1046, opened 2018-11-14
Exploded entities issues fix (#819)

xanderdin:
I believe this fixes that old issue. Tested with different combinations of blocks, including different layers, different colors, line types, widths, blocks inside blocks to several insertion levels. Encountered no side effects or differences in exploded results.

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -3007,6 +3007,56 @@ bool RS_Modification::round(const RS_Vector& coord,
 }
 
 
+/**
+ * Repetitive recursive block of code for the explode() method.
+ */
+static void update_exploded_children_recursively(
+        RS_EntityContainer* ec,
+        RS_Entity* e,
+        RS_Entity* clone,
+        RS2::ResolveLevel rl,
+        bool resolveLayer,
+        bool resolvePen) {
+
+    if (!ec) {
+        return;
+    }
+    if (!e) {
+        return;
+    }
+    if (!clone) {
+        return;
+    }
+
+    if (resolveLayer) {
+        clone->setLayer(ec->getLayer());
+    } else {
+        clone->setLayer(e->getLayer());
+    }
+
+    if (resolvePen) {
+        //clone->setPen(ec->getPen(true));
+        clone->setPen(ec->getPen(false));
+    } else {
+        clone->setPen(e->getPen(false));
+    }
+
+    clone->update();
+
+    if (clone->isContainer()) {
+        // Note: reassigning ec and e here, so keep
+        // that in mind when writing code below this block.
+        ec = (RS_EntityContainer*) clone;
+        for (e = ec->firstEntity(rl); e; e = ec->nextEntity(rl)) {
+            if (e) {
+                // Run the same code for every children recursively
+                update_exploded_children_recursively(ec, clone, e,
+                        rl, resolveLayer, resolvePen);
+            }
+        }
+    }
+}
+
 
 /**
  * Removes the selected entity containers and adds the entities in them as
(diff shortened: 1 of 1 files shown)


xanderdin:
Oh, the things you described happened to be unexpected to me. I tested undo but not after file reopen. Should I close the pull request for now (or you just reject it?) until I get more time to look for a better solution?

lordofbikes:
Don't worry, this happens. I only found this by chance.
You can leave this open if you can fix this approach.
We can still close it later, if you find a better solution or prefer to start a new branch for this.

xanderdin:
Just to be clear. The issue with file reopening is not consequence of my fix. The same behavior is in the master, and I think it is rather another issue, not the one described in #819. Still don't know how to fix it.

lordofbikes:
As I said, I found this by accident on my code review, but, to be honest, didn't cross check with master.
Hopefully I find some time on the weekend to have a deeper look at your fix and this issue.
Many thanks for your efforts!

xanderdin:
Meanwhile, let me explain my fix and my decisions. Hope it will help during deeper code review.

As described in #819, when exploding a rectangle only one line out of four retains properties of its parent container. And that line is the first one. So it is pretty clear the others should get its properties the same way as the first does. That's the main idea.

I found out quickly that `resolvePen` variable in explode() is always false. In LibreCAD 2.1.3 exploded entities alway were getting their pen properties from their container by the line of code (notice **ec**):

https://github.com/LibreCAD/LibreCAD/blob/f21fde6e5ee4a63b68e72ad59a3fa4268dbe1314/librecad/src/lib/modification/rs_modification.cpp#L3018

After https://github.com/LibreCAD/LibreCAD/commit/e3c6e6ca73fae7434bb53940c4c26229239b43e1 fix, exploded entities are getting their pen properties always from themselves (notice **e2**), not their parent container (resolvePen is always false).

https://github.com/LibreCAD/LibreCAD/blob/8604f171ee380f294102da6154adf77ab754d403/librecad/src/lib/modification/rs_modification.cpp#L3093-L3097

By changing resolvePen to true the issue seems to be fixed... for a simple container, e.g. rectangle or a single polyline. But what if a container contains a block of rectangles? Then, when exploded those rectangles all were getting the same deficiency with only one line retained pen properties.

Thus I got the idea of traversal down the tree of exploded entities and updating their properties for all descendants of exploded block.

As far as code for setting properties is the same for all descendants, I decided to move it from explode() to a single function which calls itself every time another container is found among descendants.

Among other lines of code I moved from explode() to update_exploded_children_recursively() is the following line of code I changed.

Was:
```
clone->setPen(ec->getPen(true));
```
I changed to:
```
clone->setPen(ec->getPen(false));
```

The reason for the change is, the former line in some cases produces bad side effects, but the latter is not. Because of the former line of code was never called (resolvePen was always false) but the latter was the same as in 2.1.3 code before the https://github.com/LibreCAD/LibreCAD/commit/e3c6e6ca73fae7434bb53940c4c26229239b43e1 fix was applied, I decided to change it as shown above.

After those changes everything became almost ok, but in some situations some entities was displayed on the screen with wrong pen properties. Happily, I noticed when clicking on 'eye' icon near any layer, toggling the layer visibility, was changing that wrong display to the right one. Investigating that issue I found that toggling layer visibility calls container->updateInserts(). (By the way, container->updateInserts() is called only when toggling single layer visibility, and not called when 'Hide/Unhide all layers' button pressed. Is that ok or a bug?) So I added that call also to the end of explode(). After that it seems everything works as expected.

I feel like parameter names of the update_exploded_children_recursively() may look somewhat misleading. But I wanted to retain their meanings from the explode(). Anyway, any suggestions are welcomed.

lordofbikes:
Many thanks, also for creating #1049

Issue #705, opened 2016-04-30
'Save' does not save a customized scale and position in 'preview'


gerardbm:
First of all, thanks for this awesome software.

Everytime I switch to the Preview Impression I have to select the scale again (1:100, instead of 1:113.514) and I have to move the drawing in a certain position.

This happens because of autocenter and autoscale. These two functions remove a saved customized impression. Also, these two functions are redundant because there are already two buttons at bottom for these cases: 'center' and 'adjust'.

Perhaps some people want autoscale and autocenter. A consensus solution would be a checkbox to allow us enable/disable these functions.

Regards.

r-a-v-a-s:
https://github.com/LibreCAD/LibreCAD/issues/592

Cantar4:
#592 in mind, solution or not at the horizon, it would be good to get a reminder 
of the scale (be it a manual 1.5:1 or an autoscale psychedelic 1.43025:1),
systematically printed at an extreme  corner of the page.

gerardbm:
This issue is fixed after the pull request 1050. It's an important improvement I've been waiting for ~3 years. With this, LibreCAD is 1000 times better IMHO, so I can save the drawings, close them, reopen the files 5 months later, modify something and print it to PDF without having to compose the layout again. Thanks, @xanderdin!

Cantar4:
100% with you but there is no more sailor at the whelm !!
 Is Lord of bikes able to release us a nightly built to this request?
--Cantar4

xanderdin:
I'm glad if my contribution makes you happy. But dear @gerardbm, please keep in mind that the 'Fixed' checkbox value is not saved in drawings. It is saved to global LibreCAD configuration. This means that if you uncheck the 'Fixed' checkbox during preview of another drawing and forget to check it back, then opening yet another drawing, which is supposed to be fixed, could still break the scale. That is how it is made, and I believe that is because dxf file cannot keep that 'fixed' value.

gerardbm:
That's good to know, @xanderdin. Anyway, it's a good improvement, I only have to check the 'Fixed' checkbox to save every file with its scale. What would happen if I opened the file in another installation of LibreCAD. Would it work?

Off topic: By the way, what would be intereseting is to have infinite previews with different layers, so I can have a model with 100 layers, a preview with 30 layers, another preview with another composition of layers, and so on... (According to your comment, I suppose DXF is not prepared for this behaviour).

xanderdin:
[...]

Issue #605, opened 2015-11-11
Fixed print scale is not saved in drawing


Harvie:
When i set fixed scale for print (eg. 1:50) it is not kept when i close librecad and reopen the drawing later...

r-a-v-a-s:
The code associated with this is currently a mess, which someone will need to clean up.
https://github.com/LibreCAD/LibreCAD/issues/592

PR #1050, opened 2018-11-25
Respect paper scale with 'fixed' checked. Issues #592, #605, #705.

### LINKED ISSUE #592: Print scale changes when re-opening file
Hello there,
I see that print scale is not stored (or at least not used) when saving & closing file, even when 'fixed' is checked, so that you have to re-type scale each time you want to re-print your document. Is it a limitation of the file format, or non-implemented code, or a bug ?
Steps to reproduce :
1. Open / create a dxf with some entities
2. Print preview -> set scale -> check 'fixed'
3. Save & quit
4. Re-open file and go to print preview.
   Thanks,
   Rupi

xanderdin:
This pull request attempts to fix #592, #605, #705.
Also it tries to make paper always fully visible when switching to preview view thus making it more consistent with any drawings.

#### Test drawings

All drawings have paper size A3, landscape.

1. Paper scale is not set, drawn a rectangle which is less than paper.
2. Paper scale is not set, drawn a rectangle which is bigger than paper.
3. Paper scale is 1:10 saved to drawing, drawn a rectangle which is less than paper.
4. Paper scale is 1:10 saved to drawing, drawn a rectangle which is bigger than paper.
5. Paper scale is 10:1 saved to drawing, drawn a rectangle which is less than paper.
6. Paper scale is 10:1 saved to drawing, drawn a rectangle which is bigger than paper.

#### 'fixed' flag is not set

All drawings display as expected, paper always fits preview window. With drawings which are bigger than paper (2, 4, 6), full drawing fits in preview window with smaller paper inside it.

#### 'fixed' flag is set

All drawings display as expected, scale always correspond to what is saved in drawings (for 3, 4, 5, 6) or 1:1 if not saved in drawings (for 1, 2). Paper and drawings are always fit to preview window.

### DIFF
--- a/librecad/src/actions/rs_actionprintpreview.cpp
+++ b/librecad/src/actions/rs_actionprintpreview.cpp
@@ -35,6 +35,7 @@
 #include "rs_coordinateevent.h"
 #include "rs_math.h"
 #include "rs_preview.h"
+#include "rs_settings.h"
 
 struct RS_ActionPrintPreview::Points {
 	RS_Vector v1;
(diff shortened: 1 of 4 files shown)


lordofbikes:
Many thanks, well done!

dellus:
This is probably the wrong place, but as someone just has been tinkering with print scaling, is it considerable to remove most of the the preset enlarging factors (2:1, 5:1 and so on) and maybe also the very large downsizing factors ( 1.10000.....) in the print-scale dropdown menu? It's always a nuisance having to scroll all through them to the most used scalings. Maybe keep 2:1 so that the users sees it's generally possible and sets it manually.

Issue #1048, opened 2018-11-21
Segfault on toggling layer properties


xanderdin:
### Steps to reproduce

1. Run LibreCAD.
2. Open any previously saved drawing.
3. Modify the drawing, e.g. draw a single line. Don't save it.
4. Try to quit LibreCAD.
5. In appeared 'Closing Drawing' dialog press 'Cancel' button.
6. Try to toggle any property of any layer by clicking on the 'eye' (or 'lock' or 'printer' or 'hash') icon of the layer.

Crash.

### Operating System and LibreCAD version info

Version: 2.2.0-rc1-73-ge469c5e8
Compiler: GNU GCC 6.3.0
Compiled on: Nov 13 2018
Qt Version: 5.7.1
Boost Version: 1.62.0
System: Debian GNU/Linux 9 (stretch)

PR #1052, opened 2018-11-29
Fixed segfault cause of #1048

xanderdin:
The segfault appeared as side effect of #1032. During my search for causes I found out that after steps described in #1048, when the 'Closing Drawing' dialog was canceled, the MDI subwindow was not activated. Therefore, **actionHandler**'s **view** and **document** wasn't properly set (they were NULL) causing the segfault. I found a piece of code probably fixing some similar issue:

https://github.com/LibreCAD/LibreCAD/blob/8604f171ee380f294102da6154adf77ab754d403/librecad/src/main/qc_dialogfactory.cpp#L126-L128

and added it to QC_ApplicationWindow::queryExit(). Segfault has gone.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -2500,7 +2500,12 @@ bool QC_ApplicationWindow::queryExit(bool force) {
         }
     }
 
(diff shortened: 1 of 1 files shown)


lordofbikes:
Many thanks, good catch :+1:

PR #1058, opened 2018-12-12
Add Plugin Offset

Making-It:
The plugin offset was based clipper algorithm.


librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -406,6 +406,8 @@ void QC_ApplicationWindow::loadPlugins() {
                 continue;
             #endif
 
+            qDebug() << fileName;
+
```
lordofbikes: we use `RS_DEBUG->print()` instead of qDebug()
also all added debug output in this file seems from testing and should be removed before merging

lordofbikes:
I will reject this pull request for several reasons:
- I added few comments between the lines about common issues
- beside the _offset_ plugin, the code contains the _myplugin_ plugin, which seems to be a test and not meant to be merged
- both plugins contains the clipper files, that is ineffective
- the two commits have +6,553,027 / −6,542,320 lines added/deleted, this will blow up the repository for effectively +10,707 lines added, whereof ~5000 lines from clipper are redundant

Beside that, it seems that the clipper license is valid to include the code into LibreCAD. But then we should consider to update the existing offset tool and not implement a plugin for that.
I haven't checked yet, if clipper comes with more dependencies we have to consider.
If you want to work on this, get in touch on IRC or Zulip to discuss this, to avoid waste of efforts (find links on https://librecad.org).

At least, when working on features or bug fixes with any forked repo, always create a new branch for it, never commit to master branch. This avoids conflicts between your fork and the upstream repo.

PR #1059, opened 2018-12-15
Modification status mark in window title

xanderdin:
Implementation of window modified mechanism (https://doc.qt.io/qt-5/qwidget.html#windowModified-prop)


lordofbikes:
Many thanks, looks great!

PR #1071, opened 2019-01-27
Multi-selection in Layer List and Block List

xanderdin:
My drawings may contain hundreds of layers or blocks. Sometimes I need to clean-up and remove many of them. But that requires a lot of clicking.

**Inconvenience**: removing _n_ layers requires _4n_ user actions (scrolling, selecting, clicking "Remove" button, clicking "Ok" button in confirmation dialog). E.g.: you need to remove 20 layers, so you have to make 80 annoying actions. The same is true and for blocks removal.

**Solution**: allowing selection of many layers at once and applying removal to all selected layers can significantly reduce user actions (down to the 3 in the best case scenario).

This pull request implements the solution. The following different item types are introduced:

* _Active item_ (current layer or block)
* _Selected item(s)_

An _active item_ can be or not be a part of _selected items_. _Active item_ in list is depicted by bold font:

_Selected items_ are highlighted by color stripes as usual:

Items are selected by clicking on them and dragging up or down. By using keyboard modifier keys and different mouse keys different kinds of selections occur.

The following actions can be performed on a group of _selected items_:

* Removal
* Toggling visibility flag
* Toggling lock flag
* Toggling printer flag
* Toggling construction flag

All those action act on items as follows:

+ If there are _selected items_, an action performed on them;
+ If there are no _selected items_, an action performed on _active item_;
+ If there are no _selected items_ or _active item_ (e.g. hidden by filter), no action is performed.

If some _selected items_ are hidden by filter, no any action affects such hidden selections. But selections are still preserved and when filter is cleared become visible and can be affected by chosen actions.

Context menus of both lists were reordered and a couple of separators were added. The principle for grouping of actions in those menus is:

* Actions which act on all items
-- separator
* Actions which act on _selected items_ or _active item_
-- separator
* Actions which act only on _active item_ or which add a new item.

**Drawbacks**:

* Now right-clicking on an item does not make it active. This should be kept in mind when applying _active item_ dependent actions (e.g. edit layer attributes, rename block, etc.). To make an item active you should left-click it first.

* _Active item_ is not a _selected item_ initially and therefore is not highlighted as it was before. It is now only depicted by bold font and displayed in status bar. I believe this is just a matter of getting used to.

Despite those little drawbacks I hope the main features could be useful for all LibreCAD users.

### DIFF
--- a/librecad/src/actions/lc_actionlayerstoggleconstruction.cpp
+++ b/librecad/src/actions/lc_actionlayerstoggleconstruction.cpp
@@ -52,20 +52,21 @@ LC_ActionLayersToggleConstruction::LC_ActionLayersToggleConstruction(
 void LC_ActionLayersToggleConstruction::trigger() {
     RS_DEBUG->print("toggle layer construction");
     if (graphic) {
-        if (a_layer) {
+        RS_LayerList* ll = graphic->getLayerList();
+        unsigned cnt = 0;
+        // toggle selected layers
+        for (auto layer: *ll) {
+            if (!layer) continue;
+            if (!layer->isVisibleInLayerList()) continue;
+            if (!layer->isSelectedInLayerList()) continue;
+            graphic->toggleLayerConstruction(layer);
+            deselectEntities(layer);
+            cnt++;
+        }
+        // if there wasn't selected layers, toggle active layer
+        if (!cnt) {
             graphic->toggleLayerConstruction(a_layer);
-
-            // deselect entities on locked layer:
-			for(auto e: *container){
-                if (e && e->isVisible() && e->getLayer()==a_layer) {
-                    if (graphicView) {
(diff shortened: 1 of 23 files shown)


lordofbikes:
I tried to continue with review today, but run out of time now, sorry.

When I started review a while ago everything worked well so far.
Iirc it was the handling of this great feature that let me delay the review.
I have to pull this again for an informed opinion, I try to do this soon.

Many thanks for now.

lordofbikes:
I finally found time for review and fixing strange behavior when using shift key for selection.
I merged it manually and applied the fix, see https://github.com/LibreCAD/LibreCAD/commit/6449ab7ca37a26f5c2de089d4c39fac14d7a714a et seq.
Many thanks again for this great feature!

manu-p:
Same suggestion on my side: the ability to select and remove a large number of blocks.

PR #1072, opened 2019-01-31
Buttons to refresh Library Browser contents

xanderdin:
This PR allows user to use newly added parts library items in Library Browser without restarting LibreCAD.

Two buttons were added:
* Refresh
* Rebuild

The 'Refresh' button allows user to see newly created directories and items in parts library. But it has a limitation -- it cannot remove deleted directories from the tree. This is where 'Rebuild' button is useful.

The 'Rebuild' button completely recreates directory structure. But it too has a limitation. After hitting this button the user loses his current position within expanded tree and gets an unexpanded tree.

So, both buttons have their usage scenarios and useful in one or another situations.


lordofbikes:
many thanks!

PR #1079, opened 2019-03-14
Option widget lifecycle

### LINKED ISSUE #859: Zooming becomes slow when drawing polylines
1. create new empty file, or open an existing file
2. Use tool Tools/Polyline/Polyline
3. Use the Zoom in our out buttons from the tool bar or using the mouse wheel.
Notice that zooming is visibly 3 times slower than usual speed of zooming (it becomes painful to use). Getting out of the polyline mode immediately stops this behaviour.

I noticed the following:
* It happens since 2.1 and is present in Master.
* It does NOT happen with the pre-built Windows version. It happens on my gentoo linux box. I cannot tell for other linux distros.
* It happens with SOME tools requiring the user to select some points, but not all tools. It happens with Polyline creation, Create Equidistant Polylines, Bevel, and Fillet. However It does happen with any of the the Circle and Ellipse drawing tools. It does not happen with Trim, Divide, Stretch modification tools.
* Zooming becomes slow even if document is empty (not connected to trying to zoom on complex layouts).
* Difficult snapping modes (such as Intersection) make it even slower, but even without snapping activated (and without grid displayed) it is still slower than what it should. Also, on the tools that do not have the problem (such as drawing circles), snapping is very fast even in intersection mode.

Because it does not happen with the pre-built version for windows, I would well accept it is connected with Qt5 under linux (not librecad code), but since it only occurs with some librecad functions, it could be that the point selection is coded a different way in different librecad funcrions so that there is a workaround. I tried Qt 5.6.2 and 5.7.1 with same results. I did not manage to compile a recent librecad with Qt4, it fails in lib/engine/rs_system.cpp:38:26 on class QStandardPaths.

Let me know if there is a way for me to produce useful debug output for this problem.

mpiffault:
Better handle destruction of option widgets, saving options to disk only when needed.
Fixes #859 .

### DIFF
--- a/librecad/src/actions/lc_actiondrawlinepolygon3.cpp
+++ b/librecad/src/actions/lc_actiondrawlinepolygon3.cpp
@@ -158,14 +158,10 @@ void LC_ActionDrawLinePolygonCenTan::updateMouseButtonHints() {
 
 
 void LC_ActionDrawLinePolygonCenTan::showOptions() {
-    RS_ActionInterface::showOptions();
-
     RS_DIALOGFACTORY->requestOptions(this, true);
 }
 
 void LC_ActionDrawLinePolygonCenTan::hideOptions() {
-    RS_ActionInterface::hideOptions();
-
     RS_DIALOGFACTORY->requestOptions(this, false);
 }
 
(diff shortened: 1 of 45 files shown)


lordofbikes:
I was conflicted between merge and reject, but finally I decided to reject this because of the bulk remove of ```showOptions()``` and ```hideOptions()```. Calling the base class implementation in overridden methods is not bad and here they are empty, so the benefit is not that much.
The other changes are also not bad, as far as I have tested, but because the issue is already solved I don't want to pull them in yet in release candidate.
Many thanks for your efforts anyhow!

mpiffault:
Yes I totally understand. Thanks for your time testing it.
Anyway, I feel like this part needs a deeper refactoring than what I did . And it was a first try at C++ for me, so I have a lot of progress to do.

Issue #1080, opened 2019-03-15
Undoing modifications made by Attributes dialog removes modified entity


xanderdin:
### Expected behavior
After applying entity modifications by using Attributes dialog, Undo should restore previous attributes.

### Observed behavior
Undoing modifications made by using Attributes dialog makes entity disappear.

### Steps to reproduce or sample file
1. Create a new drawing.
2. Draw a line.
3. Select the line and call Attributes dialog (e.g. by clicking menu Tools->Modify->Attributes).
4. In Attributes dialog, change color (e.g. to 'Red') and press 'OK' button.
5. Undo changes by pressing Ctrl+Z or by clicking menu Edit->Undo.

### Operating System and LibreCAD version info
Version: 2.2.0-rc1-105-gc2635841
Compiler: GNU GCC 6.3.0
Compiled on: Mar 15 2019
Qt Version: 5.7.1
Boost Version: 1.62.0
System: Debian GNU/Linux 9 (stretch)

It may resemble #1049, but I think they have different roots because behaviors differ.

xanderdin:
This issue is direct consequence of changes to `RS_Modification::changeAttributes(RS_AttributesData& data)` introduced by PR #762. I would ask for reverting that PR because, in my opinion, it causes more harm than good. Besides the current issue, it also can lock user in a situation when, for example, color of all entities of a block is changed but no possibility to undo those changes.

Issue #1065, opened 2019-01-05
Change layer of inserts by layer list click should not modify the layers of block entities


lordofbikes:

xanderdin:
The same **observed behavior** (layers of sub entities are changed) is when changing layer of an INSERT by using **Attributes** dialog. To the contrary, when changing layer of an INSERT by using **Properties** dialog, it acts as **expected behavior** (layers of sub entities are unchanged).

Issue #755, opened 2016-06-09
Changing attributes of a block does not change attributes of sub-entities


Equoda:
Attributes of sub-entities doesn't change when apply `Attributes` tool to the block of entities.

Steps to reproduce:
1. Create new draft
2. Create a shape (rectangle)
3. Create a shape (rectangle)
4. Create a block with 2. and 3. (select shapes -> `Create Block -> b1`)
5. Change `b1` attributes with `Attributes` tool (select `b1` -> `Attributes -> Color: red`)

Result: no visible changes in color.

`Pen` property has been changed only for `Insert` entity corresponding to `b1`, not for its sub-entities (`RS_Modification::changeAttributes`).

By the way, I can see such pieces of code for sub-entities cloning procedure as:

``` .cpp

entity->setPen(RS_Pen(RS2::FlagInvalid));
entity->setLayer(nullptr);

```

Is there were intention to show a block of entities for user using attributes of the whole block/insert, not individual sub-entities? Why such kind of initialization has been used?

Equoda:
It's all about:

Version: 2.2.0-alpha-19-gd5b8b19
Compiler: GNU GCC 4.9.2
Compiled on: Jun 13 2016
Qt Version: 5.3.2
Boost Version: 1.55.0

On:  Debian 8 x86_64 GNU/Linux

Equoda:
Fixed by pull request #762

kfoltman:
I'm not sure if this approach is correct. I mean, the new code as written will change the block *definition*, not just a single instance. Example:

- create a block
- insert the block 3 times
- change the colour on one of the 3 inserted blocks
- save, close and open the project
- now all the instances of the block have the changed colour

I came across it while working on some polyline undo bugs, and I can try figuring it out if you want.

kfoltman:
My gut feeling is that the block rendering should be checking if the insert object it belongs to has a pen override and either use the original attributes in the block definition, or the override value specified in the insert object (if present). This way the attribute change in the insert will cause the specific instance of the block to be drawn with the changed attributes, while preserving the block definition itself.

lordofbikes:
I agree, that the steps you describe are not correct.
But I have to put in a veto on your approach.

There are no attribute overrides for blocks in DXF reference. A block is a container of entities kept in a separate table of the DXF file. The instance in the drawing is a INSERT entity which has attributes like coordinates of the reference point, the rotation angle and X/Y scale factors. But no attributes like color or layer.
[...]

PR #1081, opened 2019-03-15
Fix for #1065

xanderdin:
As a reference for this fix I was using code from `RS_ActionModifyEntity::trigger()` (it is used by Properties dialog when changing layer of an INSERT). I considered this more appropriate for the issue than using code similar to that of the Attributes dialog (`RS_Modification::changeAttributes(RS_AttributesData& data)`), because the latter still has unresolved issues (#755, #1080).


lordofbikes:
Many thanks!
Possibly this fix used in ```RS_Modification::changeAttributes(RS_AttributesData& data)``` can solve #755 and #1080 too? Haven't checked the details, only a thought :wink:

PR #1084, opened 2019-03-25
Fixfor SourceForge Bug 596

kenrector:
This is a fix for [sourceforge/librecad/bugs/#596](https://sourceforge.net/p/librecad/bugs/596/) titled "DXF with elevated ARC's is not drawn properly".

The bug report and subsequent testing shows that LibreCad does not correctly handle the OCS to WCS mapping defined in the Autodesk DXF Reference material.  In this particular example the end of a cylinder generated by a 3D CAD program and drawn as a 2D side view in lLibreCad, appears as a horizontal semicircular arc instead of a vertical line.  In fact, any arcs and circles requiring the OCS to WCS transformation are drawn as simply circles or semicircles along a horizontal axis.

Although this bug has been around since the original LibreCad and Qcad implementations I believe it is only detected now because of the recent availability of 3D CAD programs that produce this dxf feature.

The fix for this bug changes the way that OCS to WCS transformations are implemented for arcs and circles.The changes include:

Modifications to the nomenclature and methods used for the OCS to WCS transformations described in a dxf file.  Uses the term OCS to WCS as a more accurate description of the Autodesk  feature instead of the previous term (extrusion).

Move the OCS to WCS transform from the low level .dxf parser out of the libdxfrw library to the rs subsystem entity initialization.  This allows flexible methods of interpreting the OCS to WCS transformation that are useful to general users of libdxfrw.  For example, a "Flatten" utility that translates 3D entities to 2D can change an entities type (to line/polyline) instead of extending the existing type (ellipse).

Derive arc and circle classes from the base ellipse class to facilitate handling transformed arc and circle entities while retaining their original properties.

These modifications allow future changes for other entity types as required by the .dxf definition.

### DIFF
--- a/libraries/libdxfrw/src/drw_entities.cpp
+++ b/libraries/libdxfrw/src/drw_entities.cpp
@@ -591,11 +591,30 @@ bool DRW_Ray::parseDwg(DRW::Version version, dwgBuffer *buf, duint32 bs){
 
 void DRW_Circle::applyExtrusion(){
     if (haveExtrusion) {
-        //NOTE: Commenting these out causes the the arcs being tested to be located
-        //on the other side of the y axis (all x dimensions are negated).
         calculateAxis(extPoint);
-        extrudePoint(extPoint, &basePoint);
+        
+        DRW_Coord v(radious,radious,0.0);
+        extrudePoint(extPoint, &v);
+
+        printf("Radious: %f       %f %f %f\n",radious,v.x,v.y,v.z);
+        
+        if (v.y > v.x) {
+            radious = v.y;
+            ratio = v.x/v.y;
+            if (ratio < 1e-10)
+                ratio = 1e-10;
+        }
+        else {
+            radious = v.x;                // get major axis in WCS
+            ratio = v.y/v.x;              // get ratio of minor to major axis
+        }
+        
+        printf("ratio:   %f\n",ratio);
+        
+        extrudePoint(extPoint, &basePoint);  //find the center point in WCS
     }
+    else
+        ratio = 1.0;
 }
 
 void DRW_Circle::parseCode(int code, dxfReader *reader){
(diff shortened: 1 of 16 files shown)


lordofbikes:
Finally I had some time to start reviewing this.
Many thanks for your efforts.

Basically your fix seems to work. I could only test with the sample from the [sourceforge ticket](https://sourceforge.net/p/librecad/bugs/596/). I see the changes in before and after results, but for a validation the 3D origin would be good. Do you have more samples, including 3D base?

I couldn't finish yet, beside some minor issues I got stuck with deriving ```RS_Arc``` and ```RS_Circle``` from ```RS_Ellipse```.
This modification leads to a segfault in the *List Entities* plugin when a circle is selected.

While the plugin issue can be debugged and solved, I'm worry about the scope of this change.
The ```RS_Ellipse``` methods are far more expensive than ```RS_Arc``` and ```RS_Circle``` in the sense of execution time. This is not an issue for simple files, but could matter on big drawings with a lot of circles. I'm undecided about this change and need more time to rethink this.

kenrector:
This file "sample_arc_edited3.dxf" contains arc and circle entities rotated according to their dxf extrusion direction vectors denoted by group codes 210, 220 and 230. 


There are 5 circles having the same center/radius properties rotated in 5 ways by the extrusion direction vectors, (0,0,0), (1,0,0), (0,1,0), (1,1,0), (0,1,1).  

There are 6 arcs having the same center/radius/endpoint properties rotated in 6 ways by the extrusion vectors (1,0,0), (0,1,0), (1,1,0), (1,0,0), (0,1,1), (1,1,1).  

There is 1 arc extracted from a Fusion360 DWG file rotated by the extrusion direction vector (1,0,0).

You must use the autozoom command to view these entities.  LC does not do this automatically when opening the dxf file for some reason.

Darn,  GIT says "We don't support that file type".  How can I submit a dxf test file.

I'm going to leave this comment and edit it when I'm able to submit the dxf test file.

kenrector:
Whoo, somehow I unintenionally closed this pull request.  I've reopened it now.  I am a tyro at git, dangerous to be around.  I hope I haven't made some mistake.

lordofbikes:
Nothing happened yet, everything is fine :wink: 
You could rename DXF files to TXT extension for upload or preferably put it into a ZIP archive.
Supported file types are listed in github help:
https://help.github.com/en/articles/file-attachments-on-issues-and-pull-requests

Have you created the sample yourself? Then one or more screenshots showing the 3D origin would be great too, to understand what's going on.

kenrector:
Attached .zip includes test files and F360 screen shot.
[#1084_samples.zip](https://github.com/LibreCAD/LibreCAD/files/3178920/1084_samples.zip)

lordofbikes:
Many thanks for the test files, they clarify the issue. It also proofs, that the fix works when opening existing drawings.
But the issues with this fix are more worse. It's impossible to draw circles, except with the 2-points tool.
All other circle tools end in a seg fault for me while drawing. And I haven't tested the arc tools now, but I expect the same behavior.
My assumption is, that this all is caused by the derivation from `RS_Ellipse`. Beside my earlier concerns about computing time, I also expect issues with DXF writer, when is comes to save such drawings after modification.

I think the implementation to fix this need some rethinking. This fix displays elevated arcs/circles correct, but what comes to my mind to consider is:
* when a user change the drawing, are the elevated entities exported correct?
* what happens, if a user modifies an elevated entity?
* how do elevated entities work with modifying tools, scale, mirror, trim and so forth?

Maybe an option is to convert elevated entities to `RS_Ellipse` or possibly we need a new entity type `LC_Elevated`?
But then still the concerns listed above has to be considered and solved.

lordofbikes:
I close this now.
As the issue exists, I created #1237 and closed the sourceforge ticket too.

PR #762, opened 2016-06-13
fix: changing attributes of a block sub-entities; #755

Equoda:
Function ```RS_Modification::changeAttributes(...)``` changed.
Overloaded recursive function ```RS_Modification::changeAttributes(...)``` added.
Sub-entities of ```RS_Insert/RS_Block``` now change their attributes together with their containers.

PR #1088, opened 2019-04-22
Fix for #1080 and #755

### LINKED ISSUE #1080: Undoing modifications made by Attributes dialog removes modified entity

xanderdin:
This tries to extend #762, fixes #1080, makes cautious approach for #755. User now has options to choose if he/she wants to apply attributes recursively to INSERTs.

UPDATED, see https://github.com/LibreCAD/LibreCAD/pull/1088#issuecomment-486094751

<del>Unfortunately, I tried but failed to implement Undo for modified recursively Block entities. Anyway, I think this is a little bit better for the moment than bare #762.</del>

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -228,16 +228,17 @@ bool RS_Modification::changeAttributes(RS_AttributesData& data) {
             bool applyDeep = e->isContainer() &&
                 (data.applyPenBlockDeep || data.applyLayerBlockDeep);
 
+            bool skipUndo = applyDeep && e->rtti() == RS2::EntityInsert;
             // Because of Undo does not work currently for modified Block-deep
             // entities, in order to prevent empty Undo entry, the line
-            // below is checking for 'applyDeep' and does not clone 'e' for
+            // below is checking for 'skipUndo' and does not clone 'e' for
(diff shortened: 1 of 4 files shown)


xanderdin:
I managed it. Applied the same approach to `RS_Modification::changeAttributes()` as in #1081 (@lordofbikes, you are right, it works for this also). Undo works for all modifications including Block-deep. Attributes dialog now looks like this:

lordofbikes:
Many thanks and great that you managed the undo issue :+1:

PR #1091, opened 2019-04-28
Margins and Tiled Printing

### LINKED ISSUE #428: option to store page margins (feature request)
I have to set page margins every time I print to pdf.

I would be nice to store it.

### LINKED ISSUE #483: Implement tiled printing
It's a feature request.
I use LibreCAD to draw simple drafts, only lines and arcs. And I wish I could print it in real scale on several pages using casual A4 printer and glue sheets together. In print preview mode I could select page number and move printing region. 
Most of people don't have those large printers, but they do want to use CAD software. LibreCAD is great ;)

baranovskiykonstantin:
fixes #428
fixes #483

### DIFF
--- a/libraries/libdxfrw/src/drw_interface.h
+++ b/libraries/libdxfrw/src/drw_interface.h
@@ -185,6 +185,11 @@ public:
      */
     virtual void addComment(const char* comment) = 0;
 
+    /**
+     * Called for PLOTSETTINGS object definition.
+     */
+    virtual void addPlotSettings(const DRW_PlotSettings *data) = 0;
+
     virtual void writeHeader(DRW_Header& data) = 0;
     virtual void writeBlocks() = 0;
     virtual void writeBlockRecords() = 0;
@@ -194,6 +199,7 @@ public:
     virtual void writeTextstyles() = 0;
     virtual void writeVports() = 0;
     virtual void writeDimstyles() = 0;
+    virtual void writeObjects() = 0;
     virtual void writeAppId() = 0;
 };
 
--- a/libraries/libdxfrw/src/drw_objects.cpp
+++ b/libraries/libdxfrw/src/drw_objects.cpp
@@ -1193,6 +1193,34 @@ bool DRW_ImageDef::parseDwg(DRW::Version version, dwgBuffer *buf, duint32 bs){
     return buf->isGood();
 }
 
+void DRW_PlotSettings::parseCode(int code, dxfReader *reader){
+    switch (code) {
+    case 5:
+        handle = reader->getHandleString();
+        break;
+    case 40:
+        marginLeft = reader->getDouble();
+        break;
+    case 41:
+        marginBottom = reader->getDouble();
+        break;
+    case 42:
+        marginRight = reader->getDouble();
+        break;
+    case 43:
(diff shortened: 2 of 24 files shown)


xanderdin:
Hi, @baranovskiykonstantin, the features you implemented look very interesting. I've played with your PR a little bit and found an issue. Please take a look at my dxf file I was testing with:

[sheet-003.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/3125320/sheet-003.dxf.zip)

First pdf (print1.pdf) is printed with LibreCAD before your changes applied:

[print1.pdf](https://github.com/LibreCAD/LibreCAD/files/3125314/print1.pdf)

Second pdf (print2.pdf) is printed with your changes:

[print2.pdf](https://github.com/LibreCAD/LibreCAD/files/3125315/print2.pdf)

The border around my drawing is cut even if I press 'No' button in 'Paper settings' dialog



This issue does not appear when printing with dxf2pdf, only when printing from GUI.

baranovskiykonstantin:
Hi, @xanderdin. It looks like you have at least two printers a real and a virtual (PDF). At that the real printer is set as default.
If I right, the following happens:
1) Print button is pressed and LibreCAD opens Print dialogue;
2) Print dialog gets the paper settings (size, orientation, margins, etc.) from the current drawing;
3) the driver of default (real) printer set margins to minumal allowed values;
4) current printer is changed to virtual PDF printer;
5) new printer is set but margins store values with previous changes;
6) OK button is pressed;
7) LibreCAD gets new values of the margins those don't match to the margins from current drawing;
8) LibreCAD shows message with question to a user "Do you want to apply changes to current drawing?":
8.a) Yes -- LibreCAD prints drawing with new margins and saves these margins values to current drawing;
8.b) No -- LibreCAD prints drawing with new margins and leaves the margins of current drawing without changes;
8.c) Cancel -- LibreCAD cancels printing and leaves current drawing without any changes.

After all, I think this issue is not a bug. This is just a pecularity of print dialogue with the default printer that has the settings limitation.

I see two ways to resolve this issue:
1. After changing printer to reset the margins;
2. Set the virtual PDF printer as a default.

If you have a better solution, I ready to hear it.

xanderdin:
@baranovskiykonstantin, thank you for detailed explanation, it helped me grasp what's going on in behind. You're right, I have a real printer. I tried your proposed solution 2 and got satisfactory result. 

Here's summary of my steps.

Despite the fact that `Print to File (PDF)` was working in LibreCAD, my system (Debian 9.9) had no *virtual PDF printer* at all. So initially I had no choice for setting such a *virtual PDF printer* to be default one. I searched and found that there is `printer-driver-cups-pdf` package in Debian repository. After installing that package I got that virtual 'PDF' printer. I set it to be default one. Then I tried to print to that default 'PDF' printer but it has its own issues. Finally as before, I chose `Print to File (PDF)` option  in print dialog.



 As you said, it took its settings for margins (all zeros) from my newly installed default virtual 'PDF' printer and printed successfully as expected. This workaround is OK for me. Thank you! :thumbsup:

lordofbikes:
Many thanks for this long missed feature, great :+1:
And thanks to @xanderdin too, for reviewing :+1:

PR #1093, opened 2019-05-02
Print Preview: Limit paper movement.

baranovskiykonstantin:
If Shift key is pressed the paper moves only horizontally.
If Control key is pressed the paper moves only vertically.


lordofbikes:
Many thanks!

PR #1095, opened 2019-05-09
Issue #930 "LibreCAD source code compile problem"

shawncurry:
- Fixed several char vs wchar problems preventing successful compile with _UNICODE flag.  Should be able to compile either way now.
- Spot tested math expressions using a few shapes like Circle: center, radius.  Seems ok so far.

### DIFF
--- a/librecad/src/lib/math/rs_math.cpp
+++ b/librecad/src/lib/math/rs_math.cpp
@@ -320,14 +320,18 @@ double RS_Math::eval(const QString& expr, bool* ok) {
     }
     try{
         mu::Parser p;
-        p.DefineConst("pi",M_PI);
+        p.DefineConst(_T("pi"),M_PI);
+#ifdef _UNICODE
+        p.SetExpr(expr_copy.toStdWString());
+#else
         p.SetExpr(expr_copy.toStdString());
+#endif
         ret=p.Eval();
         *ok=true;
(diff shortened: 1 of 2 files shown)


shawncurry:
I found that the muParser lib was being built for UTF-16 (wchar) due to the compiler defs (_UNICODE) that qt was passing.  The parser itself was building just fine that way, however doing so caused problems in rs_math.cpp and plot.cpp, which appeared to assume it was built for char, and that was not the case.

So essentially there were 2 different ways this problem could have been fixed.  Either force muParser to be built as ANSI (which the .pro files seemed to be set up to do, but it wasn't working), or fix the offending calls that were being made into it so that they select the correct char type according to the build flags.  I chose the 2nd path.

My fix corrects these units by using the macros supplied for muParserDef.h.  But there was one missing:  a conversion from a QString to the mu::string_type.  That's where a bit of ugliness crept in. 

The toMUPString() helper I added didn't seem to have a great place to live.  It could have gone with the rest of the unicode macros in muParserDef.h, except muParser doesn't necessarily care about QStrings.  So instead I just located it where it was needed.  A 3rd option I considered was adding a "lc_char.h", but this helper would have been the only thing in there.

lordofbikes:
Close it as the changes are in #1096 too.

Issue #696, opened 2016-04-11
Compile error: CDataSen::Serialize should be defined with "const"


pgovind:
"jwwdoc.cpp" defines the function CDataSen::Serialize. On _MSC_VER, the declaration makes the function const. However, the definition is not const and this leads to compile errors. I made a local change to make it const, but there are other error messages on visual studio, the first of which is on the next line :

    ofstr	<< (double)m_start.x << (double)m_start.y
        << (double)m_end.x << (double)m_end.y;

where the compiler complains that more than one operator "<<" is defined for operands std::ofstream << double. 

Not sure how others are not seeing this. Is nobody else on windows or am I doing something wrong?

r-a-v-a-s:
We use the MinGW version of Qt.

r-a-v-a-s:
I don't think anyone is opposed to supporting msvc;
dxli and I don't use Windows,
and lordofbikes probably isn't interested enough.

Feel free to submit a patch.

rvt:
We can remove jww support if we want to.

shawncurry:
I have fixed all of the MSVC toolchain issues, including this one.  https://github.com/LibreCAD/LibreCAD/pull/1096

I managed to preserve the jww support in the process.  I can still import all of the jww files I found included in the jwcad project, with the exception of the ones containing Japanese characters in the filenames.  The current release also has a problem with these ones, likely a separate filename handling issue.

PR #1096, opened 2019-05-11
Bugfix/issue914

shawncurry:
This changeset fixes the Visual Studio build and updates it to 2017.  The new project files were generated by Qt Visual Studio Tools (for 2017): https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools-19123 

In order to allow the toolkit to generate a clean, build-able project, I made a couple of changes to common.pri.  These should affect the compiler flags used in the MSVC toolchain only.

I included the project files because the old 2010 files were previously included in source control.  It might be a good idea to instead omit all of the VS project files, in favor of simply maintaining the Qt project files in such a way as to allow the Qt Tools to produce a good VS project out of them.  Then anyone who wants to work in VS will simply need to generate their own using the Tools.  (EDIT: I have now removed them - see note below)

There was also a major problem with the jwwlib:  Issue #696.  It was defining overloaded stream operators for several windows.h types, and these were in conflict with the existing ones from MS.  They couldn't simply be removed however, because their behavior seemed to be required for successful import of jww files. 
 Additionally, there appeared to be some intentional loss of data casting that was important to preserve.

In order to work around the conflicting stream operators and maintain the current behavior, I had to introduce a new set of types for jwwlib to use.  I called them "jwTYPE" where TYPE is the corresponding windows.h type.  In MSVC these will map to the windows.h types, otherwise they will map to standard c++ types.

This pull request also includes my commit for Issue #930, as it was required in order to build with Qt 5.12.3/MinGW 7.3.

These changes have all been tested on Win10 with the VS2017 toolchain, as well as the aforementioned Qt toolchain.  JWW import was successfully tested with the examples I found included with the jwcad project.


shawncurry:
After further consideration I have removed the Visual Studio project files.  The file path info from custom.pro gets included in them, so they should instead by generated by the Qt Tools by any developer who wishes to use Visual Studio.  

Any future changes needed for the Visual Studio project should be made by way of the Qt project files.

I also made a change to .gitignore to ignore VS project files to reflect the fact they should not be checked in.

shawncurry:
Once these changes have been merged, the instructions for building in Visual Studio will be as follows:

- Follow all the directions for building in Qt Creator.
- Install the appropriate [Qt Visual Studio Tools](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools-19123) plugin for your version of Visual Studio. 
- [RTFM](https://doc.qt.io/qtvstools/qtvstools-managing-projects.html). Set it up. 
- Qt VS Tools > Open Qt Project File > librecad.pro 
- Build 
- If you are planning to contribute, please do not check the VS project files in.  They will contain the personalized settings from your custom.pro.

PR #1097, opened 2019-05-12
- fix #981 [2.2.0 Release Candidate] Negative numbers are not accepte…

### LINKED ISSUE #981: [2.2.0 Release Candidate] Negative numbers are not accepted for Copy/Move input points
### Expected behavior
Copy/ Move > First point (0,0) > Second point (-1,0) > Launches Copy/Move Window

### Observed behavior
Copy/ Move > First point (0,0) > Second point (-1,0) > No response with negative input

As a workaround you can flip the points, but it is inconvenient. Prior versions did not have this issue.

MS Windows 10
2.2.0 Release Candidate

shawncurry:
- fix #981 [2.2.0 Release Candidate] Negative numbers are not accepted for Copy/Move input points
- reverses a commit authored by webmite from 10 May 2017:  Update rs_math.cpp eval fn to allow Imperial shorthand
- Issue #878, Pull Request #885
- this string editing approach appears to be bug-prone; a better solution would be to apply some pre-evaluation of the terms
- code appears to also suffer from incorrect unicode handling.

### DIFF
--- a/librecad/src/lib/math/rs_math.cpp
+++ b/librecad/src/lib/math/rs_math.cpp
@@ -33,8 +33,6 @@
 #include <QString>
 #include <QDebug>
 
-#include "rs_settings.h"
-#include "rs_units.h"
 #include "rs_math.h"
 #include "rs_vector.h"
 #include "rs_debug.h"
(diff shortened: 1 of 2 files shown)


shawncurry:
Spot tested on Win10 with Qt5.12.3/MinGW7.3, and the new Visual Studio (2017) build.  Negative numbers appear to be working again.

lordofbikes:
Many thanks for battling this down :+1: 
It seems that the missing link was the use of inch in application preferences. This was mentioned the 1st time 12 days ago. As an European, using mm, it worked for me all the time. Now I know why :wink: 

I decided to merge, because the regression weighs more than the benefit. Also, you mentioned too, there were some more issues in this solution.

Anyhow, this was a welcome and long asked for feature and so I would catch your offer to work on a more stable solution for imperial units.

shawncurry:
> Anyhow, this was a welcome and long asked for feature and so I would catch your offer to work on a more stable solution for imperial units.

Thanks Armin @lordofbikes.  I'll see what I can do.  Being in the US, the imperial units thing is probably going to come up anyway.  I have somewhat of a specialty with imperial/metric units issues, though I'm not sure if I'm better at causing them than fixing them!

Issue #1085, opened 2019-04-09
Zoom panning can only be used once and must then be reactivated again


michael0710:
While using zoom panning by clicking on the button in the toolbar...

### Expected behavior
...I want to pan as long as the zoom panning button is activated (Just like the button called Window Zoom)

### Observed behavior
...I can only pan once and as soon as I release the mouse button the button is still activated but I am not able to pan with a further click on the drawing area. I need to click on the (still activated) zoom panning button and then it will work again for one pan.

### Steps to reproduce or sample file
Just draw a simple line to be able to see something moving and it will work with every .dxf file (I am not sure if it works with other filetypes but I believe it won't.

### Operating System and LibreCAD version info
Version: 2.2.0-rc1
Compiler: GNU GCC 4.9.1
Compiled on: Feb 16 2018
Qt Version: 5.4.1
Boost Version: 1.53.0
System: Windows 10

shawncurry:
This appears to have been caused by some refactoring by dxli from 4 September 2011:  commit cc6bc9e2 "bug # 3348447, need to test more".  Handling of the middle-mouse button was removed from the mouseReleaseEvent handler.  However, this was a disambiguation necessary for proper handling of both middle-click panning, and the "Zoom Panning" (toolbar) mode.

For the case of middle-click panning, the mode should end when the mouse button is released.  For the toolbar mode, panning should return to the "open hand" state, until the mode is ended by pressing the ESC key or selecting Window Zoom mode.

pseudo-code:
```
mouseReleaseEvent(MouseButton b) {
  if (b == MouseButton::middle)
    endPanning();
  else
    resumePanning();
}
```
I have added this fix to my other pending commit.

lordofbikes:
Before the fix it was not needed to reset the Pan tool, as it only worked once.
Now it's necessary to reset the tool, when panning is finished.
With ESC reset works, but most tools in LibreCAD, if not all, could be reset by right click too.
Right click must be implemented to reset the Pan tool.

shawncurry:
Included in pull request #1125

Issue #764, opened 2016-06-14
spacebar for accepting commands


r-a-v-a-s:
I'll work on a way to have spacebar accept commands (like Enter).

http://forum.librecad.org/Do-you-use-the-cal-command-td5713782.html

r-a-v-a-s:
Space-bar for entering commands is now a reality! :sunglasses: 

The cal command now functions as a toggle:

1. cal [space-bar]
2. 2+2 [space-bar]
3. sqrt(5) [space-bar]
4. cal [space-bar]

xbutton1 has been removed as a cmd line trigger... we will see if anyone complains;
no one has ever commented about it; and I might add it as a custom menu activator.

Cantar4:
LC 2.2.0-alpha-24 installed,
how to (provisionally :) devote the space-bar to Free Snap ?

PR #1099, opened 2019-05-13
- fix #1098 Add a preference to allow whitespace in command interface

### LINKED ISSUE #1098: Add a preference to allow whitespace in command interface
As an EU of the LibreCAD command interface, I prefer the old 2.1.3 method which allowed me to use white space in my commands.

### Expected behavior

I can use white space.

### Observed behavior

Typing a space causes immediate evaluation before I was ready.

Use LibreCAD 2.2 rc1

Win10, built from master, using Qt5.12.3/MinGW7.3, also Visual Studio 2017

This behavior was added for Issue #764.  It appears to have been designed to mimic AutoCAD functionality.  This should probably have a preference; I think it will be confusing for new users who are inexperienced with AutoCAD,

shawncurry:
- add new preference to allow whitespace in the command interface (the 2.1.3 behavior)
- add new checkable button to command interface popup menu to toggle the pref
- check the pref before interpreting space as eval trigger

### DIFF
--- a/librecad/src/actions/rs_actionzoompan.cpp
+++ b/librecad/src/actions/rs_actionzoompan.cpp
@@ -98,9 +98,12 @@ void RS_ActionZoomPan::mousePressEvent(QMouseEvent* e) {
 
 
 
-void RS_ActionZoomPan::mouseReleaseEvent(QMouseEvent* /*e*/) {
-        setStatus(SetPanEnd);
-        trigger();
+void RS_ActionZoomPan::mouseReleaseEvent(QMouseEvent* e) {
+	if (e->button() == Qt::MiddleButton)
+		setStatus(SetPanEnd);
+	else
+		setStatus(SetPanStart);
+	trigger();
     //RS_DEBUG->print("RS_ActionZoomPan::mousePressEvent(): %f %f", v1.x, v1.y);
 }
 
--- a/librecad/src/ui/forms/qg_commandwidget.cpp
+++ b/librecad/src/ui/forms/qg_commandwidget.cpp
@@ -37,7 +37,6 @@
 #include "rs_commandevent.h"
 #include "rs_system.h"
 #include "rs_utility.h"
-#include "rs_settings.h"
 
 /*
  *  Constructs a QG_CommandWidget as a child of 'parent', with the
--- a/librecad/src/ui/forms/qg_commandwidget.cpp
+++ b/librecad/src/ui/forms/qg_commandwidget.cpp
@@ -37,6 +37,7 @@
 #include "rs_commandevent.h"
 #include "rs_system.h"
 #include "rs_utility.h"
+#include "rs_settings.h"
 
 /*
  *  Constructs a QG_CommandWidget as a child of 'parent', with the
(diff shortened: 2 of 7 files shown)


shawncurry:
TODO:

- Submit caption for review
- Translation

shawncurry:
allowwhitespace

shawncurry:
I defaulted it to true, by the way; preserving the 2.1.3 behavior.  Must be toggled to enable #764.

vwren:
Just an observation for a minor tweak: I think the function would be more clear if the checkbox indicated what the whitespace was actually doing.  Keep the same default, but flip the meaning, because what you're really selecting is "Space Enters Data" (default is false).

To me, unchecking "Allow Whitespace" suggests that you simply can't put any spaces in, i.e. space is blocked, not that the space bar actually does something.

shawncurry:
> Just an observation for a minor tweak: I think the function would be more clear if the checkbox indicated what the whitespace was actually doing. Keep the same default, but flip the meaning, because what you're really selecting is "Space Enters Data" (default is false).
> 
> To me, unchecking "Allow Whitespace" suggests that you simply can't put any spaces in, i.e. space is blocked, not that the space bar actually does something.

Thanks vwren.  I'm in total agreement about flipping the meaning.  It's what I wanted to do in the first place, but I couldn't think of a good way to word it.  I considered things like "Spacebar accepts commands" (from the title of #764) but I could just imagine someone in another country reading the translation and then trying to speak into their keyboard - lol.

fa201:
"One may validate the entered command by pressing RETURN or SPACE BAR" ?

shawncurry:
> "One may validate the entered command by pressing RETURN or SPACE BAR" ?

Looks like a winner 👍 lol

toolong

shawncurry:
That reminded me though, about the other part that made it difficult.  We got about 4 words to describe it before all The Title Case Starts Looking Funny.  

That's another thing my tech writer will be hassling me about.  Proper titles only, everything else gets sentence case.  I guess this sort of thing is the reason.

- SPACE Evaluation
- Evaluate on SPACE
- SPACE Accepts My Commands
- Accept Commands on SPACE
- Can You Hear Me Major Tom
- Enter Commands on SPACE

I'm drawing a blank here guys.

shawncurry:
Evaluate when SPACE BAR is pressed

That one seems great to me.  I'm taking the rest of the buttons to sentence case.  What do you think?

sentence

lordofbikes:
In the [forum thread](http://forum.librecad.org/Do-you-use-the-cal-command-tp5713782p5713938.html) mentioned in #764, is this too:
> Also note that space-bar will toggle free snap when the cmd line is empty;
this is intended to help with placing dimensions. 

This is hard coded and puzzled users too, because SPACE BAR is used to activate the command line also.
Users who are used to use SPACE BAR for command line activation were annoyed by this unrelated toggling action.
So having an option for this too would be great, default is off.

Both options are not intended to be toggled all the time. I think users who use these options will stay with them and are aware of the consequences.
What do you think about moving them into the application preferences:
grafik
We can add a new group box *Keyboard settings* for these options.
This will also give more place for comprehensive check box labels.
Example code: #1087 added a check box to application preferences recently.

shawncurry:
Hi Armin,

You are right, this setting would likely not be one that ever needs toggling once the user has chosen their preference.  The Application Preferences were the first place I went to look for this setting. That is a better place for it. 

I guess maybe I didn’t want to start thinking about that dialog yet. It seems likely that it will suffer from the “two column translation problem” I jokingly referenced earlier in this conversation.  The phrase “Modify layer of selected entities, at layer activation,” will almost certainly overrun its design length, and get cut off, when the translated phrase is substituted. 

This entire dialog should probably be redesigned to avoid these sorts of problems.

- Quick and dirty:  Give “Program Defaults” and “Startup” their own tab
- Better: reorganize the whole thing

Of course that problem deserves a ticket of its own. For now I’ll just find a place for it in Prefs.

fa201:
My proposal was more documentation oriented, I agree it is too long. 
Regarding this command line validation, I admit this is very useful and efficient to activate and validate with <space> but I can understand that it is tricky for developers. The famous : one hand on mouse and one hand on keyboard for light fast inputs !!!
Would it be acceptable from coding perspective to have exclusive options : if <space> activates the command line then another letter / letter combination can validate ?
This should exclude automatically commands and 2-letter keycode. Keep in mind that <shift> also activate command line.
I like very much the idea of option panel defining these parameters !

shawncurry:
prefs

See #1107 Create a preference to toggle Free Snap mode when SPACE BAR is pressed

shawncurry:
I've added a fix for #1085 "Zoom panning can only be used once and must then be reactivated again"

PR #1111, opened 2019-05-28
fix build with gcc 9

jirislaby:
These symbols were reported as multiply defined with gcc9. So make them
local to the translation unit.


lordofbikes:
I fixed this another way in d0a0ef284b7fcc5c0d65b92c30855f62b637653f.
Please find an explanation in #1117.
Many thanks anyway!

shawncurry:
Need to find another solution, this one breaks the windows build.

jirislaby:
@shawncurry this PR is not needed anymore as #1125 was merged instead... Or am I missing something?

lordofbikes:
@jirislaby don't worry, it was not your PR which broke the MSVC build, it was my solution, pushed without PR.
#1125 isn't merged yet, I'll gonna check it with gcc 9 before, to ensure that all compilers are satisfied again.

PR #1117, opened 2019-06-06
Use unique variable names to prevent multiple definition link errors

spotrh:
First noticed in Fedora 30. 

All of these files define the same named variable for list of entity types (enTypeList):
librecad/src/actions/rs_actiondrawcircletan1_2p.cpp
librecad/src/actions/rs_actiondrawcircletan2_1p.cpp
librecad/src/actions/rs_actiondrawcircletan2.cpp
librecad/src/actions/rs_actiondrawcircletan3.cpp

Additionally, these two files define the same named variable for circleType:
librecad/src/actions/rs_actiondrawlinetangent1.cpp
librecad/src/actions/rs_actiondrawlinetangent2.cpp

That variable gets exposed as a symbol in the compiled object files, and they conflict at link time:

/usr/bin/ld: ../../generated/librecad/obj/rs_actiondrawcircletan2_1p.o:(.rodata+0x0): multiple definition of `_ZGRN12_GLOBAL__N_110enTypeListE_'; ../../generated/librecad/obj/rs_actiondrawcircletan1_2p.o:(.rodata+0x0): first defined here
/usr/bin/ld: ../../generated/librecad/obj/rs_actiondrawcircletan2.o:(.rodata+0x0): multiple definition of `_ZGRN12_GLOBAL__N_110enTypeListE_'; ../../generated/librecad/obj/rs_actiondrawcircletan1_2p.o:(.rodata+0x0): first defined here
/usr/bin/ld: ../../generated/librecad/obj/rs_actiondrawcircletan3.o:(.rodata+0x0): multiple definition of `_ZGRN12_GLOBAL__N_110enTypeListE_'; ../../generated/librecad/obj/rs_actiondrawcircletan1_2p.o:(.rodata+0x0): first defined here
/usr/bin/ld: ../../generated/librecad/obj/rs_actiondrawlinerelangle.o:(.rodata+0x0): multiple definition of `_ZGRN12_GLOBAL__N_110enTypeListE_'; ../../generated/librecad/obj/rs_actiondrawcircletan1_2p.o:(.rodata+0x0): first defined here
/usr/bin/ld: ../../generated/librecad/obj/rs_actiondrawlinetangent2.o:(.rodata+0x0): multiple definition of `_ZGRN12_GLOBAL__N_110circleTypeE_'; ../../generated/librecad/obj/rs_actiondrawlinetangent1.o:(.rodata+0x0): first defined here
collect2: error: ld returned 1 exit status

By renaming these two variables to be unique across the codebase, we avoid the linking error. This PR makes that change, using the existing naming schema provided by those files.

### DIFF
--- a/librecad/src/actions/rs_actiondrawcircletan1_2p.cpp
+++ b/librecad/src/actions/rs_actiondrawcircletan1_2p.cpp
@@ -38,7 +38,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 namespace{
 //list of entity types supported by current action
-auto enTypeList={RS2::EntityLine, RS2::EntityArc, RS2::EntityCircle};
+auto enTypeList1_2={RS2::EntityLine, RS2::EntityArc, RS2::EntityCircle};
 }
 
 struct RS_ActionDrawCircleTan1_2P::Points {
@@ -181,7 +181,7 @@ void RS_ActionDrawCircleTan1_2P::mouseMoveEvent(QMouseEvent* e) {
     }
     case SetCenter: {
 
-        //        RS_Entity*  en = catchEntity(e, enTypeList, RS2::ResolveAll);
+        //        RS_Entity*  en = catchEntity(e, enTypeList1_2, RS2::ResolveAll);
 		pPoints->coord= graphicView->toGraph(e->x(), e->y());
         //        circles[getStatus()]=static_cast<RS_Line*>(en);
         if(preparePreview()) {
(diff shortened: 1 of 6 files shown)


lordofbikes:
Many thanks for this detailed report and providing a valid solution :+1: 

Basically I would say this is a linker issue, which should be solved at its source too.
The scope of an unnamed namespace is only global inside its translation unit and should not interfere with other units.

Have you already reported this to the linker maintainers?
If not, can you please post more information about your setup ```g++ -v```, ```ld -v``` and the command line when the linker is called? Then I can report this.

Concerning your solution I have no doubts, it's valid and solves the issue.
But after review I see no reason why these constants should live globally at all. I assume this is a relic from pre C++11 times. 
So when touching this code I'd prefer a more robust solution by moving the ```enTypeList``` into the class declaration as private member, e.g.:
``` cpp
class RS_ActionDrawCircleTan1_2P : public RS_PreviewActionInterface 
{
: 
: 
private:
    std::initializer_list<RS2::EntityType> enTypeList {RS2::EntityLine, RS2::EntityArc, RS2::EntityCircle};
}
```

You're welcome to make these changes to your pull request, but it's also fine, when you say that I should do this for you.

lordofbikes:
I just recognized, that this is a duplicate of #1111 too.
I fixed this by my proposed solution in d0a0ef284b7fcc5c0d65b92c30855f62b637653f.
Many thanks anyway!

Issue #1108, opened 2019-05-23
Improve User Experience while working with multiple drawings


shawncurry:
As an End User who often works with multiple drawings open, I want to have my drawing window settings remembered, and to be able to save all my changes with one click.

### Expected behavior

The program remembers my preferences and provides a way to "Save All" changes.

### Observed behavior

It does not.

### Steps to reproduce or sample file

- Open multiple drawings
- Choose Drawings -> Tab mode
- Click Zoom Extents
- Close LibreCAD  <-- prompts to save each file
- Reopen LibreCAD  <-- returns to Sub-window mode, with windows maximized

### Operating System and LibreCAD version info

Win 10
Version: 2.2.0-rc1
Compiler: GNU GCC 4.9.1
Compiled on: Feb 16 2018
Qt Version: 5.4.1
Boost Version: 1.53.0

shawncurry:
Let's face it - the Drawings menu is clunky.  When you switch between tab and sub-window, the menu choices change.  The "feel" of this menu could benefit from some checkable menu items, and possibly sub-menus.

The Save dialog that appears when you close the program, could possibly use a "Do this for the next <n> items" checkbox.  It also has a dated look.

A "Save All" button in the toolbar and File menu would also be handy.

shawncurry:
The auto-save behavior should also be reviewed for this case. 

#1086 improve autosave for unnamed files

shawncurry:
Actions

File -->
Close
Close All
Save unnamed1
Save unnamed1 As...
Save All

pseudo-code:
```
save() {
  if (dirty && showSaveAsIfNeeded(currentFileName; default="") == canceled)
    return false;
  return true;
}

saveAll() {
  foreach (document) {
    if (save() == false)
      return false;
  }
  return true;
}
```

etc

shawncurry:
- Tab names should not include the path; only the file name
- Full path can be displayed in the hover hint

shawncurry:
![newdrawingmenu](https://user-images.githubusercontent.com/50385855/59171505-33b95680-8b11-11e9-892c-8395d29cb534.png)
![newdrawingmenu2](https://user-images.githubusercontent.com/50385855/59171516-39af3780-8b11-11e9-82e8-d52223de616e.png)
![newfilemenu](https://user-images.githubusercontent.com/50385855/59171536-52b7e880-8b11-11e9-8f4a-67915529bcaa.png)
[...]

PR #1128, opened 2019-06-23
Bugfix/saveasandclose

shawncurry:
This fixes several issues with the changes for #1108.
- Fixes an endless loop that occurred when trying to save a write-locked file
-- (present save as dialog if the file is not writable)
- Fixes a bug preventing Print Preview window from closing, except by closing it's parent window
- Fixes a windows-specific file permission checking issue (enable support for NTFS security in Qt)

It also includes the changes from my other open PR, which fixes a problem with the windows build, and the panning toolbar mode.

### DIFF
--- a/librecad/src/main/qc_mdiwindow.cpp
+++ b/librecad/src/main/qc_mdiwindow.cpp
@@ -262,9 +262,15 @@ void QC_MDIWindow::slotWindowClosing()
 void QC_MDIWindow::closeEvent(QCloseEvent* ce) {
 
     RS_DEBUG->print("QC_MDIWindow::closeEvent begin");
-    
-	ce->ignore(); // handling delegated to QApplication
-	emit(signalClosing(this));
+
+	if (!parentWindow) {
+		ce->ignore(); // handling delegated to QApplication
+		emit(signalClosing(this));
+	} else {
+		slotWindowClosing();
+		ce->accept();
+	}		
+		
     RS_DEBUG->print("QC_MDIWindow::closeEvent end");
 }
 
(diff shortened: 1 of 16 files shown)


lordofbikes (commented):
Sadly neither GCC 7 nor GCC 9 build your fix.
Linking fails because of undefined reference to ```qt_ntfs_permission_lookup```.
Maybe defining Windows guards around ```qt_ntfs_permission_lookup``` can fix this, when it works on Linux either.
I have to check this or if I miss a Qt package on my Linux test machines.

shawncurry:
Hi Armin @lordofbikes,

I don't think you're missing anything; I think I misunderstood the [Qt manual](https://doc.qt.io/qt-5/qfileinfo.html) on this change.  It doesn't explicitly say that it's not defined in other environments, but it's certainly only something that is needed for Windows file systems.  I have added some #ifdef _WINDOWS guards around these lines.

librecad/src/main/qc_mdiwindow.cpp
```diff
@@ -449,6 +458,9 @@ bool QC_MDIWindow::slotFileSaveAs(bool &cancelled) {
     QG_FileDialog dlg(this);
     QString fn = dlg.getSaveFile(&t);
 	if (document && !fn.isEmpty()) {
+		QFileInfo info(fn);
+		if (!info.isWritable())
```
lordofbikes: ```isWritable()``` is false for new files, at least on Linux.
This inhibits to save new files.
Try to save the default *Unnamed* drawing, it's impossible.
The app doesn't hang, as you can close with *cancel*, but only without saving modifications.

lordofbikes:
Thanks @shawncurry, the build works now, but sadly still not the fix.
I can't save a new file. See my comment in code above.

librecad/src/main/qc_mdiwindow.cpp
shawncurry: I believe I have fixed both problems you mentioned, but before I submit I'm going to do some more thorough testing, probably with a spreadsheet to make sure I've covered all of the cases.  Hopefully I can begin to redeem myself on the next feature.

lordofbikes:
I cherry-picked following:
3bc9338, ad5b815, 7279db7, 0321595
and rejected:
3789303, c0f0aed, d56bae2

lordofbikes:
cherry-picked 3789303, the reject was by accident

PR #1136, opened 2019-08-12
Make layout of the Divide plugin independent of pixel density

gwhitney:
Prior to this commit, the dialog for the Divide plugin contained
  numerous explicit specifications of the number of pixels to leave for
  various items. As a result, the dialog was unreadable to the point of
  being nearly unusable on displays with pixel densities outside the
  originally contemplated design range.

  This change removes all explicit geometry specifications
  in favor of using QLayout objects to construct the dialog and place all of
  the widgets. This way, the dialog can flexibly adapt to whatever display
  and fonts are being used.

  As a small side cleanup, this commit resolved one confusion between the
  size and quantity fields, and made the labeling of these primary parameters
  clearer. Also, it removed the explicit calculation of where the error message
  box would be placed on screen, which was using deprecated Qt calls; the
  plugin should rely on Qt to do the right thing in terms of locating alert
  boxes.

  And finally, as an aid to understanding the code, this commit removed
  the "magic numbers" in the call to dividedlg::choice in favor of a
  private enum so that the calls could be more self-documenting.

### DIFF
--- a/plugins/divide/divide.cpp
+++ b/plugins/divide/divide.cpp
@@ -72,19 +72,23 @@ void divide::execComm( Document_Interface *doc,
         msgBox.setText( text );
         msgBox.setIcon( QMessageBox::Warning );
 
-        msgBox.show(); //need show to get msgBox size
-        QPoint centerXY = findWindowCentre();
-        int x = centerXY.rx() - ( msgBox.width() / 2 );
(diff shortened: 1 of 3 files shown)


lordofbikes:
Well done, many thanks!

PR #1137, opened 2019-08-12
Fix persistence of the angle in a MoveRotate operation

gwhitney:
Prior to this change, because of an inconsistency between
    librecad/src/ui/forms/qg_moverotateoptions.cpp using "MoveRotate" for
    the name of the settings holding the angle but
    librecad/src/ui/forms/qg_lg_moverotate.cpp using "MoveRotateAngle", the
    previously used value of this angle would not show up in the next
    invocation of the MoveRotate dialog; rather, it would always be 30 degrees.

    This commit fixes this unhelpful behavior by conforming both files
    to use "MoveRotateAngle" as the name of the setting. Now the previously
    used value of the angle does show up in the next invocation of the dialog.

### DIFF
--- a/librecad/src/ui/forms/qg_moverotateoptions.cpp
+++ b/librecad/src/ui/forms/qg_moverotateoptions.cpp
@@ -44,7 +44,7 @@ QG_MoveRotateOptions::QG_MoveRotateOptions(QWidget* parent, Qt::WindowFlags fl)
 
 void QG_MoveRotateOptions::saveSettings() {
     RS_SETTINGS->beginGroup("/Modify");
-	RS_SETTINGS->writeEntry("/MoveRotate", ui->leAngle->text());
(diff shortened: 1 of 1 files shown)


lordofbikes:
This fix still doesn't work for me on Linux.
On which OS have you applied and tested this?

lordofbikes:
Many thanks!
I finally get behind this and solved it with an additional commit. It was a bit tricky, because the tool option bar widget is created and destroyed multiple times within one move and rotate action. And when the angle was changed in the final dialog, the option bar angle finally saved the old value.
So it should work by now, but it's still messy. The main issue is about syncing values between tool option bar and tool dialogs. This issue may exist in other tools with *tool option bar widget* **and** *tool dialog* too.

gwhitney:
Thanks! Sorry I missed your previous message back in March, it was the middle of a tough time for me. Very much appreciate your improving this fix - I am a big LibreCAD fan and will be using it again when things settle down.

PR #1147, opened 2019-09-27
Angle constraint with pressed "shift" key also for moving and copying entities

susnicek:
The LibreCAD has a great feature: when drawing a line by a mouse and holding the "shift" key during placing the line endpoint, one can move the cursor only along certain angles (0°, 15°, 30°, 45°, etc.).

This code modification enables this feature also for moving and copying entities. I. e. when the reference point is selected and the entity is moved, by pressing the "Shift" key one can move the reference point (and the entity) only along the certain angles (0°, 15°, 30°, 45°, etc.) measured from the original reference point position.

### DIFF
--- a/librecad/src/actions/rs_actiondefault.cpp
+++ b/librecad/src/actions/rs_actiondefault.cpp
@@ -35,6 +35,10 @@
 #include "rs_preview.h"
 #include "rs_debug.h"
 
+#include "rs_line.h"
+#include "cmath"
+#include "rs_coordinateevent.h"
+
 struct RS_ActionDefault::Points {
 	RS_Vector v1;
 	RS_Vector v2;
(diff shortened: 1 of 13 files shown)


lordofbikes:
Many thanks!

lordofbikes:
The implementation of `RS_Snapper::snapModeToInt()` and `RS_Snapper::intToSnapMode()` was very error-prone. I didn't recognize the erroneous implementation of `snapAngle` when merging this.
The implementation based on the correct order of flags and it was reverse in both methods.

Now I found the issue, that snap mode wasn't restored correct after application restart. This was because the bits were messed up between save and restore.
Hopefully my new implementation of `RS_SnapMode` helps to avoid this in future.
Even when the flags are ordered now, the result is the same even when they were ordered randomly.

susnicek:
Sorry for troubles. I probably messed the bits, during the moths I forgot the details about my commit. Thank you for making it foolproof.

lordofbikes:
No problem, that's how life goes :wink:

PR #1167, opened 2019-12-11
Fit to whole page.

baranovskiykonstantin:
Hard coded empty field between a drawing and page edges is not needed anymore.
To this purpose a user has to use page margins.


lordofbikes:
Many thanks!

PR #1168, opened 2019-12-11
Print Preview: Line width scaling

baranovskiykonstantin:
By default pen width is not scaled on print and print preview.
This is the standard (AutoCAD like) behavior.
But some times it may be useful to enable the pen width scaling when
printing A1 drawing on A4 paper in testing purposes, for example.

### DIFF
--- a/librecad/res/icons/icons.qrc
+++ b/librecad/res/icons/icons.qrc
@@ -200,5 +200,6 @@
         <file>noprint.svg</file>
         <file>close_all.svg</file>
         <file>save_all.svg</file>
+        <file>scaleLineWidth.svg</file>
     </qresource>
 </RCC>
--- a/librecad/res/icons/scaleLineWidth.svg
+++ b/librecad/res/icons/scaleLineWidth.svg
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="256"
+   height="256"
+   viewBox="0 0 72.248892 72.248892"
+   id="svg4295"
+   version="1.1"
+   inkscape:version="0.92.4 5da689c313, 2019-01-14"
+   sodipodi:docname="scaleLineWidth.svg"
+   inkscape:export-filename="D:\LC_Recources\icons\scale.png"
+   inkscape:export-xdpi="11.25"
+   inkscape:export-ydpi="11.25">
+  <metadata
+     id="metadata4325">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs4323" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1364"
+     inkscape:window-height="725"
+     id="namedview4321"
+     showgrid="true"
+     units="px"
+     inkscape:snap-grids="true"
+     inkscape:zoom="1.8613281"
+     inkscape:cx="37.166193"
+     inkscape:cy="137.66745"
+     inkscape:window-x="0"
+     inkscape:window-y="20"
+     inkscape:window-maximized="0"
+     inkscape:current-layer="svg4295"
+     inkscape:object-nodes="true"
+     inkscape:snap-object-midpoints="false">
+    <inkscape:grid
+       type="xygrid"
+       id="grid4867" />
+  </sodipodi:namedview>
+  <path
+     style="opacity:1;fill:none;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:2.82222223;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+     d="M 10.922,33.866668 H 34.064224 V 65.108669"
+     id="rect4140-6"
+     inkscape:connector-curvature="0"
+     sodipodi:nodetypes="ccc" />
+  <path
+     style="fill:none;fill-rule:evenodd;stroke:#000000;stroke-width:5.64444447;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+     d="m 10.922,10.809115 h 46.848892 l -6e-6,54.017332"
+     id="path4697"
+     inkscape:connector-curvature="0"
+     sodipodi:nodetypes="ccc" />
+  <path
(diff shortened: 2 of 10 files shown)


lordofbikes:
Great improvement, well done.
Many thanks!

dellus:
Testing it now as it is merged I found it works properly (as I have imagined it) only with drawings originally intended for 1:1 scale. But if I for instance want to print out an A2 1:50 architectural drawing as A4 1:100, I get very thin lines, like with linewidth 0.
So to work as I had expected the line width scaling would have to be based on the ratio of the intended to the actual print scale.
Are you aware of this restriction, has this been considered? Maybe with thinking, well, it's still a solution to produce a legible small size print of a larger plan?
I don't have a good idea how to do it better myself, wouldn't be right to couple it to General Scale in the dimensions settings. So no criticism.

baranovskiykonstantin:
@dellus, there is no need to change general scale.
If you have a drawing A2 1:50 and you wish to print it out on A4 paper, do the next:
1) load your drawing to LibreCAD;
2) open print preview;
3) set paper size to A4;
4) set _Print Scale_ to 1:2;
5) click _Center to page_ button;
6) activate width scaling and width will be reduced accordingly to print scale.

dellus:
Hello Konstantin,
seems you didn't understand me right. I didn't want to change General
Scale, just thought of it as an indicator in the file for the intended
print scale. And with a drawing 1:50 I didn't mean it drawn 1/50 the
original size, but as usual in CAD at true size, then scaled down in the
print process only.
I have attached some sample files, did it as E-mail for not to litter at
github.
Used 1:100 A4 downsized to 1:200 A6 so it can be tested with ordinary
printers. First the dxf, I didn't include the tree bitmap, not of
importance.
arch100a4.pdf is as the normal intended print looks. arch200a6_nolinescale
as it looks downscaled in print preview without linewidth scaling: lines
too fat. arch200a6_linescale_as_is = with  linewidth scaling : 0.5 mm
/200 = 0.0025mm, way too thin!
arch200a6_linescale _should  -  that's what I want to achieve: in 1:200
lines half as thick as in 1:100.

Cheers - dellus

baranovskiykonstantin:
> I have attached some sample files

@dellus, I don't see any attachment. Could you send it to my e-mail, please?

baranovskiykonstantin:
You prepared the drawing in wrong way.
To correct the drawing I have done the next:
1) open dialog _Options -> Current Drawing Preferences_;
2) reset _General Scale_ to 1;
3) set _Length Factor_ to 100;
4) applay the settings with OK button;
5) run command _Tools -> Select -> Select All;
6) run command _Tools -> Modify -> Scale;
7) specify reference point to 0, 0;
8) set scale factors for X and Y to "1/100";
9) finish modification with OK button.

After these manipulations the drawing looks correctly.
I'm attaching the original drawing and the edited one so you can compare them and feel the difference.

[drawings.zip](https://github.com/LibreCAD/LibreCAD/files/4355043/drawings.zip)

dellus:
I wanted to avoid restructuring the original drawing, instead doing it via
the print process only. As I said in my former posting, in CAD the
standard procedure is to draw in real world sizes, and to scale down in
printing only. Or am I wrong? Might have to ask a third person.
But never mind, you have shown a workaround to achieve the desired output.

dellus:
Some more to say: A solution in my sense would need the scaling for the
drawing's geometrics and for the line widths to be set separately, or for
the line widths to be found out automatically. Thus my thought of General
Scale, but it's not obligatory to use General Scale.

baranovskiykonstantin:
@dellus, I've prepared the patch: https://github.com/LibreCAD/LibreCAD/pull/1208

dellus:
I cannot test your patch. Does it mean linewidth is scaled directly by the
(inverted) value of General Scale? That's not what I was thinking of. In
my idea it would be scaled by the ratio of General Scale to Print Scale in
Print Preview. For example a drawing set up for A4 1:100 shall be printed
A6 1:200, the linewidth scaling would be 100:200 = 1/2 or 0.5. This
LibreCAD could calculate by itself.
But what if General Scale has been left to 1 and adaption of text height,
arrow size etc of dimensions are done individually? Then the calculation
would be 1:200 = 0.005.
Hmm, same result as in the unpatched version. So it seems to work for both
scenarios. Or is my thinking wrong?

dellus:
At second look it's still not perfect for any scenario. Input and opinion
of other users/developers might be helpful. But maybe it's not worth the
hassle for a not so important goal. Perhaps you should invest your time
and efforts into more important issues...

baranovskiykonstantin:
> At second look it's still not perfect for any scenario.

The attached archive contains three files. Two of them:
arch200a6_linescale_as_is.pdf
arch200a6_linescale_as_should.pdf
are from your e-mail.
Another one:
arch200a6_linescale_as_is_patched.pdf
was generated by me from patched version of LibreCAD.
*_patched.pdf matches *_should.pdf so I don't understand what is wrong now.

[patched.zip](https://github.com/LibreCAD/LibreCAD/files/4363086/patched.zip)

GSStnb:
HI Dellus and Konstantin,
I'm a little confused, but if I understand correctly the objective is to scale a printed drawing (similar to what you would get if an A2 drawing was printed to a pdf and then the pdf was printed to an A4 page).
If that is not right, here are a couple of thoughts:
   1. Line thickness should remain constant as a real-world measure, irrespective of the drawing scale. For example, 0.18mm is typically used for an outline of an object.  That would be true for a drawing of an 1" bolt and scaled up to fit on a A2 page or a building drawn on an A4 - the outline width will always be 0.18 on the printed page.  Back in the day, using technical pens and pencil, the line width was constant no matter what the scale was.  The scale was determined early on so the object made best use of the "page-space".
   2. If it is a matter of scaling a print (vs a drawing), then the line width should be scaled.  But, if that is the case, then printing (or exporting) to a pdf and reducing it to fit a smaller paper serves the same purpose.

baranovskiykonstantin:
@GSStnb, Line Width Scaling is a new option that is available on Print Preview toolbar.
The first case from your message is a default behavior and it corresponds to disabled state of line scaling.
The second case is a target of enabling line scaling.

dellus:
@ Konstantin:  Sorry if I have made you angry. Your patched.pdf looks
right. I will wait and see how it works when available in the nightly
rc-build.

@ GSStnb:  The issue is your Nr. 2),   Konstantin has introduced as a new
feature the possibility to scale the linewidth at printing for a quick way
for a smaller printout. The problem was that if you want to scale down for
instance an A2 1:100 print to A4 1:200, the line width became 1/200 of the
width, very thin.

Your suggestion to make a pdf first and then print that to A4 size sheet
is a good workaround.
But probably we don't need that with Konstantin's patch.

GSStnb:
HI Dellus and Konstantin, that helps.  Thanks clearing up my confusion.

baranovskiykonstantin:
> Sorry if I have made you angry.

@dellus, I am not angry at all. I just trying to understand, have we been reached the goal or it needs more work.

dellus:
I see that at your new commit the green ok - tick is missing, there's that orange warning dot instead. Does this mean you are waiting for a final answer by me? 
If this is the case: I don't know how you have conceived the reworked feature. If I assume it's the way I have imagined (using ratio of general scale to print scale), it will work with files using General Scale to adjust dimension elements' sizes to print scale, and with files set up to be printed 1:1. It will not work properly with files originally set up to be printed in some scale like 1:100 and do not use General Scale, but have manually adjusted dimension elements' sizes.
So I wrote a bit nonsense before.
This is such a variant of my sample file: file:///D:/arch_cm_100_a4_uebung_220_nogenscale.dxf

As most people will use General Scale, I think this is a shortcoming to live with, it would be too complicated to cover this scenario too. I don't have an ideo how to do it. So I vote for to give it a go as it is.

dellus:
File insertion didn't work again. Will send to your email adress.

baranovskiykonstantin:
> Does this mean you are waiting for a final answer by me?

No, it is not. That is an automatic checking system.
The build has passed successfully:
https://travis-ci.org/github/LibreCAD/LibreCAD/builds/665131523
but the result is not reflected yet on the pull-request page.

> So I vote for to give it a go as it is.

OK.

lordofbikes:
I've merged #1208 now, but without much testing.
@dellus, can you please check with nightly builds?
This change is in 2.2.0-rc1-234-g1f7f4b64, so you need -234- or higher number.

dellus:
Works now as it should, also properly with other print scales than 1:1, if General Scale is used.
Version: 2.2.0-rc1-242-ge8d9114c

PR #1169, opened 2019-12-11
Fix Print Preview closing

baranovskiykonstantin:


lordofbikes:
I cherry-picked the commits 22203e6, cb7ac73 and d3e0b75.

For the d3e0b75 changes I recognized some minor issues:
* The print preview is not more than the window title bar when normalized, possibly we have to set a minimum size for the window.
* When the drawing and print preview are arranged side by side, changes in the drawing are applied to the print preview only when it gets the focus. With previous behavior this was not an issue, but now it could be disturbing. Maybe we need some kind of parent/child sync mechanism here.

I rejected 2bd7b07 for now. I understand the sense of this change, but I can also imagine, that there are user who like the auto zoom in print preview. They will miss the functionality when change this fix.
I suggest an option checkbox in application preferences, where the user can control the print preview auto zoom.

If you want to create the application preference option, I suggest to move this commit to another branch and close this PR.

Many thanks for your efforts!

baranovskiykonstantin:
> I rejected 2bd7b07 for now. I understand the sense of this change, but I can also imagine, that there are user who like the auto zoom in print preview. They will miss the functionality when change this fix.

@lordofbikes, that commit is needed for fixing the next issue:
1) create new drawing;
2) draw a circle bigger than paper (for example with radius 1000 mm);
3) open print preview (drawing looks fine at this step);
4) set printing scale to 1:1;
5) set mark to checkbox "fixed" to lock printing scale;
6) close print preview;
7) open print preview again - you will see that the size of the paper is about to zero.

When the drawing is bigger than the paper and the printing scale is fixed we have to apply auto zoom.

lordofbikes:
I cherry picked the rejected commit now too.
Many thanks for explaining the issue it solves.

PR #1174, opened 2019-12-15
Saving new and unmodified drawings as...

### LINKED ISSUE #1156: "Save As" works only with modified drawings
1. Open a file and modify it
2. Save the file.
3. "Save As" the same drawing. The file should be always "saved as", but no effect.
4. Modify again.
5. "Save As" the same drawing. The file is correctly "saved as".

### LibreCAD: 2.2.0-rc1-242-gc4387f10
Compilatore: GNU GCC 5.4.0
Compilato in data: Aug 28 2019
Versione Qt: 5.5.1
Boost Version: 1.58.0

baranovskiykonstantin:
Fixes #1156


lordofbikes:
I close this, because most changes already found there way into the master branch.

Except fd8f468, I don't understand the need for this. When I open a new drawing, why should the unnamed document be marked as modified? It will be marked as modified with the first action that changes the document, even zooming does this. When the unnamed document was not modified, I don't want to get asked at closing if I want to save it.

Many thanks anyway!

baranovskiykonstantin:
I was confused there was no way to save a new empty drawing on beginning. But the next commit aab3ca5 is resolving this issue.
You are right, the fd8f468 commit is redundant.

PR #1175, opened 2019-12-21
Fix pasting from clipboard to block.

### LINKED ISSUE #1011: Cannot paste geometry into a block.
Happy to help the development of this great program!

### Expected behavior
Geometry can be copied from one block into another, or from the "parent" drawing into the "child" block. Sorry if I am using the wrong vocabulary.

### Observed behavior
1. Geometry copied from the parent drawing and pasted into a child block does not show as a preview in the target block. When you click where to place the geometry or specify a reference point, no geometry appears in any layer of the target block.
2. Geometry copied from one block and pasted into another shows the preview of the copied geometry until you click where to place it or specify a reference point. Then it disappears. The geometry is not visible in any layer of the target block.

Neither of these behaviors show anything in the command line beyond prompting for reference points.

Create a new drawing. Draw a line in your drawing.
Create a block and name it "A". Attempt to copy the line from the drawing and paste it into block A. First behavior occurs here for me.
Draw a line in block A.
Create a block and name it "B". Attempt to copy the line from Block A and paste it into block B. The second behavior occurs here for me.

I have attached an example file in case it may be helpful.
[Testfile.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/2333760/Testfile.dxf.zip)

Version: 2.2.0-rc1-36-g439d75d
Compiler: GNU GCC 5.4.0
Compiled on: Aug  7 2018
Qt Version: 5.5.1
Boost Version: 1.58.0
System: Linux Mint 18.3

baranovskiykonstantin:
Fixes #1011

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -506,7 +506,7 @@ void RS_Modification::paste(const RS_PasteData& data, RS_Graphic* source) {
 
     // create block to paste entities as a whole
     QString name_old = "paste-block";
-    if (data.blockName != NULL) {
+    if (data.blockName != nullptr) {
         name_old = data.blockName;
     }
     QString name_new = name_old;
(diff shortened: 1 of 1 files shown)


lordofbikes:
Many thanks for fixing this!

PR #1177, opened 2019-12-25
Block may contain inserts of other blocks except itself.

baranovskiykonstantin:


lordofbikes:
Many thanks for this enhancement.

This was a bit tricky, because of conflicts with #1071 
Also undo is still a little crappy.
Undo was part of the conflicts and I tried several options, but none was satisfying.
With your implementation, adding the block remove to the graphics undo cycle I had some unexpected data loss:
1. open a block with inserts
2. delete the block which is inserted in the first block
3. the block is deleted from block list and both drawings, the main document and 1st block
4. in block editor press STRG-Z once, what restores all the inserts but not the block
5. close the block editor
6. the inserts are still visible in main document
7. save the document, close and reopen it
8. the inserts are missing because the block list wasn't restored

Putting all undos into the graphics undo cycle failed, because then undo is impossible from block editor.
Add everything to document and graphic undo cycles was also bad. Undo in the block editor restores the block and inserts, but then undo in the main document deletes them again.
My solution now is to add everything into one document undo cycle. This restores block list and inserts with one undo. But the drawback is, that when the block editor is closed, its undo stack is deleted and undo of deleted blocks is impossible from main document.
This is not optimal, but I think the most reasonable solution for now.

PR #1178, opened 2019-12-26
Handle fractional imperial dimensions

Sleet01:
Implementation for https://sourceforge.net/p/librecad/feature-requests/41/
Adds the ability to parse commands containing unit symbols.

The idea behind this implementation was to add a few new measurement specification parsing options based on the number of feature requests I found, and to make it possible (and not too difficult) to add further parsing options in the future as needed.

Accepted command patterns (Negative symbol "-" may precede measurement string):
1) DMS with required degree and optional minutes/seconds:
- Degrees: degree[s] | deg | [Dd] | ° <required, else not matched>
- Minutes: minute[s] | min | [Mm] | ' <optional>
- Seconds: second[s] | sec | [Ss] | " <optional>
2) Metric string containing meters, centimeters, millimeters:
- Meters: meter[s] | m <optional>
- centimeters: centimeter[s] | centi | cm <optional>
- millimeters: millimeter[s] | milli | mm <optional>
3) Imperial with rational inch parts, e.g. "11ft6-7/8in":
- Yards: yard[s] | yd <optional>
- Feet: feet | foot | ft | ' <optional>
- Inches: inches | inch | in | " <optional>
-- Rational inch part: [-+]NN/DD <optional, must precede unit symbol>

Provisos:
1) Arithmetic operators are not allowed within measurements containing unit symbols (parsing would be too complex).
2) DMS, Metric, and Imperial measurement unit symbols may not be combined or mixed.
3) If no unit symbols are used, assumes the command is using default units for the GUI.
3) If no unit symbol is provided, or none of the above patterns match, the original command will be passed through to p.eval() unchanged, possibly causing syntax errors.
3) Currently there is no unit testing of the new functions.

I would like to add some basic unit tests so that future changes to the regex or logic can be vetted, but I'm not sure how to run the existing tests.  Is QT Creator required?

Many thanks!

### DIFF
--- a/librecad/src/lib/math/rs_math.cpp
+++ b/librecad/src/lib/math/rs_math.cpp
@@ -264,25 +264,33 @@ double RS_Math::eval(const QString& expr, double def) {
  * various unit symbols into the current user unit (cm or inch)
  */
 QString RS_Math::derationalize(const QString& expr) {
-	RS_DEBUG->print("RS_Math::derationalize: '%s'", expr);
+	std::cout << "RS_Math::derationalize: " << expr.toStdString() << std::endl; 
 	QRegularExpression qreg(
-		R"(^(?:(?P<yards>\d+\.?\d*)(?:yards|yard|yd))?)"
+		R"(^(?P<sign>-?))"
+		R"((?:(?P<yards>\d+\.?\d*)(?:yards|yard|yd))?)"
 		R"((?:(?P<feet>\d+\.?\d*)(?:feet|foot|ft|'))?)"
 		R"((?:(?P<inches>\d+\.?\d*)[-+]?)"
 		R"((?P<fraction>(?P<numerator>\d+)\/(?P<denominator>\d+))?(?:inches|inch|in|"|))?)"
 	);
 	QRegularExpressionMatch match = qreg.match(expr);
 	if (match.hasMatch()){
+		std::cout << "RS_Math:derationalize: matches: " << match.capturedTexts().join(", ").toStdString() << std::endl;
 		double total = 0.0;
-		QString num = (match.captured("numerator") != "") ? match.captured("numerator") : "0.0";
-		QString denom = (match.captured("denominator") != "") ? match.captured("denominator") : "0.0";
-		QString inches = (match.captured("inches") != "") ? match.captured("inches") : "0.0";
-		QString feet = (match.captured("feet") != "") ? match.captured("feet") : "0.0";
-		QString yards = (match.captured("yards") != "") ? match.captured("yards") : "0.0";
+		int sign = (match.captured("sign").isNull() || match.captured("sign") == "") ? 1 : -1;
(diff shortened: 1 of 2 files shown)


dellus:
Missing proper imperial architectural input has always been an issue and reason for complaints, as recently: https://forum.librecad.org/Units-displayed-when-inserting-dimensions-tp5718204.html
So I have investigated how it's handled in AutoCAD. Quote from AutoCAD 2009 User's Guide:

Enter Distances in Imperial Architectural Format
In architectural format, to enter feet and inches, indicate feet using the prime
symbol ('), for example, 72'3. You don't need to enter quotation marks (“) to
specify inches.  

I don't have access to AutoCAD, but have tested it with NanoCAD 5, which is available for free and claims to stick to AutoCAD conventions. It works there. I think it's worth considering to be compatible. Does your model support this?

Sleet01:
> Missing proper imperial architectural input has always been an issue and reason for complaints, as recently: https://forum.librecad.org/Units-displayed-when-inserting-dimensions-tp5718204.html
> So I have investigated how it's handled in AutoCAD. Quote from AutoCAD 2009 User's Guide:
> 
> Enter Distances in Imperial Architectural Format
> In architectural format, to enter feet and inches, indicate feet using the prime
> symbol ('), for example, 72'3. You don't need to enter quotation marks (“) to
> specify inches.
> 
> I don't have access to AutoCAD, but have tested it with NanoCAD 5, which is available for free and claims to stick to AutoCAD conventions. It works there. I think it's worth considering to be compatible. Does your model support this?

It does not currently, but it would be simple enough to make the inch (") mark optional when the foot mark (') is detected.  I'll write that up tomorrow.

clayreimann:
Any updates on this? Even without the optional `"` I think that this should be merged!

Sleet01:
> Any updates on this? Even without the optional `"` I think that this should be merged!

Sorry, I haven't worked on this in a bit.  Let me poke around and see if I got far enough to wrap it all up.

atorrey:
Another dimensioning option that I'd like is to have 'dual dimensions' so that you can have both imperial and metric dimensions on the same drawing (i.e. '1.0" (25.4mm)' or similar)

This would be really useful for drawings that will be shared so that they are useful regardless of what the viewers preferred system is.  I see it frequently on drawings I download from commercial sites.

It is also useful when doing things like trying to make a metric dimensioned part on an older CNC machine that is easier to program in imperial or vice versa...

------------------
Arthur Torrey - ***@***.***>
-------------------

dellus:
In 2.2.1-alpha-125 now if if you draw a dimension on a 100 feet length in architectural mode it reads 3 -15/16" instead of 8' - 4".
Seems a case of inches - mm confusion, factor 25.4 .

GSStnb:
"...a 100 feet length..." I suspect you meant 100 inches(?)  If it is 100 feet, that would be a bigger problem.

dellus:
Sorry, yes of course I meant 100 inches.

PR #1180, opened 2019-12-28
Fix saving to file of the block with inserts.

### LINKED ISSUE #943: Segmentation fault when saving block (rs_blocklist.cpp:229)
### Expected behavior
Open the save dialog window.

### Observed behavior
Segmentation Fault.

Open the attached dxf file, try to save the 'grouped_stacks' block to a file. [segfault_grouped_stacks_block.dxf.gz](https://github.com/LibreCAD/LibreCAD/files/1609212/segfault_grouped_stacks_block.dxf.gz)
I've tried to debug this issue and I've found that the segmentation fault occurs in the rs_blocklist.cpp:229 line:
```
for (RS_Block* blk: *list) {
	if (blk->getName() == name) {
		RS_DEBUG->print(RS_Debug::D_DEBUGGING, "RS_BlockList::find(): OK");
		return blk;
	}
	auto node = blk->getBlockList(); // <-- Segmentation fault
	if (!searched.count(node)) {
		searched.insert(list);
		nodes.push_back(blk->getBlockList());
	}
}
```
For an unknown reason this only happens with the grouped_stacks block (the others save just fine).

LibreCAD 2.2.0-alpha-244-g54269f5b git 54269f5b02c7890871873b8506d816f939554ca5.
Platform: ArchLinux x86_64. 
Compiled with GCC 7.2.1

baranovskiykonstantin:
Fixes #943

### DIFF
--- a/librecad/src/actions/rs_actionblockssave.cpp
+++ b/librecad/src/actions/rs_actionblockssave.cpp
@@ -63,6 +63,7 @@ void RS_ActionBlocksSave::trigger() {
         if(b) {
 			RS_Graphic g(nullptr);
             g.setOwner(false);
+            g.getBlockList()->setOwner(false);
 
            g.clearLayers();
 //           g.addLayer(b->getLayer());
--- a/librecad/src/lib/engine/rs_blocklist.cpp
+++ b/librecad/src/lib/engine/rs_blocklist.cpp
@@ -214,23 +214,9 @@ RS_Block* RS_BlockList::find(const QString& name) {
     }
 	// Todo : reduce this from O(N) to O(log(N)) complexity based on sorted list or hash
 	//DFS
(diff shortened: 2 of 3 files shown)


lordofbikes:
Many thanks for this fix!
For the sake of completeness, can you briefly explain the issue which was caused by the reverted commit from dxli?

baranovskiykonstantin:
I don't remember exactly what the issue was. But after looking over the changes I can note that the blocklist is drawing-wide, not block-wide.

```
		for (RS_Block* b: *list) {
			if (b->getName() == name)
				return b;
			auto node = b->getBlockList();                         <---
			if (!searched.count(node)) {
				searched.insert(list);
				nodes.push_back(b->getBlockList());
			}
```

PR #1181, opened 2019-12-28
Modifications: calculate borders of new entities

### LINKED ISSUE #810: Dimension lines and text disappear when zooming in
I recently noticed that the dimension lines and text disappear when I zoom into my drawing. This wasn't always the case, I used to zoom in and out freely, without fear of losing my dimensions. After playing around with the settings in my LibreCAD, I discovered that this was because the default dimensions settings are set to 'By Block' for the colour and width of the: text size and position; extension lines; and dimension lines, arrows, and ticks. Once I set these to 'By Layer' the dimensions and their related text stay visible when I zoom in.

However, these settings have to be updated in 'Current Drawing Preferences' every time I open LibreCAD. I don't know if this is a bug or if there is an easier way to fix the issue, but would appreciate some insight/ fixes.

I have included my version information below:

Version: 2.2.0-alpha-98-g7a4b5bb
Compiler: GNU GCC 6.1.1
Compiled on: Sep  6 2016
Qt Version: 5.7.0
Boost Version: 1.60.0

Thanks!

### LINKED ISSUE #1146: Copied text disappears when original zoomed/panned off screen.
I'm using Ubuntu 18.04 and having a problem with copied text. The copied text disappears when the original disappears from the screen as the result of zooming in.

In the same way the problem exists when panning: when the original is panned off the screen, the copy isn't shown any more on the screen... When I zoom or pan the original text back on the screen, the copied text reappears.

Version: 2.1.2
Compiler: GNU GCC 6.2.0
Compiled on: Sep 17 2016
Qt Version: 5.9.5
Boost Version: 1.62.0
System: Ubuntu 18.04.3 LTS

baranovskiykonstantin:
Fixes #810
Fixes #1146

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -2028,6 +2028,8 @@ void RS_Modification::addNewEntities(std::vector<RS_Entity*>& addList)
         }
     }
 
+    container->calculateBorders();
+
     if (graphicView) {
         graphicView->redraw(RS2::RedrawDrawing);
     }


lordofbikes:
Many thanks!

PR #1188, opened 2020-01-18
(WIP) Make PDF Print/Export work without having to use Print Preview first

### LINKED ISSUE #751: Printing into pdf produces empty page
When I try to print a drawing into a pdf without starting preview first I get an empty page. File size then is 1.2kB. It happens always. It is reproducible.

What I did, step by step:
1. load a dxf.
2. export to pdf -> empty page
3. print to pdf -> empty page
4. preview
5. print to pdf -> works fine
1. load another dxf
2. print to pdf -> empty page
3. preview
4. print to pdf -> works fine
5. load third dxf
6. preview
7. print to pdf -> works fine

martinvanzijl:
Fixes #751.

Exporting/Printing PDF files now works without having to use Print Preview first. Tested on Ubuntu 18.04.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -2259,6 +2259,8 @@ void QC_ApplicationWindow::slotFilePrint(bool printPDF) {
         return;
     }
 
+    graphic->fitToPage();
+
     statusBar()->showMessage(tr("Printing..."));
     QPrinter printer(QPrinter::HighResolution);
 


lordofbikes:
Many thanks for this contribution!
Sadly, this is not the complete solution. PDF export would be limited to fit to page.
If one would generate a PDF with specified scale, via print preview, this will not work with this fix anymore.

martinvanzijl:
@lordofbikes Thanks for your comment. I guess I should have tested more thoroughly!

I tried this since the call to `graphic->fitToPage();` was the only difference I could see in the code for "print preview" vs. "print/export".

Anyway, I think I see the real problem. If you just use "print/export" without doing a print preview first, it will export the area around the **center** of the diagram. If there are no elements near the center, this exports an empty page.

I attached a demo below to explain this (zipped OGV), done with master branch without this fix:

[export-problem-demo.zip](https://github.com/LibreCAD/LibreCAD/files/4705067/export-problem-demo.zip)

At least with print preview, it does a "fit to page" at the start, which means the diagram's elements will be included on page, regardless of whether they are near the center or not.

martinvanzijl:
As for how to fix the issue properly, I can suggest either:

1) Use fit-to-page if the user has not done Print Preview yet. If the user has done Print Preview before, use the settings from there instead. OR
2) **Always** show print preview before printing/exporting.

Anyway, feel free to close this branch.

lordofbikes:
Don't worry, this is a real issue indeed and your efforts are more than welcome.
I tend to your solution 1.
Then inexperienced users will have an output at least and often a fit to page print is enough.
Maybe a little hint message (about print preview) in the command history can help when `fitToPage()` was applied.
I think a missing drawing area is the cause for this issue. In `slotFilePrint()` we have access to `RS_Graphic* graphic`, I assume there is the drawing area and insertion point set from print preview.
So I think the solution is to check if `graphic` contains a valid area and if not, call `fitToPage()`.

If you can have a look at this, you're welcome. Then simply push modifications to this branch.
But if you can't, no problem, I'll keep this open until someone has battled this issue down.

dellus:
I have long since been of the opinion it would be better to always run printing and pdf/bitmap export via PrintPreview. I have recently discovered that QCAD does it this way, kind of. Maybe worth to have a look on it.
qcadppv

fa201:
I agree with @dellus. The process would be smoother and easier for newcomers.
We could have one print menu in which we set up the print/export and then print/export.

Salamek:
Still happening to me on 2.2.0... on Archlinux... i was getting white empty page when using File -> Export -> Export as PDF, when i did the same from print preview it worked ok...

dxli:
Thanks for working on this, @martinvanzijl, and for the analysis in this thread — it pointed at the fix.

Closing as superseded. Since ed79434d5, **Print...** and **Export as PDF** open Print Preview instead of printing when it isn't already open, and the preview fits and centers the drawing on the page when it opens. That is option 2 from the discussion above, and it avoids the problem @lordofbikes raised: with the fixed print scale option set, the preview keeps the drawing's scale and placement instead of fitting. This is in every 2.2.1 release and on master.

The `fitToPage()` call here would now run only from inside Print Preview, where it would discard the scale and placement set there, so it should not be merged.

PR #1195, opened 2020-02-24
Code cleanup and deprecated functions

pgeorgiev98:
The changes done here are mostly fixing deprecated usages of Qt functions and some basic changes that greatly reduce the number of warnings that show during compilation.


lordofbikes:
Merged manually.
Many thanks for this refreshment!

PR #1200, opened 2020-03-02
Fix for 'move to bottom' command with existing selection

### LINKED ISSUE #1040: Move to Bottom Command
### Expected behavior
Moving objects to top or bottom should work
### Observed behavior
I can move an object that is behind another to the top, I cannot move an object that is on top to the bottom. I also cannot raise over entity or lower after entity. What am I doing wrong?
To try to move an object to the bottom, I 1st select the top object, then  I press the "move to bottom" icon. Nothing will happen when I do this.
about

martinvanzijl:
Fixes #1040. The "Move to Bottom" command now works when there is an existing selection. Demo below:

move-to-bottom-demo


lordofbikes:
Many thanks for this fix!
I reopened #1200, please see my comment there.

lordofbikes:
I don't want to miss to praise you for those outstanding pull requests.

martinvanzijl:
@lordofbikes You're welcome!

PR #1201, opened 2020-03-04
Create equidistant polylines in currently selected layer

### LINKED ISSUE #682: "Create equidistant Polyline" should create the polyline in the current layer
The tool "Create Equidistant Polyline" creates a polyline in the same layer as the original polyline. This behavior is contrary to other tools that add a line/polyline to the drawing: the line offset tool, the concentric circle tool will add the new line/polyline to the current layer. Also, creating in the current layer is more flexible than letting it it the same layer (it adds one more degree of freedom). 

(The user can put the equidistant polyline in the current layer by changing it afterwards, but this is cumbersome as it involves manually selecting and manually acting on each single polyline everytime.)

martinvanzijl:
Fixes #682. Equidistant polylines are now created in the currently selected layer.

### DIFF
--- a/librecad/src/actions/rs_actionpolylineequidistant.cpp
+++ b/librecad/src/actions/rs_actionpolylineequidistant.cpp
@@ -154,7 +154,7 @@ bool RS_ActionPolylineEquidistant::makeContour() {
 
     for (int num=1; num<=number || (number==0 && num<=1); num++) {
         RS_Polyline* newPolyline = new RS_Polyline(container);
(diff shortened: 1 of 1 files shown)


lordofbikes:
Many thanks for this fix.
I recognized, that the layer of the equidistant was correct, but the properties was set fix.
E.g. when the current layer has a different color, line style or width, the equidistant still had the same look like the original. I think this is not the intention, when the equidistant is drawn on a different layer.
I fixed this with 0de519c.

martinvanzijl:
Thanks for the correction.

Issue #1162, opened 2019-11-23
Can't import SVG???


teo1978:
I have an SVG file and I want to import it.

I tried File -> Open but svg isn't among the allowed formats.

Under File -> Import there are two options: "Insert Image" and "Block". Confusing.

I tried "Insert Image" and selected the file but it does absolutely nothing. Anyway, I don't suppose that'w what I would have wanted (I don't want to insert an image, I want to import the vector drawing in the SVG).

So I tried "Block" but again, the available file types don't include SVG.

Is there seriously no way to import an SVG, or is it hidden somewhere not obvious?

TomaszPluta:
> I tried "Insert Image" and selected the file but it does absolutely nothing

It does :)  The problem is scale of image with regard  to actual zoom. 
After clicking "open" there is  "specify reference point" info showed. An yellow dotted lines will appear on screen. If you look at it - right and top line are also dotted with a white coulour - it is part of white rectangle which points to place where image should be inserted. When you zoom out, you are sure to see the whole rectangle. 

Maybe it is worth trying to do an auto-scale image to current zoom - I will take a look at it. I would  also  try to investigate possibility of opening svg  not only import the svg image.

![rectinsert](https://user-images.githubusercontent.com/12385009/76711115-77c73080-670d-11ea-9d02-ecd24e8d5b4a.png)

PR #1204, opened 2020-03-17
added auto-zoom after inserting an image

TomaszPluta:
Added auto-zoom after redrawing the inserted image in "RS_ActionDrawImage::trigger"  function. 
Due to "Can't import SVG???" #1162  - This does not completly resolve  author's request  (according to the project wiki SVG cannot be open), but now there should be no confusion ("where is my image?!")  after insertion.


lordofbikes:
Many thanks!

Issue #1186, opened 2020-01-05
Print Preview dropdown scales list 


dellus:
In Print Preview it is a nuisance to have to scroll through a long list of enlarging scales everytime until you get to the most used ones. I suggest to remove all these enlarging scales, like 1000:1, which hardly any user needs, except he is a watchmaker or IC circuit designer. Also some of the normal scales, preferably those more than 1:10000 could be omitted, as used only for special purposes. In the rare cases those exotic scales are needed they can be defined manually by input. 

All windows versions, in Linux it's a bit different/better.

lordofbikes:
I merged #1206 which make things a little better.
A final solution could be to make this totally customizable in application preferences, then every user can adjust scales for his convenience.

PR #1206, opened 2020-03-18
removed too big and too small scales

TomaszPluta:
Print scale list constrained due to "Print Preview dropdown scales list  #1186 ". Magnifications greater than 1000:1 and smaller than 1:10000 are removed.


lordofbikes:
Many thanks.

Issue #1215, opened 2020-04-14
Help: commands linkshould be updated.


fa201:
### Expected behavior
Menu **Help/On line/Commands** should redirect to https://librecad.readthedocs.io/en/latest/ref/tools.html

### Observed behavior
Menu **Help/On line/Commands** currently directs to deprecated Wiki.

### Operating System and LibreCAD version info
Version: 2.2.0-rc1-190-g51b21cc7
Compiler: GNU GCC 7.5.0
Compiled on: Mar 29 2020
Qt Version: 5.9.5
Boost Version: 1.65.1
System: Ubuntu 18.04.4 LTS

lordofbikes:
The link to the wiki is currently the best place with the command line commands, which is the intention behind this link.
The [Tools](https://librecad.readthedocs.io/en/latest/ref/tools.html) section in the manual has different meaning.
The commands should be migrated from the wiki to the manual section [The Command Line](https://librecad.readthedocs.io/en/latest/guides/cmdline.html?highlight=commands#the-command-line) or possibly added as new reference in the manual appendix?

fa201:
From my point of view, having 2 places in the documentation listing commands (one in the reference chapters and one in command line / appendix) is not right for the same reason than one should not duplicate parts of code...

Before opening this issue I thought the same way but it is also very natural to have the commands just besides the explanation...

lordofbikes:
Of course, you are right @fa201, my fault, sorry for the confusion.
I missed the command column when I took a quick look at the tools table.
Maybe its time to get some glasses :rofl:

fa201:
@lordofbikes there is no offense :-)
Thanks for the PR.

PR #1216, opened 2020-04-20
Update wiki links

TomaszPluta:
Wiki links updated due to  issue #1215


lordofbikes:
Many thanks for your efforts.
There are 4 new files in this PR which are not needed, possibly added by accident?
And not all links in your PR are correct.
I fixed this with the above mentioned commit 631d6ceb983fe8630fc5693b5f7cc7b080d024d0 and close this PR.
Issue #1215 is kept open to track the commands migration from old wiki to the manual.

PR #1224, opened 2020-05-20
Fix build with Qt 5.15 (missing QPainterPath include)

a17r:


lordofbikes:
Many thanks for this fix.

PR #1227, opened 2020-05-26
Update qg_commandedit.cpp

irsbugs:
# Allow comments in command files.

qg_commandedit.cpp has `void QG_CommandEdit::readCommandFile(const QString& path)` function to read command files. There is no provision to add comments in a command file. 

This request is to allows lines commencing with a hash ( # ) to be treated as a comment and ignored from being passed to the `processInput(line)` function.

Regards, Ian.


lordofbikes:
Many thanks!

Issue #962, opened 2018-03-21
Colors(black and white) don't switch while exporting


philectro:
### Expected behavior
While exporting from white on black work view, it should be automatically set to black on white while exporting.
### Observed behavior
The lines from the working view disappear while exporting to picture(white on white).
### Steps to reproduce or sample file
Creating something on black background, export with white background.
### Operating System and LibreCAD version info
All version tested on all OS tested,Ubuntu and Windows 10 CFU, 2.0.9, 2.1.3, 2.2 rc1

Note: sorry if this is not the right way to use the program, i don't report it for myself.

shawncurry:
Confirmed.  But it seems much worse than described above.  I'm having serious problems with any newly created drawings. Older drawing seem fine.  2.2 rc1

- File --> New
- Draw a rectangle
- File --> Export --> PDF
- Open it

![outofrange](https://user-images.githubusercontent.com/50385855/58364560-5cd8b680-7e84-11e9-94a3-22fedafc5d59.png)

- Save the DXF - Close/Open LibreCAD - Open DXF - ExportPDF
- Same result!
- Use the examples from Issue #1056 "Files created in LibreCAD do not open in AutoCAD"
- Exported PDF is fine

fa201:
In case it helps to hunt the bug, it works if you print the PDF instead of exporting it, at least with my Ubuntu.

Issue #553, opened 2015-08-09
export pdf creates blank page


r-a-v-a-s:
Export → Export as PDF
always produces a PDF that is a single blank white page.

dxli:
can not reproduce this bug

bursi:
I experienced this problem as well, and in my case, it was caused by a scaling issue. 

You should check the dimensions and units of your drawing as well as the paper size. 
(both can be set in "Edit > Current Drawing Preferences")

If you have a drawing which is a few meters wide and you have an A4 paper selected, you'll only get a small cut of it on your pdf - which may very well be empty. 

To fix it, click "File > Print Preview" and then choose an appropriate scale. Then directly click "File > Export > Export As pdf". it is essential that you do so directly without closing the print preview. 

This worked for me, the drawing in the pdf had the same scale as in the print preview.

r-a-v-a-s:
Thanks for the information. I was able to create a non-blank pdf.

r-a-v-a-s:
I'm wondering if a dialog to choose the paper size would be appropriate.
https://en.wikipedia.org/wiki/Principle_of_least_astonishment

dxli:
we can modify the graphicview part to detect blank paper for printing, and even the detailed information:

like empty page, which covers only 0.01% of the drawing range.

r-a-v-a-s:
That seems viable.

kspartin:
Here is a drawing that I cannot export as PDF. I have it set for Arch E. I select a 1:10 scale and Export-> Export as PDF. This worked in 2.7 without a problem. This drawing was originally drawn in 2.7.
[Supports for Tank Access Platform - Sht 2.zip](https://github.com/LibreCAD/LibreCAD/files/215609/Supports.for.Tank.Access.Platform.-.Sht.2.zip)

kspartin:
2.0.8 and 2.0.9 have this problem. 2.0.7.1 works perfectly.

kspartin:
Basically, what I have figured out is that the english (us) paper sizes don't work. The metric A sizes seem to work fine.

r-a-v-a-s:
I confirm that it creates a blank PDF using the latest code.
Something else to note is that Arche E is described as 35.984 X 47.992
instead of 36 x 48.

r-a-v-a-s:
@kspartin  I just added support for the Arch paper sizes.
I'm not sure how it worked for you in 2.0.7...
the code was equally broken in that version.

[...]

PR #1232, opened 2020-05-30
Fix argument of RS_GraphicView::setBackground in export

milan11:
Qt::black is an enum value and it caused RS_Color(unsigned int f) constructor to be called (but its int argument should be flags and not a color from Qt's enum). So basically the meaning of the int value was changed.

This change surely fixes this problem:
- draw a black line
- export as PNG
- Background: white
- Colouring: coloured
- problem: the exported PNG does not contain the line (it is blank)

This change may cause a fix in bugs #962 and #553 (or at least solve some confusion in these bugs).

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -2159,9 +2159,9 @@ bool QC_ApplicationWindow::slotFileExport(const QString& name,
 
 	RS_StaticGraphicView gv(size.width(), size.height(), &painter, &borders);
     if (black) {
-		gv.setBackground(Qt::black);
+		gv.setBackground(QColor(Qt::black));
     } else {
-		gv.setBackground(Qt::white);
+		gv.setBackground(QColor(Qt::white));
     }
     gv.setContainer(graphic);
     gv.zoomAuto(false);


lordofbikes:
This is a tough catch! Many thanks for finding this.

While your solution surely works, I'd rather prefer this in `rs_color.h`:
`RS_Color(const Qt::GlobalColor color) : QColor(color), RS_Flags() {}`
With your solution it is not ruled out, that the same error happens in future.
With the additional constructor `setBackground()` can be called with `Qt::GlobalColor` in future without issues.

If you want, you can change that in this branch to keep your share in this fix.
Otherwise I can commit it directly.

Have you tested the impact of this change in the mentioned issues?
With my tests I still couldn't export white lines on black background.
So can you confirm that it should work or is there possibly another issue?
Maybe with my setup only?

milan11:
Thank you very much for reviewing this.

Of course, your solution is better. Please feel free to disregard this pull request and commit your solution directly.

Unfortunately, I have not tested the mentioned issues. I just mentioned them in case they are somehow related, but I am not sure.

For me, white lines on black background do not work either, even after this fix.

More specifically - I tried to export PNG and I observed:

Before fix:
A. Bakcground: White, Colouring: Black / White - OK
B. Bakcground: White, Colouring: Coloured - WRONG
C. Bakcground: Black, Colouring: Black / White - WRONG
D. Bakcground: Black, Colouring: Coloured - OK

After fix:
A. Bakcground: White, Colouring: Black / White - OK
B. Bakcground: White, Colouring: Coloured - OK
C. Bakcground: Black, Colouring: Black / White - WRONG
D. Bakcground: Black, Colouring: Coloured - OK

After you mentioning the white lines on black background, I was afraid that this fix has broken some other scenario - that e.g. this obvious error in code was compensating some other error. But it is probably not the case:
- we fixed scenario B
- scenario C remained broken
- we did not break A or D

The while lines on black background (scenario C) is some other problem. After some debugging, I discovered the line
`lpen.setColor(RS_Color(0,0,0));`
in
`void RS_PainterQt::setPen(const RS_Pen& pen)`

Without that line it works correctly. It should somehow decide between black and white here, or leave white untouched (considering it a correct color).

lordofbikes:
Thanks for your validation.
I pushed the constructor and also a fix for #962.
Export b/w on black background, your scenario C, works now too.

Again many thanks for your efforts!

milan11:
I checked current master and all the scenarios (A, B, C, D) work correctly for me (I even tried with different colors). Thank you!

PR #1233, opened 2020-05-31
fix null pointer dereference in main window

v-tyrtov:
There is possibility of null pointer dereference
in some systems while opening or creating a file

Detected with wayland-based tiling WM (sway)

Signed-off-by: Vyacheslav Tyrtov <v.tyrtov@gmail.com>

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -1655,9 +1655,9 @@ void QC_ApplicationWindow::slotFileNewTemplate() {
     QMdiSubWindow* old=activedMdiSubWindow;
     QRect geo;
     bool maximized=false;
-    if(old ) {//save old geometry
-        geo=activedMdiSubWindow->geometry();
-        maximized=activedMdiSubWindow->isMaximized();
+    if(old != nullptr ) {//save old geometry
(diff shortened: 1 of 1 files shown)


lordofbikes:
Many thanks for this fix.

lordofbikes:
Not to myself: test even obviously correct pull requests thorough :wink: 

I recognized by chance next day, when LibreCAD was started from command line with a drawing to open, it crashes. The cause was found fast in this fix.
But this was finally caused by a not intialized pointer.
`activedMdiSubWindow` was garbage when assigned first time to `old`.
But it was initialized in some slot, on its way to `if (old)` and thus the old code worked.
With this fix the garbage pointer in `old` was called and it crashed on start up.
Now `activedMdiSubWindow` is initialized with `nullptr` and the issue is solved for me.

@v-tyrtov can you please confirm, that this has no negative impact with wayland.

v-tyrtov:
Hi, 
I was lucky not to catch the  same crash on my system even with uninitialized variables.
Anyway, your patch is Ok for me. Please note that I didn't carry out thoughtful testing - just opened several drawing from command line and from UI.

So it's ok for wayland-based system.

lordofbikes:
Great, many thanks @v-tyrtov for testing this.

PR #1234, opened 2020-05-31
libdxfrw: fix spline parsing

v-tyrtov:
Splines were parsed incorrectly when weight == 0 - fixed

weight type was changed to int to print it correctly during debug
(there is no DRW_dbg::print(unsigned char) and std::string was used)

Signed-off-by: Vyacheslav Tyrtov <v.tyrtov@gmail.com>


lordofbikes:
Many thanks!

Issue #1256, opened 2020-07-04
Increase visibility of lines on similarly colored backgrounds


JeremyRuhland:
### Expected behavior
Objects who's color match closely with the background color should have their colors modified for improved visibility.
For example, black lines drawn on a `#1f2730` background (default autocad background color) should appear white for visibility. This is the normal behavior in most CAD programs.

### Observed behavior
Librecad intelligently changes line colors to black or white for improved visibility only when the line color exactly matches the background color.
For example, magenta lines on a magenta background become black, black lines on a black background become white.
However, black lines on a `#000001` background remain black despite the background color appearing black to the human eye.

### Proposal
Globally changing line color behavior would probably be disruptive for users who expect the line color shown on screen to match their selected line color. Such a change would probably not be desirable and would definitely require a option in the application settings to turn off. I think a reasonable option for now would be to implement fuzzy color matching only on black lines.

I don't have a deep understanding of C++, QT or the codebase but I've been searching for the part responsible for color checking. It looks like its occurring here:
https://github.com/LibreCAD/LibreCAD/blob/9f2c32dd4b9dc2ed5e028b682473b098c48ad3e0/librecad/src/lib/gui/rs_graphicview.cpp#L1026-L1029

Foreground color is set by this function according to the luminosity of the background:
https://github.com/LibreCAD/LibreCAD/blob/9f2c32dd4b9dc2ed5e028b682473b098c48ad3e0/librecad/src/lib/gui/rs_graphicview.cpp#L1761-L1770

I propose implementing a color difference checker and modifying the above if statement in `setPenForEntity` to add an additional behavior: If the pen color is black and the difference in color between it and the background falls below some critical human perception limit then the pen color will switch to the foreground color.

### Operating System and LibreCAD version info
Version: 2.1.2
Compiler: GNU GCC 6.2.0
Compiled on: Sep 17 2016
Qt Version: 5.9.5
Boost Version: 1.62.0
System: Ubuntu 18.04.4 LTS
Kern: 4.15.0-76-generic

PR #1257, opened 2020-07-04
Enhance visibility of black lines on dark backgrounds

JeremyRuhland:
Implementing functionality proposed in #1256 
Program compiled and tested as working.

### DIFF
--- a/librecad/src/lib/engine/rs_color.cpp
+++ b/librecad/src/lib/engine/rs_color.cpp
@@ -1,4 +1,5 @@
 #include <iostream>
+#include <cmath>
 #include "rs_color.h"
 
 //This method is used for plugins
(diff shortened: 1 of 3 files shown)


dellus:
If someone deliberately wants to use white lines or white solid hatches, for instance above an other coloured area, or to hide parts of a drawing with a white cover, this is prevented by the automatic black / white swap. Up to now there is the work around to use a slightly off nearly white, like RGB 255,255,254.
The fuzzy colour filter you propose prohibits this, doesn't it?

JeremyRuhland:
I'm not sure I fully understand the situation you're describing. The fuzzy filter only affects lines set to black (it's kind of confusing nomenclature that black and white are the same color internally).
The work around you describe would still function as intended.

On July 6, 2020 2:17:45 PM PDT, dellus <notifications@github.com> wrote:
>If someone deliberately wants to use white lines or white solid
>hatches, for instance above an other coloured area, or to hide parts of
>a drawing with a white cover, this is prevented by the automatic black
>/ white swap. Up to now there is the work around to use a slightly off
>nearly white, like RGB 255,255,254.
>The fuzzy colour filter you propose prohibits this, doesn't it?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/LibreCAD/LibreCAD/pull/1257#issuecomment-654470257

-- 

Jeremy Ruhland
jeremy@goopypanther.org

dellus:
Ok, I have got it wrong. So my issue is a different topic.
Just for your interest, to clear what I was talking about, a sample file and how it should look printed:

[offwhitesample.zip](https://github.com/LibreCAD/LibreCAD/files/4884696/offwhitesample.zip)

lordofbikes:
Many thanks!

PR #1292, opened 2020-11-25
Fix typos

luzpaz:
Found via `codespell -q 3 -S *.ts,./libraries,*.dxf,*.lff -L ans,atleast,ba,lightyear`


lordofbikes:
Many thanks!

PR #1314, opened 2021-01-29
cppcheck: Same expression on both sides of '&&'

serval2412:


lordofbikes:
Many thanks for this fix!

Issue #1282, opened 2020-09-28
Impossible Condition 


maddin200:
[librecad/src/ui/forms/qg_polylineequidistantoptions.cpp line 116

void QG_PolylineEquidistantOptions::updateNumber() {
	if (action) {
		const QString& l = ui->leNumber->text();
		bool ok;
		unsigned i=static_cast<unsigned>(RS_Math::eval(l,&ok)+0.5);
		if (!ok || (i==0 && i>100))   // <-- impossible
			i=1;
		ui->leNumber->setText(QString::number(i));
		action->setNumber(i);
	}
}

DuncanLithgow:
Could you maybe explain a bit what this bug report is about?

maddin200:
Hi, 
The variable i can't be  == 0 and at the same time > 100  ( (i == 0) && (i >100) ); might be logical or  required

fa201:
@lordofbikes  Can you  check if this condition correction can be implemented in next release ? Thanks.

lordofbikes:
I wonder, how you found this, it should have been detected by static analysis.
Many thanks and great praise.
For curiosity's sake I looked for the source of this issue and I found it back in the first commit 2012, seems to be original QCAD code.

PR #1315, opened 2021-01-29
cppcheck: Logical conjunction always evaluates to false

serval2412:


lordofbikes:
Many thanks for this fix and sorry for missing it, when I fixed #1282.
So this was solved with 8b2b62d1a3fbd2f5510c6cd5bc9502d554c03004

PR #1320, opened 2021-02-06
manual changes

vuphin:


librecad/src/lib/gui/rs_graphicview.cpp
```diff
@@ -1374,8 +1374,19 @@ void RS_GraphicView::drawRelativeZero(RS_Painter *painter) {
...
+    painter->drawLine(RS_Vector(vp.x-zr, vp.y-zr),
+                    RS_Vector(vp.x+zr, vp.y-zr)
+                      );
+    painter->drawLine(RS_Vector(vp.x-zr, vp.y-zr),
+                      RS_Vector(vp.x-zr, vp.y+zr)
+                      );
+    painter->drawLine(RS_Vector(vp.x+zr, vp.y-zr),
+                      RS_Vector(vp.x+zr, vp.y+zr)
+                      );
+    //painter->drawCircle(vp, 5);
```
vuphin, line 1389: The relativeZero is rather square than circle

lordofbikes:
Many thanks for your efforts and your intention to share them.
Sadly there are several issues with your pull request.

I see the intention of your changes, but they base on your personal taste.
It is no good practice to hard code visual changes and annoy all the users who are happy with the current state.
So the approach is to keep the current design by default and implement alternative designs which could be used through settings in application preferences.
I also recommend to get in touch with the community in our forum or chat channels, then you can clarify your ideas with experienced users and estimate the chances for an accepted pull request.

Concerning the quality:
* use separate branches for different topics, that makes review easier for us
* with separate branches you can also give more expressive titles to pull requests
* avoid to work on your forks `master` branch, that makes it easier for you to keep in sync with the master repo
* try to test better before you commit, you have 10 commits for 3 topics, that makes review and merge more work for us
* use `git diff` before committing, then you can avoid to commit accidentally made changes

I'll add some comments in the code to clarify these points.

librecad/src/lib/gui/rs_linetypepattern.cpp
```diff
@@ -72,4 +72,4 @@ const RS_LineTypePattern RS_LineTypePattern::patternBorderLine2{{6.0, -3.0, 6.0,
 const RS_LineTypePattern RS_LineTypePattern::patternBorderLineX2{{24.0, -8.0, 24.0, -8., 0.2, -8.}};
 
 const RS_LineTypePattern RS_LineTypePattern::patternBlockLine{{0.5, -0.5}};
-const RS_LineTypePattern RS_LineTypePattern::patternSelected{10.0};
+const RS_LineTypePattern RS_LineTypePattern::patternSelected={10.0};
```
lordofbikes, line 75: The first assignment already worked. When you are in doubt about the correct or preferred syntax, don't hesitate to ask.
The preferred place for code related discussions is on our [Zulip chat](https://librecad.zulipchat.com)

librecad/src/src.pro
```diff
@@ -285,6 +285,7 @@ SOURCES += \
     lib/gui/rs_graphicview.cpp \
     lib/gui/rs_grid.cpp \
     lib/gui/rs_linetypepattern.cpp \
+    lib/gui/rs_linetypepattern.cpp \
```
lordofbikes, line 288: accidentally changed?
This is wrong and could be avoided when using `git diff` before committing.

librecad/src/lib/gui/rs_graphicview.cpp
```diff
@@ -1374,8 +1374,19 @@ void RS_GraphicView::drawRelativeZero(RS_Painter *painter) {
...
+    painter->drawLine(RS_Vector(vp.x-zr, vp.y-zr),
+                    RS_Vector(vp.x+zr, vp.y-zr)
+                      );
+    painter->drawLine(RS_Vector(vp.x-zr, vp.y-zr),
+                      RS_Vector(vp.x-zr, vp.y+zr)
+                      );
+    painter->drawLine(RS_Vector(vp.x+zr, vp.y-zr),
+                      RS_Vector(vp.x+zr, vp.y+zr)
+                      );
+    /**painter->drawCircle(vp, 5);*/
```
lordofbikes, line 1389: This needed 4 commits.
Think twice and try to test more carefully before committing.

librecad/src/lib/gui/rs_linetypepattern.cpp
```diff
@@ -29,13 +29,13 @@
...
 {
     totalLength=0.;
-	num = pattern.size();
-	for(double const& l: this->pattern){
-		totalLength += fabs(l);
-	}
+    num = pattern.size();
+    for(double const& l: this->pattern){
+        totalLength += fabs(l);
+    }
```
lordofbikes, line 38: Avoid to change formatting and whitespaces when it is not necessary.

PR #1325, opened 2021-03-02
List entities plugin: Add text insertion point and text content to its output

perepujal:
Currently the Height value of a point imported via the asciifile plugin only could be stored as a text near the point.
This commit would allow the List Entities plugin to retrieve the point of insertion of any text plus the text contents.
That would make possible to, via an apropiate script, correlate a point with its text height value.
HTH
Pere


lordofbikes:
Many thanks! :+1:

PR #1343, opened 2021-04-30
Widget spacer

GSStnb:
When using the dock areas (specifically the left area) for dock widgets, the widgets expand to fill the available area and results in the icons being spread out over the dock area.  Adding additional dock widgets will fill the area and reduce the amount of space between icons, but may require more dock widgets than are available (notably on HiDPI displays).

This PR adds an empty 'spacer' widget to the left dock area.  The added widget has no icons or any other decorations / features, but is simply to fill the empty space within the left dock area.  The spacer widget can be resized to prevent the other docks from expanding and spreading out the icons.


GSStnb:
Found a flaw to be addressed - the filler dock disappeared and there is no way to bring it back (?)

PR #1345, opened 2021-05-04
add boost tuple include to fix build

jirislaby:
With boost 1.76, we see:
```
lib/engine/rs_ellipse.cpp:70:15: error: 'tuple' in namespace 'boost::math' does not name a template type
   70 |  boost::math::tuple<double, double, double> operator()(double const& z) const {
      |               ^~~~~
```
Fix this by including the proper boost header.


lordofbikes:
Many thanks!
Perfect fix.

Issue #690, opened 2016-03-23
New symbols for "unicode.lff"


Symbian9:
There are few ukrainian symbols for "unicode.lff"
<code>[0404] Є
5,0;2,0;-0,2,A-0.41;0,7;2,9,A-0.41;5,9
3,5;0,5

[0406] І
0,0;0,9.

[0407] Ї
1.5,0;1.5,9.
0,11.5;0,12.
3.,11.5;3.,12.

[0454] є
4,0;1.5,0;-0,1.5,A-0.41;0,4.5;1.5,6,A-0.41;46
2.5,3;0,3

[0456] і
0,0;0,6.
0,8.5;0,9.

[0457] ї
1.5,0;1.5,6.
0,8.5;0,9.
3.,8.5;3.,9.

[0490] Ґ
0,0;0,9;5,9;5,10

[0491] ґ
0,0;0,6;4,6;4,7</code>
Some of them are similar to already presented symbols, so its possible paste next as alternative
<code>[0404] Є
5,0;2,0;-0,2,A-0.41;0,7;2,9,A-0.41;5,9
3,5;0,5

[0406] І
C0049

[0407] Ї
C00cf

[0454] є
4,0;1.5,0;-0,1.5,A-0.41;0,4.5;1.5,6,A-0.41;46
2.5,3;0,3

[0456] і
C0069

[0457] ї
C00ef

[0490] Ґ
0,0;0,9;5,9;5,10

[0491] ґ
0,0;0,6;4,6;4,7</code>
Add them to https://github.com/LibreCAD/LibreCAD/blob/master/librecad/support/fonts/unicode.lff

dxli:
@Symbian9 

Thanks a lot!

Can you send in a pull request?

Symbian9:
Cant make a pull, because I have only Symbian phone that can't fully work with github...

Evil-Spirit:
There are may be a bug. Missed newlines.

Symbian9:
> There are may be a bug. Missed newlines.

@Evil-Spirit, is it still an issue here?

Symbian9:
@dxli @lordofbikes, Please, check comments by @Evil-Spirit on some possible issues with `є` and `Ї` characters in `unicode.lff`:

- https://github.com/LibreCAD/LibreCAD/commit/ca6f39feb916390596c7ed3a0cfd88e935cd4e74#commitcomment-17007934
[...]

PR #1350, opened 2021-05-21
Fix Ukrainian characters LFF markup in unicode.lff

Symbian9:
See https://github.com/LibreCAD/LibreCAD/issues/690#issuecomment-206942438


phkahler:
@Symbian9  Why the removal of 0406 and 0456 and 0457?  I'm not sure removing glyphs is a good idea.  It also looks like some glyphs were in the original file twice, is that correct? I'm also curious about the changes making a multi-line entry from a single-line one without any visible change. Did a tool do that automatically?

I'm not sure why, but I'm stuck on this idea of having the two projects use the exact same version of this font ;-)

Symbian9:
>  Why the removal of 0406 and 0456 and 0457?

@phkahler, I'm not removed it!

I just removed duplicates of *raw code* & left only *reference link to similar character* instead.

Symbian9:
@lordofbikes, please, merge this and https://github.com/LibreCAD/LibreCAD/pull/1365 which are just brings few fixes & additions to `unicode.lff` font.

Symbian9:
@lordofbikes, any news on this pull merging status?

lordofbikes:
@Symbian9 focus is currently on final release of 2.2.0.
The build systems are revamped to create all needed packages for Windows, MacOS and Linux.
Before the final release I'll check all pull request and merge what is safe without impacting stability.

lordofbikes:
Many thanks!

Issue #1353, opened 2021-06-01
Incorrect import of DXF file with large Hatch


johnthagen:
### Expected behavior

The following AutoCAD DXF 2013 file: [dxf-2013-hatch.dxf](https://github.com/LibreCAD/LibreCAD/files/6580099/dxf-2013-hatch.dxf.txt)

Renders correctly within eDrawings Viewer 29.2.0.0029 on macOS 10.15.7:

<img width="1579" alt="Screen Shot 2021-06-01 at 3 15 27 PM" src="https://user-images.githubusercontent.com/10340167/120378299-3d36bd00-c2ec-11eb-9a1f-dd3a05a12481.png">

And also in Autodesk DWG TrueView 2022 on Windows 10:

<img width="1527" alt="Screen Shot 2021-06-01 at 3 23 50 PM" src="https://user-images.githubusercontent.com/10340167/120379220-715ead80-c2ed-11eb-8459-d38142c085ef.png">

But in LibreCAD 2.2.0-rc2 on macOS 10.15.7:

There are significant rendering artifacts:

<img width="2028" alt="Screen Shot 2021-06-01 at 3 25 39 PM" src="https://user-images.githubusercontent.com/10340167/120379372-a9fe8700-c2ed-11eb-9e9b-cd2ece21f7f2.png">

### Operating System and LibreCAD version info

```
Version: 2.2.0-rc2
Compiler: Clang 10.0.0
Compiled on: Dec 31 2020
Qt Version: 5.14.2
Boost Version: 1.71.0
System: macOS 10.15
```

pollywinters:
Applying PR #1354, I find that there is an error produced in file librecad/src/src.pro
The line `lib/gui/rs_linetypepattern.cpp \` becomes duplicated, resulting in build failure.
Removing the duplicate of the line allows the build to complete. The sample DXF file provided by https://github.com/johnthagen will then open and display correctly.
Note added to PR #1354

PR #1354, opened 2021-06-05
Fix issue #1353

vuphin:
remove only a line to fix


johnthagen:
Cross referencing issue #1353

jonadem (commented):
Is commit 0f5224954f5a47363d6339317c9317515f58bcca really needed ?

At my side, it doesn't want to compile with this.
I removed this duplicated line and afterwards, it builded just fine.

jonadem:
This is a file from Autocad opened with the code from master:
LC_master
This is with the pull request:
LC_vuphin

This is a great improvement !

pollywinters:
Applying this pull request, I find that there is an error produced in file librecad/src/src.pro, as per the comment by [jonadem](https://github.com/jonadem) above dated 26 Jun 2021.
The line `lib/gui/rs_linetypepattern.cpp \` becomes duplicated, resulting in build failure.
Removing the duplicate of the line allows the build to complete. The sample DXF file provided by [johnthagen](https://github.com/johnthagen) will then open and display correctly.
Note added to issue #1353

librecad/src/lib/engine/rs_hatch.cpp
```diff
@@ -607,7 +607,7 @@ void RS_Hatch::draw(RS_Painter* painter, RS_GraphicView* view, double& /*pattern
 //                    if (! (pa.size()>0 && (pa.last() - pt1).manhattanLength()<=2)) {
 //                        jp<<pt1;
 //                    }
-                    if(pa.size() && (pa.last()-pt1).manhattanLength()>=1)
+                    //if(pa.size() && (pa.last()-pt1).manhattanLength()>=1)
```
lordofbikes, line 610: This condition should avoid to store points twice. A container containing a bunch of lines have end points matching with the start point of the following entity. Storing all points will double the size of the resulting polygon.
So the condition has its right to exist, but it failed on the very first start point which led to the issue.

lordofbikes:
There was too many noise in this PR, 5 files changed for 1 line removed. And it also contained an error in the project file.
Also on reviewing I found out, that removing the if condition was not the correct solution.
So I fixed it now with https://github.com/LibreCAD/LibreCAD/commit/5480eba6af927db91ca43338815b40fdd4d3130b.

Many thanks anyhow for pointing this out and your effort to battle it down.

Issue #1349, opened 2021-05-20
Add new glyphs to the vectorfont from Solvespace


phkahler:
Solvespace uses the vector font from LibreCAD and has had some character added since then:
https://github.com/solvespace/solvespace/issues/1042#issuecomment-844820653

Some Ukrainian characters as well as the Depth, Counterbore, and Countersink commonly used to annotate holes.

I can make PR to merge these into the current LibreCAD font file, or we could go the other way and merge LibreCAD changes into solvespace and give them back. It seems like a good idea to share these updates between projects. If we get back in sync I'll try to remember to submit changes to both if/when they happen again.

Symbian9:
JFTR, I'm an author of those *"Some Ukrainian characters"* (i.e.`Ґ`, `ґ`, `І`, `і`, `Ї`, `і`, `Є`, `є`), which are already merged in LibreCAD (and then was merged to SolveSpace too!):

- https://github.com/LibreCAD/LibreCAD/issues/690

But @Evil-Spirit later found small issue with some of those characters (e.g. `є` and `Ї`):

- https://github.com/LibreCAD/LibreCAD/commit/ca6f39feb916390596c7ed3a0cfd88e935cd4e74#commitcomment-17007934
- https://github.com/LibreCAD/LibreCAD/commit/ca6f39feb916390596c7ed3a0cfd88e935cd4e74#commitcomment-17007980

And there was an extra patch to `є` charater added in SolveSpace added:

- https://github.com/solvespace/solvespace/commit/9c08398f51d5040245001c53ccd8388107221216#diff-e99cc430914390ed01b41b9eeefd419ec4c099e6b15d4fd5b29612e21cb8cfa0

So, ***extra check both*** versions of `unicode.lff` (SolveSpace's `master` & LibreCAD's `master`) before merge each to another or vice versa!

app4soft:
@dxli This issue could be closed as solved.

PR #1363, opened 2021-07-11
Update some glyphs. Add symbols for unicode 2334, 2335, 21A7 for Coun…

phkahler:
…terbore, Countersink, and depth. usefult for hole annotaions.

Someone should review the diff on this. I replaced the font file with the one from solvespace thinking our version was ahead, but there seem to be some deletions and incidental changes. Looking at the commit clearly shows which glyphs have changed, but other than the 3 listed in the PR I don't know why those differences exist.

Addresses #1349


phkahler:
After looking at this more closely, I think I should just add the additional glyphs to the LibreCAD version of this font and then pull the result back over to SolveSpace to unify them. Having the the github display of the diff was helpful in reaching that conclusion.

PR #1367, opened 2021-07-28
fix: increase maximum sizes of statusbar widgets to accommodate hidpi

gwhitney:
Since all of these widgets shrink to fit their content anyway, there is
  really no downside to larger maximum sizes, and they prevent the text
  content of the widgets from being clipped on hidpi displays.


lordofbikes:
Many thanks!

PR #1382, opened 2021-09-21
fix: now it is possible to cancel the close process.

hawk2411:
Hi,

I found a bug about closing an MDI window. The File Close dialog always closed the current window even if the user clicked the cancel button. If the user opened a drawing a few times, added something, closed the drawing without saving and canceled the closing, the whole application crashed.

I think my changes fix the bug.

### DIFF
--- a/librecad/src/main/qc_mdiwindow.h
+++ b/librecad/src/main/qc_mdiwindow.h
@@ -115,7 +115,7 @@ public:
     bool has_children();
 
 signals:
-    void signalClosing(QC_MDIWindow*);
+    bool signalClosing(QC_MDIWindow*);
 
 protected:
     void closeEvent(QCloseEvent*);
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ LibreCAD.pbproj/
 LibreCAD.dmg
 Info.plist
 
+#CLion internal directory
+.idea/
+
 # globs, anywhere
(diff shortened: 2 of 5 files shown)


librecad/src/main/qc_mdiwindow.h
```diff
@@ -115,7 +115,7 @@ public slots:
     bool has_children();
 
 signals:
-    void signalClosing(QC_MDIWindow*);
+    bool signalClosing(QC_MDIWindow*);
```
lordofbikes, line 118: [Qt documentation](https://doc.qt.io/qt-5/signalsandslots.html#signals) is very clear about signals:
```
They can never have return types (i.e. use void).
```
I know that moc and compiler accept this and that it basically works under certain conditions.
But for LibreCAD, which is build by many people, with different operating systems and a bunch of different compilers and Qt versions we should avoid this.

I have to take a more detailed look to understand the issue and your solution to possibly find a more robust and compatible solution.

hawk2411: Ok you right. There some other possible solution: 

1. Refactor the handling of closing a MDI window; ask the user in MDI and if he still want close the window emit a Save signal to the Application instance
2. add a second parameter to signal and  slotFileClosing of type QCloseEvent pointer and send it to Application instance, so it can handled it correct there.

PR #1383, opened 2021-09-22
fix: now it is possible to cancel the close process.

hawk2411:
Hi.

this is a second solution with void signal functions. ;-)


lordofbikes:
Sadly, using a pointer parameter addresses the same issue as using a return value.
See my latest https://github.com/LibreCAD/LibreCAD/issues/1403#issuecomment-1146577534.
Anyhow many thanks for your efforts :+1:

PR #1390, opened 2021-10-15
In Arch mode hide feet when 0

ndeubert:
Simply just don't print out feet when they are 0. I searched around in Architectural standards documents and didn't find anything saying it was required even when feet was 0. The only thing I couldn't figure out how to test was creating a negative dimension, so if that is still possible please let me know and I will try it, otherwise I can clean up the logic too.


ndeubert:
When do you expect to merge this?

PR #1392, opened 2021-10-17
Set text height in textoptions

### LINKED ISSUE #1375: [Feature request] Edit text size along the top bar
When I run the `text` command, I can change the actual text and angle on the fly.

I'd like to also be able to edit the text height in the same way, rather than having to cancel the text command and re-run it in order to edit the size.

Here's the bar I mean:

image

Version: 2.2.0-rc2
Compiler: GNU GCC 7.3.0
Compiled on: Dec 31 2020
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 10 (10.0)

MarcelloTheArcane:
Fixes #1375

This adds an input box to the top `qg_textoptions.ui` that allows editing the text height while keeping the text box active.

I've added it to `qg_mtextoptions.ui` too.

image

### DIFF
--- a/librecad/src/actions/rs_actiondrawmtext.cpp
+++ b/librecad/src/actions/rs_actiondrawmtext.cpp
@@ -294,5 +294,13 @@ double RS_ActionDrawMText::getAngle() {
 	return data->angle;
 }
 
+void RS_ActionDrawMText::setHeight(double h) {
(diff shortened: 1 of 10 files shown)


dxli:
already implemented in master(2.2.2)

PR #1406, opened 2021-11-16
save block to file, fixes #1400

### LINKED ISSUE #1400: Feature request: save block to file behaviour
The current process for saving a block to a file is counterintuitive and results in loss of data 

In CAD systems other than LibreCad, clicking File>Save will save the object that is _currently open in front of the user_, not other objects that are open in other workspaces, and therefore invisible to the user. Some systems have a "Save all" option, but there is usually a warning about unexpected behavior.

The block save process using the "Save active block to file" widget is detailed in the documentation, but RTFM is not a substitute for good UI design.

### Expected behavior

Clicking "File>Save" prompts for a filename and then saves the block that is currently open in the editor to disc as a regular drawing file.

### Observed behavior

Currently, the _only_ way to save a block to a file is to use the "Save active block to file" widget, which is not obvious nor intuitive to most users familiar with other CAD systems. Alternatively, the user may delete everything from the drawing but the block they wish to save, explode the block and then remove all blocks from the drawing before saving the drawing (containing only the block) to a file.

Using "File>Save" instead of the widget opens a "File>Save as" requestor, as the user would expect, but instead of saving the block that is currently open in the editor window, the current drawing (NOT the current active block) is saved instead. If the user, believing they are saving the block currently open in the editor window, chooses to overwrite an existing  block, the block file is overwritten with the current drawing. The drawing filename is also replaced with the nominated block filename, so once out of block edit mode and back into the main drawing, File>Save will not write the drawing back to the original file name.

### Suggested fix

1. File>Save should save only the block that is currently presented to the user (i.e open in the editor).
2. Alternatively, the "File>Save" and "File>Save as" menu items could be greyed out while edititng a block. 

Given that it is possible to open multiple drawings and then switch between them to edit them, and that a block that is open in an editor window behaves in every other respect exactly like a drawing, option 1 is the preferable one.

    - Right-click a block in the block list and select "edit", or click "Edit Block" icon
    - The selected block is opened in the editor window.
    - click "File>Save" or "File>Save as" to save the block
    - A "File>Save as" requestor appears in both cases. Entire main drawing is saved to the block file on disc instead of just the block.
    - Close block edit window to return to main drawing
    - Click File>Save
     -File is saved to the block filename on disc instead of the main drawing filename

Version: 2.1.3
Compiler: GNU GCC 9.3.0
Compiled on: Mar 23 2020
Qt Version: 5.12.8
Boost Version: 1.71.0
System: Ubuntu 20.04.3 LTS

melwyncarlo:
* Solved issue #1400

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -28,7 +28,7 @@
 **********************************************************************/
 
 // Changes: https://github.com/LibreCAD/LibreCAD/commits/master/librecad/src/main/qc_applicationwindow.cpp
-#include <iostream>
+
 #include "qc_applicationwindow.h"
 
 #include <QStatusBar>
@@ -106,6 +106,8 @@
 #include "lc_penwizard.h"
 #include "textfileviewer.h"
 #include "lc_undosection.h"
+
+#include "rs_actionblockssave.h"
 
 #include <boost/version.hpp>
 
@@ -1920,9 +1922,6 @@ void QC_ApplicationWindow::slotFileSave() {
 		recentFiles->updateRecentFilesMenu();
 }
 
-#include "qg_actionhandler.h"
-#include "rs_actionblockssave.h"
-#include <iostream>
 
 /**
  * Menu file -> save as.
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -28,7 +28,7 @@
 **********************************************************************/
 
 // Changes: https://github.com/LibreCAD/LibreCAD/commits/master/librecad/src/main/qc_applicationwindow.cpp
-
+#include <iostream>
 #include "qc_applicationwindow.h"
 
 #include <QStatusBar>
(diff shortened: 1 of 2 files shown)


librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -993,6 +1002,11 @@ void QC_ApplicationWindow::slotWindowActivated(QMdiSubWindow* w) {
     activedMdiSubWindow=w;
 
     QC_MDIWindow* m = qobject_cast<QC_MDIWindow*>(w);
+    if (w->windowTitle().contains("Block")) {
```
lordofbikes, line 1005: This is highly insecure.
- The title is set with `tr("Block '%1'")`, thus in translated environment a block window title must not contain "Block"
- When the drawing name itself contains "Block" we'll have wrong interpretation too

lordofbikes:
Beside my comment about insecure block window recognition, the implementation is based on hijacking `slotFileSaveAs()`. This is not a clean solution.
When this feature is implemented at all, the existing mechanism, using `m_map` with a new `QAction`, a separate menu entry and a separate slot for 'Save block as...' should be used.
Also I'm still unconvinced if this is really needed, only because one user didn't find the clearly visible 'Save block' button in the block list.

dxli:
feature already implemented

PR #1407, opened 2021-11-16
snap on middle fails for last point on spline, fixes #1395

### LINKED ISSUE #1395: spline through points: snap on middle point fails for last one, when more than one middle point is used
### Expected behavior
Using *snap on middle point* with e.g. 3 middle points, must snap on 3 points of the *spline through points* entity.

### Observed behavior
Only the first 2 points are snapped, the last one is missed.
It is always the last one missing, when more than 1 middle points are used.

* start a new drawing
* create a random *spline through points*
* switch off all snaps, except middle points
* raise the number of middle points in the tool option bar above 1
* use another tool, e.g. draw angle line, and see how it snaps to the spline

OS doesn't matter, tested with latest master

[Reported on forum](https://forum.librecad.org/Snap-midpoints-on-a-spline-tp5721141.html)

melwyncarlo:
* Solved issue #1395

The code talks about _splines_, but bears the equation of a _quadratic Bézier curve_.
Also, there is an analytic equation (which is generic) to compute the n-th degree Bézier curve, 
but the author of the code uses a very different approach (which is not compatible with the 
aforementioned analytic equation).

### DIFF
--- a/librecad/src/lib/engine/lc_splinepoints.cpp
+++ b/librecad/src/lib/engine/lc_splinepoints.cpp
@@ -999,7 +999,9 @@ RS_Vector LC_SplinePoints::GetSplinePointAtDist(double dDist, int iStartSeg,
 
 	if(dDist <= dQuadDist)
 	{
-		double dt = GetQuadPointAtDist(vStart, vControl, vEnd, 0.0, dDist);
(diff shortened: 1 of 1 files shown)


lordofbikes:
many thanks :+1:

PR #1409, opened 2021-11-16
dimension lines didn't change layer when text alignment is horizontal, fixes #1396

### LINKED ISSUE #1396: Dimension lines do not move to new layer when text alignment is set to Horizontal
When Text Alignment is set to Horizontal, moving a Dimension from one layer to another does not move the entire dimension to the new layer.  The dimension's text, arrows and extension lines are moved but its dimension line remains on the current layer.

This bug does not occur if Text Alignment is set to Aligned, only when Text Alignment is set to Horizontal.

Steps to reproduce:

Go to Options - Current Drawing Preferences - Text Alignment.
Set the Text Alignment to Horizontal
Create a Dimension.  It can be Aligned, Linear, Horizontal, Vertical or Diametric.
Select the Dimension and move it to a different layer.
Most of the Dimension (text, arrows, extension lines) will be moved to the new layer however the dimension line remains on the current layer.

Operating System and LibreCAD version info
Linux Mint 20.2
Librecad 2.2.0-rc2

In the attached photo, the dimensions shown were moved from the "blue" layer to the "red" layer.  Note how the dimension lines are still blue (color by layer is enabled).

Selection_001

melwyncarlo:
* Solved issue #1396

_BOTH_ the bugs mentioned in this issue have been resolved.

### DIFF
--- a/librecad/src/lib/engine/rs_dimangular.cpp
+++ b/librecad/src/lib/engine/rs_dimangular.cpp
@@ -131,6 +131,7 @@ RS_Entity* RS_DimAngular::clone() const
     d->setOwner( isOwner());
     d->initId();
     d->detach();
+    d->updateDim();
 
     return d;
 }
@@ -341,19 +342,23 @@ void RS_DimAngular::updateDim(bool autoText /*= false*/)
 
     RS_Vector distV;
     double textAngle {0.0};
-    double angle1 {textPos.angleTo( dimCenter) - M_PI_2};
 
-    // rotate text so it's readable from the bottom or right (ISO)
-    // quadrant 1 & 4
-    if (angle1 > M_PI_2 * 3.0 + 0.001
-        || angle1 < M_PI_2 + 0.001) {
-        distV.setPolar( av.gap(), angle1 + M_PI_2);
-        textAngle = angle1;
-    }
-    // quadrant 2 & 3
-    else {
-        distV.setPolar( av.gap(), angle1 - M_PI_2);
-        textAngle = angle1 + M_PI;
+    if (!this->getInsideHorizontalText())
+    {
+        double angle1 {textPos.angleTo( dimCenter) - M_PI_2};
+
+        // rotate text so it's readable from the bottom or right (ISO)
+        // quadrant 1 & 4
+        if (angle1 > M_PI_2 * 3.0 + 0.001
+            || angle1 < M_PI_2 + 0.001) {
+            distV.setPolar( av.gap(), angle1 + M_PI_2);
+            textAngle = angle1;
+        }
+        // quadrant 2 & 3
+        else {
+            distV.setPolar( av.gap(), angle1 - M_PI_2);
+            textAngle = angle1 + M_PI;
+        }
     }
 
     // move text away from dimension line:
--- a/librecad/src/lib/engine/rs_dimdiametric.cpp
+++ b/librecad/src/lib/engine/rs_dimdiametric.cpp
@@ -76,6 +76,7 @@ RS_Entity* RS_DimDiametric::clone() const {
 	d->setOwner(isOwner());
 	d->initId();
 	d->detach();
+   d->updateDim();
 	return d;
 }
 
--- a/librecad/src/lib/engine/rs_dimension.cpp
+++ b/librecad/src/lib/engine/rs_dimension.cpp
@@ -431,9 +431,13 @@ void RS_Dimension::updateCreateHorizontalTextDimensionLine(const RS_Vector& p1,
 
     // finally, add the dimension line(s) and text to the drawing
     if (outsideArrows && dimensionLineOutside1) {
+        dimensionLineOutside1->setPen(pen);
+        dimensionLineOutside2->setPen(pen);
         addEntity(dimensionLineOutside1);
         addEntity(dimensionLineOutside2);
     } else if (splitDimensionLine && dimensionLineInside1) {
+        dimensionLineInside1->setPen(pen);
+        dimensionLineInside2->setPen(pen);
         addEntity(dimensionLineInside1);
         addEntity(dimensionLineInside2);
     } else {
--- a/librecad/src/lib/engine/rs_dimlinear.cpp
+++ b/librecad/src/lib/engine/rs_dimlinear.cpp
@@ -83,6 +83,7 @@ RS_Entity* RS_DimLinear::clone() const {
 	d->setOwner(isOwner());
 	d->initId();
 	d->detach();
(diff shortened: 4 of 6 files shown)


jbergengh:
The issue appears to be fixed for all dimension types except Aligned dimensions.
In this screenshot, the dimensions on the first row were created on the Green layer.
On the second row they were created on Green and moved to Red.
You can see all were moved but the dimension line on the Aligned dimension was not moved.
Almost there!
Thank you melwyn.
Aligned

jbergengh:
Regarding the other bug (Horizontal text not supported in angular dimensions).
The dimension lines should have a gap around the numerical value instead of overwriting the text.
Similar to the horizontal and vertical dimensions in this screenshot.
angular gap

jbergengh:
Question:  Please excuse my ignorance, I am a noob at github.
I noticed the version number of LibreCAD when testing this fix was 2.2.0-alpha.
Shouldn't it be 2.2.0-rc2 (the later version than alpha?)

lordofbikes:
@jbergengh the version number is created from git tags. I assume that `-alpha` is the default when no tags are found.
You have to add `--tags` to your `git pull` command to fetch tags.
When you use sources from a zip download, you have no tags, but this is no problem.
The feature is there, when you have the right sources.

melwyncarlo:
With regards to the first bug, it shouldn't happen if you ensure that the `Use current layer` checkbox is selected.

By default, I have a habit of keeping both the checkboxes selected while performing such an operation.

To solve it, I first have to understand the difference between the two options. I have a feeling that _both_ the checkboxes are _redundant_, because if the entity is moved/copied within the same layer, it should inherit the current layer's attributes, and if it is moved to a new layer, then it should inherit that new layer's attributes.

jbergengh:
@melwyncarlo here is a video that I hope will help diagnose the bug.
[https://youtu.be/KQD08u2uqwI](url)

jbergengh:
@melwyncarlo here is a video to demonstrate the issue

https://youtu.be/KQD08u2uqwI

melwyncarlo:
> The issue appears to be fixed for all dimension types except Aligned dimensions. In this screenshot, the dimensions on the first row were created on the Green layer. On the second row they were created on Green and moved to Red. You can see all were moved but the dimension line on the Aligned dimension was not moved.

The issue was that the dimensions were not being updated post-transferring them onto a new layer.
It's sorted now! :smiley:

melwyncarlo:
> Regarding the other bug (Horizontal text not supported in angular dimensions). The dimension lines should have a gap around the numerical value instead of overwriting the text. Similar to the horizontal and vertical dimensions in this screenshot.

I've sorted this as well, in #1427.

jbergengh:
@melwyncarlo I tested #1409 and it is working perfectly now.  Thank you very much. Excellent job.

melwyncarlo:
@jbergengh You're welcome, and thank you very much! :smiley:

jbergengh:
I found another case where dimension lines on dimensions with horizontal text are not getting updated.
Create some dimensions with horizontal text, then go to Options - Current Drawing Preferences - and change any of the dimension settings, such as Linear precision, Arrow size, text height, or anything else that would cause the dimension to be updated and re-drawn.  Same condition as before - the dimension line does not update.  Maybe this issue more widespread than we thought.

jbergengh:
Also draw dimension on green layer, change active layer to red, delete the dimension and undo the delete.
Dimension line comes back as red.
This issue could be very widespread and maybe not worth the effort to track it all down.
What do you think?

melwyncarlo:
This problem arose because the dimension lines created in the case of horizontal alignment were different from that of the aligned case. The aligned dimension line had an option to update its color, while the dimension lines in the horizontal alignment case did not; now, they do.

I've tested it all of your conditions, and because this fix was internal, it should fix all other conditions, too.

jbergengh:
@melwyncarlo could you please try the following steps to see if you get the same results I do?  Using your latest "Task_4" fixes.

1. Open LibreCAD
2. Options - Application Preferences - Defaults - Modify layer... Checked On
3. Open the attached file "Jam Cleats.dxf"
4.   Observe: Dimension lines are Red.  They should be Magenta (color assigned to the Dimensions layer)
5. Select all the dimensions in one of the views.
6. Click the "Dimensions" layer (the same layer the dimensions are already on)
7.   Observe: The selected dimensions turn Magenta.  That is good and as expected.  They dimensions are "repaired".
8. Create a new layer named "Green" with a default pen color of "Green"
9. Verify the Green layer is the active layer
10. Options - Current Drawing Preferences - Dimensions - Format units - Linear units:  Change "Fractional" to "Decimal"
11. OK
12.   Observe: All dimension lines revert back to Green, including the ones that were repaired in step 7.  Should be Magenta.
13. Select all dimensions on drawing.
14. Click "Dimensions" layer to "repair" them again.
15.   Observe:  All dimensions turn Magenta.  Good.  As expected.
16. Save the drawing as a new file "test.dxf"
17. Exit LibreCAD
18. Restart LibreCAD
19. Open "test.dxf" drawing
20.   Observe:  All dimensions are back to green again, even though they were fixed in steps 15 & 16
[Jam Cleats.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/7612714/Jam.Cleats.dxf.zip)
.

melwyncarlo:
Go to `Options → Drawing Preferences → Dimensions`.

Under `Text size & position`, `Extension lines` and `Dimension lines, arrows & ticks` **(all three sections)**, change the `Color` from `By Layer` to `By Block`. That should fix it across _saves_.

Why does it do that? I've got no idea in the slightest. Using `By Block` sets the color to the currently active layer, and using `By Layer` sets it to the entity's layer. It's not how it's supposed to be, that's as far as I can tell. Also, none of my changes in _Task_4_ involve setting a color to an entity.

jbergengh:
Changing all three sections from Color by Layer to Color by Block solved the issue when using Task_4.

I don't understand why it would work this way because selecting an entity and changing its Attributes to By Block strips that entity of its Color and/or Width and/or Line type information, and makes it ignore what its layer's settings are.  Having to use Color by Block in drawing preferences to get Coloring by Layer is confusing to me.  Maybe I'm not understanding something.

Nevertheless, your fixes work perfectly when I set drawing preferences to By Block.  Excellent work as always,  Melwyn

melwyncarlo:
Thank you very much @jbergengh! :smiley:

No, you have said it correctly, atleast as per the DXF documentation. It is definitely a bug, though I cannot find its source.

dxli:
issues already fixed

Issue #1368, opened 2021-08-02
Bug: Mouse icon is really big in v2.2.0-rc2


MarcelloTheArcane:
### Expected behavior

I'm pretty sure the mouse icon was much smaller before - I updated LibreCAD to get the print tiling feature.

In this PDF instruction manual you can see a much smaller mouse icon:
![image](https://user-images.githubusercontent.com/21159570/127860368-b78a1367-f22d-4a77-b33f-0c38d3ef386a.png)

### Observed behavior

It's huge!  So much so that I think when it disappears for some actions, the rest of the application jumps around.

![image](https://user-images.githubusercontent.com/21159570/127860192-e32287f4-15f7-4b88-ba05-a69b79546e84.png)

### Steps to reproduce or sample file

N/A - just open LibreCAD.  It's an application issue rather than specific file issue.

### Operating System and LibreCAD version info

Version: 2.2.0-rc2
Compiler: GNU GCC 7.3.0
Compiled on: Dec 31 2020
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 10 (10.0)

fa201:
Possibly linked to #1303, #1026, #1010.

MarcelloTheArcane:
Thanks @fa201 - sorry I didn't see those.

Shall I close this issue, as it's being tracked in others?

fa201:
I believe not as some report were supposed to be OS specific (Ubuntu). Anyway it is not link to the chain of issues related to this subject so it should be easy to close all of them when it is solved.

melwyncarlo:
The icon's size is implemented as is. The big mouse icon (located at `librecad/res/icons`) size was **80x80** pixels. The small mouse icon (located at `librecad/res/ui`) size is **24x24** pixels. I think _this_ icon was used before. To implement the right icon, just ensure that the icon size is about **25x25** pixels.

lordofbikes:
fixed with eb0223968761df405c16a4885fff36c2fd05e9d1

Issue #1303, opened 2021-01-10
Status bar redimensioning: mouse image is not redimensioned


fa201:
### Expected behavior
Mouse image should be redimensioned as the status bar height.

### Observed behavior
When redimensioning the status bar through **Options/Widget options/Status bar**, the mouse image is not redimensioned. 
Size changed from 32 to 30. Test changed from 11 to 8.
![image](https://user-images.githubusercontent.com/49532187/104124248-1208fc80-5350-11eb-9edc-a38a2230a80f.png)

### Operating System and LibreCAD version info
Version: 2.2.0-rc2
Compiler: GNU GCC 7.5.0
Compiled on: Jan 10 2021
Qt Version: 5.9.5
Boost Version: 1.65.1
System: Ubuntu 18.04.5 LTS
Screen definition if relevant : 1920x1080.

Possibly linked to #1010 and #1026.

lordofbikes:
This is directly related to #1026 and should be solved with a status bar formatting revamp

lordofbikes:
fixed with eb0223968761df405c16a4885fff36c2fd05e9d1

PR #1410, opened 2021-11-16
status bar height and mouse icon size, fixes #1026, #1303, #1368

### LINKED ISSUE #1026: status bar is resized during showMessage()
### Expected behavior
The status bar height is fixed

### Observed behavior
During showMessage(), e.g. on auto save, the status bar height is reduced.
This makes the layout jump on every auto save cycle.

Open a new file
Draw some entities
Wait for auto save

For faster success reduce the auto save timer to minimum, means 1 minute

Ubuntu 18.10
LibreCAD 2.1.2

see forum thread http://forum.librecad.org/cursor-jumps-during-autosave-tp5716571.html for more details

melwyncarlo:
* Solved issue #1368
* Solved issue #1303 
* Solved issue #1026 

<hr>

**You may ignore this now :** 
> The icon's size is implemented as is. The big mouse icon (located at `librecad/res/icons`) size was **80x80** pixels. The small mouse icon (located at `librecad/res/ui`) size is **24x24** pixels. I think _this_ icon was used before. To implement the right icon, just ensure that the icon size is about **25x25** pixels.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -204,7 +204,7 @@ QC_ApplicationWindow::QC_ApplicationWindow()
     settings.endGroup();
 
     mouseWidget->findChild<QLabel*>("lMousePixmap")->setPixmap(
-    mouseWidget->findChild<QLabel*>("lMousePixmap")->pixmap()->scaled(statusBarHeight, statusBarHeight));
+    QPixmap(":/icons/mouse.svg").scaled(statusBarHeight, statusBarHeight));
 
     RS_DEBUG->print("QC_ApplicationWindow::QC_ApplicationWindow: creating LC_CentralWidget");
 
@@ -3239,7 +3239,7 @@ void QC_ApplicationWindow::widgetOptionsDialog()
             statusBar()->setFixedHeight(statusbar_height);
 
             mouseWidget->findChild<QLabel*>("lMousePixmap")->setPixmap(
-            mouseWidget->findChild<QLabel*>("lMousePixmap")->pixmap()->scaled(statusbar_height, statusbar_height));
+            QPixmap(":/icons/mouse.svg").scaled(statusbar_height, statusbar_height));
         }
     }
     settings.endGroup();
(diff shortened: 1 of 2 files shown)


melwyncarlo:
I figured out the actual problem: how to scale the image via the program code, and not the physical image file.

The QPixmap class has a function called `scaled` that does the image rescaling.
I have not preserved the aspect ratio. If you wish to preserve the aspect ratio, 
then I could append a third argument to the function: `Qt::KeepAspectRatio`

melwyncarlo:
**Is this better?**

<hr>

image

image

melwyncarlo:
My v1 commit does not cause the status bar to change height unexpectedly. But the mouse icon is too small.

My v2 commit allows for the SVG mouse icon to be scaled and displayed, but that causes the status bar to change height unexpectedly.

I tried converting the `setMinimumHeight` function of the status bar to `setFixedHeight` instead; yet no relief. I think that the problem is an indirect effect of something else.

melwyncarlo:
My v3 commit does it best. The `qg_mousewidget.cpp` has been reset back to its **master** branch version.

All changes have been concentrated in the file `qc_applicationwindow.cpp`.

librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -3226,9 +3234,12 @@ void QC_ApplicationWindow::widgetOptionsDialog()
         settings.setValue("AllowStatusbarHeight", allow_statusbar_height);
         if (allow_statusbar_height)
         {
-            int statusbar_height = dlg.statusbar_height_spinbox->value();
+            statusbar_height = dlg.statusbar_height_spinbox->value();
             settings.setValue("StatusbarHeight", statusbar_height);
-            statusBar()->setMinimumHeight(statusbar_height);
+            statusBar()->setFixedHeight(statusbar_height);
+
+            mouseWidget->findChild<QLabel*>("lMousePixmap")->setPixmap(
```
melwyncarlo, line 3241: This is *BAD* (commit v3). It basically rescales the pixmap image, but relatively. So after scaling down, and then rescaling up, it rasterizes the image. 

Fixed in commit v4.

lordofbikes:
fixed with eb0223968761df405c16a4885fff36c2fd05e9d1

PR #1411, opened 2021-11-16
keep application preferences enabled in menu, fixes #1369

### LINKED ISSUE #1369: Allow "Application Preferences" when no document is open
### Expected behavior

I would like to be able to view and edit the application preferences, even if no document is open.

### Observed behavior

Currently the `Application Preferences` option is greyed out if no document is open.

image

1. Open LibreCAD
2. Go to `Options`
3. Observe that `Application Preferences` is disabled

Version: 2.2.0-rc2
Compiler: GNU GCC 7.3.0
Compiled on: Dec 31 2020
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 10 (10.0)

melwyncarlo:
* Solved issue #1369

If needed, some other options, too, could be kept enabled.


lordofbikes:
many thanks :+1:

I merged manually, because the only menu point to disable is *Current Drawing Preferences*, all the others can stay enabled, even when no drawing is open.
Also I added some type safety for `connect()`

PR #1414, opened 2021-11-17
update recent files list when file is saved as ..., fixes #1364

### LINKED ISSUE #1364: Recent Files list not updated at 'Save As'
### Expected behavior
Create a new file.
Save it as 'my_file.dxf'.
'my_file.dxf' should be in the list of **Recent Files**

### Observed behavior
'my_file.dxf' is not in the list of **Recent Files**.

Version: 2.2.0-rc2-15-g76828565
Compiler: GNU GCC 7.5.0
Compiled on: Apr 17 2021
Qt Version: 5.12.8
Boost Version: 1.65.1
System: Ubuntu 20.04.2 LTS

It looks like the filename it added to **Recent Files** only after **Open** whereas it should be added also after **Save As**.

melwyncarlo:
* Solved issue #1364


lordofbikes:
many thanks :+1:

PR #1415, opened 2021-11-17
Resolved the 'open existing/deleted file' issues

melwyncarlo:
In both stable and development versions, 

if you open up an already open (existing) file, it is supposed to show a warning; which it does.

But then it also opens up another copy of the same file.

I've solved that issue. Now, it no longer opens up multiple copies of the same file; 
it shows the warning, as usual, and that's it; it safely returns back from the `slotFileOpen` function.

<hr>

Also, upon attempting to open a deleted file from the recently open menu, it shows the aborted error message;
but it does not remove the file listing from the recent menu. I've added a bit of code that removes that particular 
file name from the recent menu.

There is an `add` function in the `qg_recentfiles.cpp` file. 
I have included a simple `remove` function to it.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -1889,9 +1889,13 @@ void QC_ApplicationWindow::slotFileOpen(const QString& fileName, RS2::FormatType
         commandWidget->appendHistory(message);
         statusBar()->showMessage(message, 2000);
 
-	} else {
-		QG_DIALOGFACTORY->commandMessage(tr("File '%1' does not exist. Opening aborted").arg(fileName));
+    }
+    else
+    {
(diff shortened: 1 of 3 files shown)


lordofbikes:
For the first point I see no need, I can  even imagine, that it's intended.
E.g. to open a copy of a modified drawing to compare it with it's former state.
Even when modifying the copy was possible, saving and overwriting the first open drawing failed. But saving the copy with a new name worked.
So changing this may protect some users, but also offend others, thus we keep it untouched.

The second point addresses a rare edge case.
On starting LibreCAD, when the recent files list is created, deleted files are not added.
So this issue only arise, when a file from recent file list is deleted, while LibreCAD is open.
Because there is no exception risen, I see this as safe. The next restart of the application solves the problem.

If it wasn't mixed up with the first issue, I would have merged the second patch.

PR #1416, opened 2021-11-18
Solved the 'freezing status bar on focus out' issue

melwyncarlo:
Prior to this code, the status bar information (especially the coordinates) **just freezes** when the mouse focuses outside of the graphics view, and also when a file is closed.

I have implemented a method to hide the widgets within the status bar when the mouse focuses out, 
and show the widgets when the mouse focuses inside one of the graphic views (i.e. one of the open files).

This method will not _(should not)_ come in the way of status bar messages.
The status bar messages will continue to show regardless.

<hr>

Also, regarding the status bar message timeout:
there were magic numbers of _2000_ scattered throughout the _qc_applicationwindow.cpp_ file.
2000 stands for 2000 milliseconds (or just 2 seconds), which is the time for which 
the status bar message will be displayed.

I've assigned them a common constant variable at the top of the file.
This would make future edits simpler.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -2783,7 +2783,7 @@ void QC_ApplicationWindow::slotViewStatusBar(bool toggle) {
 
     - by Melwyn Francis Carlo
 */
-void QC_ApplicationWindow::status_bar_user_widgets_visibility(bool clear)
+void QC_ApplicationWindow::status_bar_widgets_visibility(bool clear)
 {
     QString statusBarMessage = statusBar()->currentMessage();
 
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -129,6 +129,9 @@ QC_ApplicationWindow* QC_ApplicationWindow::appWindow = nullptr;
  *	*/
 #define WTB_MAX_SIZE        79
 
+#define DEFAULT_STATUS_BAR_MESSAGE_TIMEOUT 2000
+
+
 /**
  * Constructor. Initializes the app.
  */
@@ -178,6 +181,7 @@ QC_ApplicationWindow::QC_ApplicationWindow()
     status_bar->addWidget(m_pActiveLayerName);
     grid_status = new TwoStackedLabels(status_bar);
     grid_status->setTopLabel(tr("Grid Status"));
+    grid_status->setObjectName("grid status");
     status_bar->addWidget(grid_status);
 
     settings.beginGroup("Widgets");
@@ -373,7 +377,7 @@ QC_ApplicationWindow::QC_ApplicationWindow()
     //plugin load
     loadPlugins();
 
-    statusBar()->showMessage(qApp->applicationName() + " Ready", 2000);
+    statusBar()->showMessage(qApp->applicationName() + " Ready", DEFAULT_STATUS_BAR_MESSAGE_TIMEOUT);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -129,6 +129,9 @@ QC_ApplicationWindow* QC_ApplicationWindow::appWindow = nullptr;
  *	*/
 #define WTB_MAX_SIZE        79
 
+#define DEFAULT_STATUS_BAR_MESSAGE_TIMEOUT 2000
```
lordofbikes, line 132: Using preprocessor macros for constants is deprecated.
In this case I think a `private` `enum` inside the class wold be best choice.
`enum`s should use camel case and start upper case.

librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -1484,6 +1488,9 @@ QC_MDIWindow* QC_ApplicationWindow::slotFileNew(RS_Document* doc) {
     connect(view, SIGNAL(gridStatusChanged(const QString&)),
             this, SLOT(updateGridStatus(const QString&)));
 
+    connect(view, SIGNAL(status_bar_widgets_visibility_request(bool)),
+            this, SLOT(status_bar_widgets_visibility(bool)));
```
lordofbikes, line 1492: The deprecated macro syntax should not be used any longer.
The preferred syntax is:
```
    connect( view, &QG_GraphicView::status_bar_widgets_visibility_request,
                   this, &QC_ApplicationWindow::status_bar_widgets_visibility);
```
This is type save and detects issues at compile time, not at run time, like the macro syntax.

This applies to new code, existing code with macro syntax should not be touched, until there are strong reasons to update.

lordofbikes:
The status bar is not freezing, there is just no coordinate update when the mouse cursor leaves the drawing area.
This is expected behavior and I can't see any issue here.
But with this change I see the same issue as #1026, but much more worse. Hiding the status bar when the mouse leaves the drawing area changes layout and creates flickering on some locations. And the jumping toolbars at the bottom makes them uncomfortable to control.

Anyhow I commented two common issues which concerns your other pull requests too and should be attended for further contributions.

So for the mentioned reason I must discard this.

melwyncarlo:
> The status bar is not freezing, there is just no coordinate update when the mouse cursor leaves the drawing area. This is expected behavior and I can't see any issue here. But with this change I see the same issue as #1026, but much more worse. Hiding the status bar when the mouse leaves the drawing area changes layout and creates flickering on some locations. And the jumping toolbars at the bottom makes them uncomfortable to control.
> 
> Anyhow I commented two common issues which concerns your other pull requests too and should be attended for further contributions.
> 
> So for the mentioned reason I must discard this.


Could you please reconsider my pull-request? _Freezing_ was the incorrect term, I get it now.
I'm not hiding the status bar, just temporarily blanking (clearing) it out, which is how it should be;
otherwise, I believe it is misleading to the user, and also untidy.

For hiding the status bar, there already is another option in the menu bar.

Regardless, the flickering should no longer happen as per my pull-request #1410 
The flickering now no longer happens even when saving or opening, etc.

Combine both these pull-requests, and there are no problems at all.

If you would consider re-opening this pull-request, I'll resolve the above mentioned code changes.

librecad/src/main/qc_applicationwindow.cpp
melwyncarlo: I did not know that. Is there a contributing developer's guidelines page that I could read through?

I wrote my code by referencing the style employed by other authors in the respective files.

librecad/src/main/qc_applicationwindow.cpp
melwyncarlo: I did not know that. I basically copy-pasted a similar code a couple lines above, and changed the variables.

PR #1417, opened 2021-11-18
specify the unit of auto save time, fixes #1412

### LINKED ISSUE #1412: Specify auto save time units
**Location :**
`Menu Bar → Options → Application Preferences → Defaults → Program Defaults → Auto save time`

**LibreCAD version :**
Latest development version 2.2.0

<hr>

It would be better to specify the units of time. 
* One way would be to append ` (seconds)` to the label.
* Another way would be to add a combo-box with the options: seconds and minutes.

The first idea is simple and uniform.

melwyncarlo:
* Solved issue #1412


lordofbikes:
many thanks :+1:

manual merge because of unrelated changes

PR #1422, opened 2021-11-19
Removed the 'All files' and 'Any files' options from the 'Import Image' and 'Export PDF' dialogs respectively

melwyncarlo:
In the **Import Image** and **Export PDF** dialogs, the '_All files_' type amongst other options, makes no sense.

The **Export SVG** and **Export Image** dialogs do not have them :

Screenshot from 2021-11-19 13-53-02

Screenshot from 2021-11-19 14-08-56

<hr>

Then, why do the **Import Image** and **Export PDF** dialogs have them?

Screenshot from 2021-11-19 14-09-58

Screenshot from 2021-11-19 14-10-16

<hr>

How would the 'Any files' option benefit the export of a specific type of file (e.g. PDF)?

Wouldn't the import of 'All files' whilst importing specific image file types lead to bugs and/or unpredictable behaviour?

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -2325,8 +2325,7 @@ void QC_ApplicationWindow::slotFilePrint(bool printPDF) {
         QString     defFilter("PDF files (*.pdf)");
         QStringList filters;
 
-        filters << defFilter
-                << "Any files (*)";
+        filters << defFilter;
 
         fileDlg.setNameFilters(filters);
         fileDlg.setFileMode(QFileDialog::AnyFile);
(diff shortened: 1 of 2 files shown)


lordofbikes:
While I don't bother about Export PDF, we should keep All/Any for Import.
In case someone has individual naming conventions, they possibly fail to find files without All/Any option.
For the same reason I'd keep Export PDF, in case someone wants to override an existing file.

This is very hypothetical, but it doesn't mind people who follow usual naming conventions. But it may bother someone who was happy about it in past.
We can consider to add Any/All to all file dialogs, but there was no request yet.

dxli:
Agreed with @lordofbikes , the extra choices could be useful in some situations. 

For "Export as PDF", the pdf extension filter is already the default, so the extra options do no cause confusion by default.

PR #1423, opened 2021-11-19
preserve filename on export - save as ..., fixes #1301

### LINKED ISSUE #1301: Export - save as window should preserve filename (just changing extension)
Hello,

In LibreCAD, export function (as pdf or as svg) opens a "save as" window.
Unfortunately, no filename is given by default.
Maybe a good idea could be to preserve filename (just changing file extension) and so,
save as window should open with such a filename as default.

Kind regards

melwyncarlo:
* Solved issue #1301

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -1493,7 +1493,7 @@ QC_MDIWindow* QC_ApplicationWindow::slotFileNew(RS_Document* doc) {
     if (w->getDocument()->rtti()==RS2::EntityBlock) {
         w->setWindowTitle(tr("Block '%1'").arg(((RS_Block*)(w->getDocument()))->getName()) + "[*]");
     } else {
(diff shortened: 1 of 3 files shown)


librecad/src/main/qc_applicationwindow.cpp
```diff
@@ -1493,7 +1493,7 @@ QC_MDIWindow* QC_ApplicationWindow::slotFileNew(RS_Document* doc) {
     if (w->getDocument()->rtti()==RS2::EntityBlock) {
         w->setWindowTitle(tr("Block '%1'").arg(((RS_Block*)(w->getDocument()))->getName()) + "[*]");
     } else {
-        w->setWindowTitle(tr("unnamed document %1").arg(id) + "[*]");
+        w->setWindowTitle(tr("Untitled Document %1").arg(id) + "[*]");
```
melwyncarlo, line 1496: If you're going to suggest a default filename as **Untitled** during save, you might as well use the same name as the default title.

Issue #1421, opened 2021-11-18
Feature request: add button to snap middle manual


gerardbm:
Enhancement to add a add a new method to snap to a point that is in the middle between two points.

This snap is implemented in QCAD. See the relative [article](https://qcad.org/doc/qcad/latest/reference/en/scripts/Snap/SnapMiddleManual/doc/SnapMiddleManual_en.html).

[Video demo](https://www.youtube.com/watch?v=4pGMmmIxgbI).

melwyncarlo:
What about the existing **Snap middle** button that does the same thing?

gerardbm:
_Snap middle_ is also useful to snap to the middle of an entity. However, this functionality (also well implemented in QCAD and Allplan) allows you find manually a middle point between two points without having an entity there. It's useful for the origin point and the end point.

Actually, old QCAD versions didn't have this function, so I was surprised when I saw this in recent versions. Since QCAD and LibreCAD are "brothers", I think it shouldn't be difficult to implement.

melwyncarlo:
Not that challenging. I can get on with it once its approved as a _feature request_.

gerardbm:
Yeah, that would be super nice! 

BTW, I should learn C++ to help making contributions. At the moment I have I compile the source code, so from time to time I do a 'git pull' to check the improvements.

melwyncarlo:
Done and tested! I'll upload my code soon.

melwyncarlo:
> BTW, I should learn C++ to help making contributions. At the moment I have I compile the source code, so from time to time I do a 'git pull' to check the improvements.

I would suggest learning from the maker itself :  The C++ Programming Language by Bjarne Stroustrup
I gave away that book, and so now I refer to :  C++ Primer Plus by Stephen Prata

There are many online resources too: [learncpp.com](https://www.learncpp.com/), [cplusplus.com](https://www.cplusplus.com/), [TutorialsPoint](https://www.tutorialspoint.com/cplusplus/index.htm), [W3Schools](https://www.w3schools.com/CPP/default.asp), etc.
There are many free e-books too.

melwyncarlo:
The best part about building from source is that you get to make mistakes and try again and again till you succeed, and learn a lot on the way.

melwyncarlo:
> Done and tested! I'll upload my code soon.
> 
> https://youtu.be/XZU929ykxvg


In the video, the dotted lines seemed too faint, so I changed them to dots-and-dashes, as seen below:

[...]

PR #1424, opened 2021-11-20
Advanced Snap Middle Manual Tool

melwyncarlo:
* Solved issue #1421

<hr>

This tool can snap (set) the relative-zero marker in the middle between two points chosen by the user.

The middle point, by default, rests at a poistion 50% relative to the position of the first point chosen, that is, at the center of the imaginary line connecting the two user-defined points.

This percentage, however, can be modified by the user in order to place the marker at a different position along the imaginary line.

<hr>

Video link :  [https://youtu.be/XZU929ykxvg](https://youtu.be/XZU929ykxvg)

<hr>

In the video, the dotted lines seemed too faint, so I changed them to dots-and-dashes, as seen below:

Screenshot from 2021-11-20 22-53-07

### DIFF
--- a/librecad/res/icons/icons.qrc
+++ b/librecad/res/icons/icons.qrc
@@ -50,6 +50,7 @@
         <file>snap_grid.svg</file>
         <file>snap_intersection.svg</file>
         <file>snap_middle.svg</file>
+        <file>snap_middle_manual.svg</file>
         <file>lock_rel_zero.svg</file>
         <file>set_rel_zero.svg</file>
         <file>arc_3_points.svg</file>
--- a/librecad/res/icons/snap_middle_manual.svg
+++ b/librecad/res/icons/snap_middle_manual.svg
@@ -1,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.2" width="72.3mm" height="72.3mm" viewBox="0 0 7230 7230" preserveAspectRatio="xMidYMid" fill-rule="evenodd" stroke-width="28.222" stroke-linejoin="round" xmlns="http://www.w3.org/2000/svg" xmlns:ooo="http://xml.openoffice.org/svg/export" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:presentation="http://sun.com/xmlns/staroffice/presentation" xmlns:smil="http://www.w3.org/2001/SMIL20/" xmlns:anim="urn:oasis:names:tc:opendocument:xmlns:animation:1.0" xml:space="preserve">
+ <defs class="ClipPathGroup">
+  <clipPath id="presentation_clip_path" clipPathUnits="userSpaceOnUse">
+   <rect x="0" y="0" width="7230" height="7230"/>
+  </clipPath>
+  <clipPath id="presentation_clip_path_shrink" clipPathUnits="userSpaceOnUse">
+   <rect x="7" y="7" width="7216" height="7216"/>
+  </clipPath>
+ </defs>
+ <defs class="TextShapeIndex">
+  <g ooo:slide="id1" ooo:id-list="id3 id4 id5 id6 id7 id8 id9"/>
+ </defs>
+ <defs class="EmbeddedBulletChars">
+  <g id="bullet-char-template-57356" transform="scale(0.00048828125,-0.00048828125)">
+   <path d="M 580,1141 L 1163,571 580,0 -4,571 580,1141 Z"/>
+  </g>
+  <g id="bullet-char-template-57354" transform="scale(0.00048828125,-0.00048828125)">
+   <path d="M 8,1128 L 1137,1128 1137,0 8,0 8,1128 Z"/>
+  </g>
+  <g id="bullet-char-template-10146" transform="scale(0.00048828125,-0.00048828125)">
+   <path d="M 174,0 L 602,739 174,1481 1456,739 174,0 Z M 1358,739 L 309,1346 659,739 1358,739 Z"/>
+  </g>
+  <g id="bullet-char-template-10132" transform="scale(0.00048828125,-0.00048828125)">
+   <path d="M 2015,739 L 1276,0 717,0 1260,543 174,543 174,936 1260,936 717,1481 1274,1481 2015,739 Z"/>
+  </g>
+  <g id="bullet-char-template-10007" transform="scale(0.00048828125,-0.00048828125)">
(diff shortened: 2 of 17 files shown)


dellus:
A nice to have feature. A bit clumsy to use, though. I would prefer it to be integrated into the drawing workflow. Like draw a line from an existing point A to the middle of point B and C. I think this is called a transparent command, a sub command used within a main command. Probably not that easy. Do you think you could tackle this? Or are inclined to at all?
If it's too much effort we can discuss whether it is better to have it this way (defining relative zero) than not to have it at all.

A similar and even more valuable feature would be Snap to the imaginary intersection of two not really intersecting lines.
More can be considered like Snap to perpendicular basepoint from a point to a line, but not so important. 

Where to place it in the user interface? You have put it at Tools. Maybe the snaps or relative zero bar is a better place.

melwyncarlo:
I did not have all this in mind. My aim was to simply accomplish a similar feature from the newer version of QCAD as requested by @gerardbm. It seems to do the exact same thing as mentioned in his reference video, but slightly improvised. 

The icon seems a little faint. I'm wondering if I should thicken the lines.
I basically mimicked QCAD's tool icon, but styled it as per the other LibreCAD icons.
(I referred to the other LibreCAD icons for color and line thickness.)

I put it in the tool bar, because it is a _tool_ and _not just a setting_.
Though I might be wrong.

melwyncarlo:
> A similar and even more valuable feature would be Snap to the imaginary intersection of two not really intersecting lines.
> More can be considered like Snap to perpendicular basepoint from a point to a line, but not so important.

Both these seem interesting.

lordofbikes:
@dellus nailed, what I already meant on Zulip, that this may be more complex than it looks on first sight.

We have no similar function in LibreCAD currently, both mentioned features don't work like common snap functions.
The snap functions work in background and are implemented in each entities code. They already do lots of work on each mouse move trigger. Iterating the document, finding nearest entities and then acquire nearest snap point from activated snap functions.

For these new features we can't implement them the same way. In a complex drawing, finding imaginary intersection or middle points on each mouse move event, will end in drastic rise of computing time and therefor slow down GUI reaction time. Also there would be a bunch of snap points for which it is not obvious to which entities they belong.

So these features need a new implementation. They work more similar to trim function or tangent drawing tools. The user must activate this function, then select two entities and the resulting intersection or middle point must be sourced to the active function. This can be drawing or modifying functions, so these features have to be implemented for all kind of tools and functions.
And copy/paste a possible implementation over and over for each entity and tool is not an acceptable solution.

Some tools need one point, some two and others even more. But possibly these imaginary points are not needed for all points of an action. So a kind of single shot activation would be appropriate I think.
1. activate any tool (e.g. draw or modify)
2. activate imaginary snap
3. select the number of entities to find the imaginary snap point
4. source the point to the tool
5. deactivate imaginary snap
6. select any other point for active tool or return back to 2. to use imaginary snap again

So this is a nice case study and a good training project, but far from implementation in main code base.

With all that said, I expect, that this feature will never be realized for LibreCAD 2.
It's not only about the efforts you may put into this. The expected global changes for this new kind of functionality will also need extensive testing and reviewing, for what we don't have the resources in foreseeable future.
I'll comment this in #1421 too, for better understanding.

melwyncarlo:
> A nice to have feature. A bit clumsy to use, though. I would prefer it to be integrated into the drawing workflow. Like draw a line from an existing point A to the middle of point B and C. I think this is called a transparent command, a sub command used within a main command. Probably not that easy. Do you think you could tackle this? Or are inclined to at all? If it's too much effort we can discuss whether it is better to have it this way (defining relative zero) than not to have it at all.

I can surely tackle it if I correctly understand it's implementation. A graphical representation would surely help.
Point A to the middle of point B and C? I'm assuming here that you are talking about the Snap to perpendicular basepoint tool?

melwyncarlo:
@lordofbikes  Instead of sourcing the point, can't the other tools just read the zero-marker position instead? 

Maybe set the zero-marker position as a private variable, and create a public ` void getRelativeZeroMarkerPosition()`?

Or create a slot function to pass that information through a variable by reference?

These tools basically (in the end) just re-position the marker, that's all! 
Nothing that special about them.
It's mostly about ensuring that the marker is re-positioned correctly.

<hr>

It is indeed a very nice case study. While I do not hope for its implementation, I will still try to create those two tools in my free time. I'm wondering if I should create pull-requests for those two tools, or if there is another means of publishing it?

dellus:
"Point A to the middle of point B and C? I'm assuming here that you are talking about the Snap to perpendicular basepoint tool?"
I mean the "Snap middle manual" tool. 
I don't manage to upload an image here. So imagine there is a square, with the lower left corner as point A, lower right corner as point B and upper left corner is point C. Now I want to draw a line from point A to the center of the square. I start with the ordinary line command. When asked to specify the first point I do so with selecting point A, probably by using ordinary "Snap on Endpoints". When asked to specify the next point, I trigger the "Snap middle manual" tool, for instance by pressing the according button you have designed. It will ask me to specify two points, which I do with ordinary snap again. After having specified the second point my line command is instantly completed ending at the center of the square and also relative zero resting there.
This is just for explanation. Lordofbikes has closed the pull request. He would rather like to see you spending your efforts on LibreCAD 3. But in your "free time" you may have your private ventures...

melwyncarlo:
Now I've understood it, and to me, that seems doable. But the question is, how do I implement the connectivity (sending data from one tool to another)?

Earlier I've mentioned two options. But @lordofbikes is not fond of adding snippets to all the action tool files.

Without touching the other tool files, the only other option that I can think of is deliberately simulating a mouse-click, using the `<QTest>` header file, like in this solution:

[https://stackoverflow.com/questions/21680565/qt5-c-automated-mouse-clicking](https://stackoverflow.com/questions/21680565/qt5-c-automated-mouse-clicking)

dellus:
Oh sweet dreams...
We have to find a subversive workaround to use these tools in LibreCAD 2. Mmmm..., what files will have to be replaced or added in an existing LibreCAD installation? An .exe that does this all?
And the question of workflow integration...
Never mind, just fantasies.

gerardbm:
> Oh sweet dreams... We have to find a subversive workaround to use these tools in LibreCAD 2. Mmmm..., what files will have to be replaced or added in an existing LibreCAD installation? An .exe that does this all? And the question of workflow integration... Never mind, just fantasies.

I think that this could be a specific branch, or even a fork of LibreCAD for this purpose.

melwyncarlo:
@gerardbm My changes are merely less than 1% of the total code. :smile:

melwyncarlo:
@dellus Not fantasies, it is possible; and no subversion, just tactical fixtures.
I've been busy lately. Regardless, there is a workaround. 
A tool _can_ be kept running whilst using another tool.
I've seen another tool utilise this technique, just can't remember which, for now.

gerardbm:
Could this be added (merged) in the version 2?

dxli:
Merged. Closing now.

Issue #1428, opened 2021-11-23
Importing a 2D DXF file generates Z-axis scaling of 0 for DXF output


egilkv:
### Expected behavior
Correct DXF file generated with Z scale of non-zero.

### Observed behavior
Having done "Save as" to a DXF file, a 3D tool like FreeCAD chokes on the DXF input since the Z scale in the input (param 43) is zero.
Error is in line 3250 of test-cu.dxf
It only seems to be a problem when the DXF file is read through an import. 
Note that the imported file (that came from KiCAD) has an X scale (parameter 41) of 1, and Y scale (parameter 42) of 1, but no Z scale (parameter 43), which by DXF spec should default to one.

### Steps to reproduce or sample file
File -> New
File -> Import -> Block -> test-F_Cu.dxf
Add one item at reference point 0,0
File -> Save as -> test-cu.dxf

### Operating System and LibreCAD version info
OS: Ubuntu 18.04.6 LTS
Version: 2.2.0-alpha (but also earlier releases)

egilkv:
Sample files:

[test-F_Cu.dxf.gz](https://github.com/LibreCAD/LibreCAD/files/7588817/test-F_Cu.dxf.gz)

[test-cu.dxf.gz](https://github.com/LibreCAD/LibreCAD/files/7588820/test-cu.dxf.gz)

lordofbikes:
This sounds somehow familiar, but I couldn't find another issue or probably a fix with a quick search, neither here, nor in https://github.com/LibreCAD/libdxfrw.
The 2.2.0-alpha label is used, when sources are not pulled by git or pulled without tags.
Can you specify your version better? Downloaded or built from source? Optimum would be a git hash.

egilkv:
I have pulled the latest source, and reproduced the issue there. Details are:

Version: 2.2.0-rc2-24-g1d314270
Compiler: GNU GCC 7.5.0
Compiled on: Nov 23 2021
Qt Version: 5.9.5
Boost Version: 1.65.1
System: Ubuntu 18.04.6 LTS

I do believe the bug can be fixed by changing line 456 of librecad/src/lib/modification/rs_modification.cpp from:

    RS_Vector vfactor = RS_Vector(factor, factor);

to:

    RS_Vector vfactor = RS_Vector(factor, factor, factor);

However, the treatment of the zscale is very lax generally, and since this is output in the resulting DXF, there are also other instances that could end up with a zscale of zero. I will try to provide a pull request.

[...]

PR #1429, opened 2021-11-23
Try to ensure that entities get a reasonable (i.e. non-zero) Z scale.

egilkv:
While LibreCAD ignores Z scale on input, generating DXF with entities that has a Z scale of zero creates problems for others.
FreeCAD will for instance choke, see https://tracker.freecadweb.org/view.php?id=4791 
An alternative way of dealing with this would be omitting writing Z scales (record 43) to the DXF in the first place.

This is a proposed fix for #1428

### DIFF
--- a/librecad/src/main/doc_plugin_interface.cpp
+++ b/librecad/src/main/doc_plugin_interface.cpp
@@ -316,6 +316,7 @@ void Plugin_Entity::getData(QHash<int, QVariant> *data){
         data->insert(DPI::STARTANGLE, d.angle );
         data->insert(DPI::XSCALE, d.scaleFactor.x );
         data->insert(DPI::YSCALE, d.scaleFactor.y );
+        data->insert(1.0, d.scaleFactor.z );
         break;}
     case RS2::EntityMText: {
         data->insert(DPI::ETYPE, DPI::MTEXT);
@@ -988,7 +989,7 @@ void Doc_plugin_interface::addImage(int handle, QPointF *start, QPointF *uvr, QP
 void Doc_plugin_interface::addInsert(QString name, QPointF ins, QPointF scale, qreal rot){
     if (doc) {
         RS_Vector ip(ins.x(), ins.y());
-        RS_Vector sp(scale.x(), scale.y());
+	RS_Vector sp(scale.x(), scale.y(), 1.0);
 
         RS_InsertData id(name, ip, sp, rot, 1, 1, RS_Vector(0.0, 0.0));
         RS_Insert* entity = new RS_Insert(doc, id);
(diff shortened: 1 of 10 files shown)


librecad/src/main/doc_plugin_interface.cpp
```diff
@@ -316,6 +316,7 @@ void Plugin_Entity::getData(QHash<int, QVariant> *data){
         data->insert(DPI::STARTANGLE, d.angle );
         data->insert(DPI::XSCALE, d.scaleFactor.x );
         data->insert(DPI::YSCALE, d.scaleFactor.y );
+        data->insert(1.0, d.scaleFactor.z );
```
lordofbikes, line 319: This should at least be `data->insert(DPI::ZSCALE, d.scaleFactor.z );`

lordofbikes:
The comment in `doc_plugin_interface.cpp` is the deal breaker I found on a first sight.
It needs more time to review the other changes.
Also I want to check if it's possible to handle this inside `libdxfrw` and avoid changes in LibreCAD.
Anyhow many thanks for your efforts. :+1:

dxli:
Thank you for tracking this down, @egilkv — the diagnosis was correct and the bug is still reproducible on `master`. Closing in favour of #2756, which fixes it a different way; the credit for finding it is yours, and #1428 stays open until that lands.

**Why a new PR rather than a rebase.** Four years on, every file this touches has moved (`librecad/src/lib/engine/` → `.../document/entities/`, `rs_actionblocksinsert.cpp` → `lc_action_block_insert.cpp`, the `rs_modification.cpp` paste hunks → `lc_copyutils.cpp`), so it no longer applies. One hunk has also landed independently: `addInsert()` now reads `data.zscale`, so the import path in the issue title is already fixed.

**Why a different approach.** This PR adds the missing third component at each of ten call sites. That is leaky by construction — `lc_dimarrowregistry.cpp:188` was written after this PR and reintroduced exactly the same `RS_Vector(x, y)` pattern, and the next new call site would too. #2756 instead enforces the invariant in `RS_InsertData`'s constructor, so a zero z becomes 1 for every present and future caller, and guards `writeInsert()` as well, which additionally covers DWG export and inserts read from a file that already carried a zero. Nine lines in three files rather than ten files.

**One defect worth flagging**, in case the same line is ever reused:

```cpp
data->insert(1.0, d.scaleFactor.z );
```

The first argument is the `QHash<int, QVariant>` key. `1.0` truncates to key **1**, which is `DPI::TEXTCONTENT` — documented as *"QString: text string"* — so a plugin asking for text content would have received a double. The intended key, `DPI::ZSCALE` (43), already exists. You were right that it was missing from `getData()`, though: it exported `XSCALE` and `YSCALE` and never `ZSCALE`. #2756 adds it with the correct key.

Your last question on #1428 — whether a zero z scale is valid DXF at all — is still worth an authoritative answer. #2756 takes the conservative line of simply never emitting one, which resolves the interoperability failure either way.

PR #1430, opened 2021-11-24
set default line width, fixes #1326

### LINKED ISSUE #1326: Change default linewidth to 0.25 mm
The default linewidth for layer 0 and all newly created layers is 0.00 mm.
New inexperienced users often are drawing with this setting without being conscious of it. On screen everything looks all right but when printing on paper or exporting to pdf often entities with linewidth 0.00 mm don't appear at all or are very faint, leaving despaired users. The behaviour varies with different LibreCAD versions, OS's, pdf-viewers, and printers or their drivers .
I suggest to change the default linewidth for layer 0 and all newly created layers to 0.25 mm. This is a widely accepted standard for a normal line. Experienced users who might need 0.00 mm (for CAM?) will know what to do.  
By the way, there is also an option to set "default" linewidth. It turns out to be the same as 0.00 mm. But there is no possibility in LibreCAD to define this default setting. In other CAD packages you can set this standard, and it only makes sense then.

melwyncarlo:
* Solved issue #1326

### DIFF
--- a/librecad/src/lib/engine/rs_entity.cpp
+++ b/librecad/src/lib/engine/rs_entity.cpp
@@ -851,6 +851,11 @@ RS_Pen RS_Entity::getPen(bool resolve) const {
             //}
         }
 
+        if (p.getWidth() == RS2::WidthDefault)
+        {
+            p.setWidth(RS2::Width07);
+        }
+
         return p;
     }
 }
--- a/librecad/src/lib/engine/rs_layer.cpp
+++ b/librecad/src/lib/engine/rs_layer.cpp
@@ -44,7 +44,7 @@ RS_LayerData::RS_LayerData(const QString& name,
  * Constructor.
  */
 RS_Layer::RS_Layer(const QString& name):
-	data(name, RS_Pen(Qt::black, RS2::Width00,RS2::SolidLine), false, false)
+	data(name, RS_Pen(Qt::black, RS2::WidthDefault, RS2::SolidLine), false, false)
 {
 }
 


dellus:
Thanks, melwyncarlo. This will spare us a lot of grief.

melwyncarlo:
You're welcome. :smiley:

dxli:
Is there a problem when dring tiny objects?

The default width could be a default width at the view layer by pixels, so it's suitable for drawing geometries.

After the geometry drawing, using can set line width in drawing units (say, mm/inch) by layer/entities setting.

lordofbikes:
The intention here is to fix printing issues. Empty pages, when printing or exporting to PDF, is regularly issued in the forum.

Default line width of `0` is mapped to 1 pixel width on screen, what works well.
But my assumption is, when printing or exporting to PDF, 1 pixel in 600 DPI or higher printer resolution, result in invisible or very faint lines.

Possibly this PR is kind of a solution. But setting default width to a higher value means, that all lines in new drawings become thicker with zooming, which is probably not wanted by experienced users.

dxli:
Maybe we can treate the default as adaptive:

1, on normal screen 1 pixel;
2, on High Definition 0.1% of width/height in pixels, minimum 1 pixel;
3, printing/exporting at least 2 pixels, or 0.2% of total width.

fa201:
@lordofbikes thicker llines while zooming in is a "normal" behavior so this should not be a limit. I assume most experienced user also assign a standard thickness appropriate to their field before drafting to avoid to do it afterwards.

Assigning a non-zero default value is just a mean to avoid inexperienced users to "lose their lines" while trying to print and come up opening issues or forum question. The design of the application should a bit more fool-proof to ease the learning curve.

I have no experience in architural or electronical designs but for mechanical design, a width of 0.25 mm is appropriate.

dxli:
implemented in master(2.2.2):

PR #1433, opened 2021-11-25
Fixed layer toggle view bug

melwyncarlo:
Without this fix, the programmer gets incorrect layer view status upon reading the layer's visibility via the `isVisibleInLayerList()` method.


melwyncarlo:
This fix causes a problem while freezing all layers together.
This is because `Freeze all layers` (as it is called in the software, which should be `Hide all layers`, as in the Doc) 
point to different (seemingly non-compatible) functions. Refer to #1434.

melwyncarlo:
I'm only working on this issue because its solution helps me fix another issue.

melwyncarlo:
I've made my solution. It helps me to obtain the correct visibility state. Also, it is bug-free, now; it does not block the toggle functions.

PR #1435, opened 2021-11-27
delete nodes from polylines and lines, fixes #1340

### LINKED ISSUE #1340: Delete a point from lines
### Expected behavior

I have a "Z" shaped line (drawn with 3 "l" lines) and i want to be able to delete the top right point and end up with an "L".

grafik

While i understand this is maybe not that easy with "l", it should definitely be possible to select one point from the line with "pl".

Expectation:
"V" drawn with:
- `l`, select the bottom point, delete point = both lines deleted
- `pl` select bottom point, delete point = one straight line from to left to top right

### Observed behavior

I'm not able to select one point at all to delete it, everytime a line is selected. Blue select-box does nothing, green selects both lines. I tried it with "l" and with "pl", its not possible. I am, however, able to move a point around with dragging it, but i fond no intuitive solution for deleting a point.

Draw lines with one point in the middle (=2 lines or 2 line-segments), try to select the middle point and try to delete it.

Version: 2.2.0-rc2
OS: Windows 10

(I tried it on top of an imported image and without an image, no change)

melwyncarlo:
* Solved issue #1340

You can now delete nodes from both polylines and sets of lines, 
via the menu, or by using the `Ctrl+D` shortcut key sequence.

You can also undo/redo the changes made.

Video link :  [https://youtu.be/5V2zIMPchfM](https://youtu.be/5V2zIMPchfM)

### DIFF
--- a/librecad/src/lib/engine/rs_line.cpp
+++ b/librecad/src/lib/engine/rs_line.cpp
@@ -38,6 +38,7 @@
 #include "rs_painterqt.h"
 #include "rs_circle.h"
 #include "lc_rect.h"
+#include "qc_applicationwindow.h"
 
 #ifdef EMU_C99
 #include "emu_c99.h"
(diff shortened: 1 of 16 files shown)


dxli:
Merged

sand1024:
Could you please to review the functionality of this merge once again? 
While the overall intent of this functionality is clear and quite reasonable, I'm not sure it should be merged, due to couple of reasons: 

1) the function seems to be not transparent to the user - for example, there are no prompt hints for command message 

2) There is subtle side-effect - relative point is moved as soon as the user selects line or polyline.  That's contradicts with generic selection of all other entities (say, what about arc that might be also a part of polyline?). Moreover, for creation of entities via Cmd widget (by coordinates) - that may add additional confusion as relative zero moves due to this side effect. 

I think it will be reasonable to revert this merge. 

Instead of this adding this action - it is more practical just expand functionality of original "delete polyline node" action. 

Now I'm working on large pull request that will add more visual interactivity to drawing/editing actions, and I'll address more convenient deletion of polyline nodes there yet without such side effect of changing relative zero.

dongxuli-concord-design:
I tested the basic workflow, and it worked for me. If it can be improved and stabilized, it can be kept in. 

Otherwise, it's better to revert ASAP, before reverting gets complicated due to code change on top of it.

sand1024:
Well, I don't think it's necessary to revert right now, yet I'll take care of this feature.

PR #1436, opened 2021-11-27
save single layer as DXF, fixes #1327

### LINKED ISSUE #1327: Save a single layer as a dxf
I want to be able to create a drawing with multiple layers and export them into individual dxf files so I can import them into KiCad. Currently I have to save a new file with all the layers and then go through and delete any unwanted information.

melwyncarlo:
* Solved issue #1327

Using this tool, you may now save either selected layers or visible layers, either as individual drawing files, or combined into a single drawing file. The layer names and attributes are saved as well.

To accomplish this task, I have also created the `LC_FileDialogService` class. This class centralizes the file I/O user interface. So, basically, this file dialog can be made to serve multiple purposes by tweaking the `*.cpp` file itself, and passing in just one argument: the `FileDialogMode` enumeration. This enumeration tells the class functions how to design the file dialog. So, for example, a specific dialog title, specific file filters, and particularly, in my case, the `Checkbox` element embedded within the file dialog.

<hr>

Video link :  [https://youtu.be/9_qHtZfwMdY](https://youtu.be/9_qHtZfwMdY)

### DIFF
--- a/librecad/src/actions/lc_actionlayersexport.cpp
+++ b/librecad/src/actions/lc_actionlayersexport.cpp
@@ -1,0 +1,271 @@
+/****************************************************************************
+**
+** This file is part of the LibreCAD project, a 2D CAD program
+**
+** Copyright (C) 2021 Melwyn Francis Carlo
+**
+** This file may be distributed and/or modified under the terms of the
+** GNU General Public License version 2 as published by the Free Software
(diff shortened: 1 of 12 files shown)


stonepreston:
This is a very helpful feature, any chance it could be merged?

PR #1437, opened 2021-11-28
copy/paste behavior of blocks, fixes #1344

### LINKED ISSUE #1344: Copy paste block from clipboard adds new block names to block list
Expected Behavour: Pasting blocks from the clipboard creates new instance of the original block name and the block list does not change.

Observed Behavour: Pasting blocks from the clipboard creates a new block name in the block list and the block list grows with every pasted block entity.  If multiple objects are pasted, each new block produces a new name in the block list and the list can grow very large.

To recreate the issue:

1. Create a drawing with just one entity: a single block composed of some primitive entites like arcs and lines, such as a door symbol.

2. In the block list, confirm there is just one block listed.

3. Using the block list, it is possible to insert as many instances of that block as you want and the block list does not change.

4. Using the Move/Copy command and perform a copy (i.e. move, retain original), again, it is possible to make as many new instances of the original block as you want and the block list does not  change.

5. BUT, by selecting an instance of the original block, and performing a ctrl-c operation (i.e. copy to clipboard) and then pasting into the same drawing, a NEW BLOCK is created in the block list.

Environment:
LibreCad Version: 2.2.0-rc2-15-g76828565
Compiler: GNU GCC 9.3.0
Compiled on: May  1 2021
Qt Version: 5.12.8
Boost Version: 1.71.0
System: Ubuntu 20.04.2 LTS

See this forum topic: https://forum.librecad.org/Librecad-keeps-adding-blocks-to-block-list-td5715315.html#a5720508

melwyncarlo:
* Solved issue #1344

New blocks are no longer created (cloned) upon cut/copy/paste; 
just their instances are inserted into the current working layer.

<hr>

Video link :  [https://youtu.be/Aarju_tMJIQ](https://youtu.be/Aarju_tMJIQ)

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -653,8 +653,12 @@ bool RS_Modification::pasteContainer(RS_Entity* entity, RS_EntityContainer* cont
 
     bool pasteToNewDrawing = false;
 
+    double insertionAngle = i->getAngle();
+
     if (bc == nullptr)
     {
+        insertionAngle = 0.0;
+
         pasteToNewDrawing = true;
 
         if (graphic->findBlock(blockName))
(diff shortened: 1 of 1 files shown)


widhi-muttaqien:
This bug returns in version 2.2.1

dxli:
Superseded by PR #1450 which was merged. The feature (showing block/layer name in status bar) is now active in LibreCAD 2.2.1.5+.

PR #1441, opened 2021-11-28
save dimension settings in current drawing preferences, fixes #1329

### LINKED ISSUE #1329: Secure dimension settings in current drawing preferences
This issue was triggered by forum thread https://forum.librecad.org/point-select-not-working-tp5720321p5720351.html
and is related to #1328

### Observed behavior
When both, arrow size AND tick size, are set to 0 in *Current Drawing Preferences*, dimensions become sticky for single click selection.
It's impossible to select any regular entity, always one of the dimensions is selected with a single left click.

Even when this is not the intention of dimensions, this case, arrow size = tick size = 0, should be secured to prevent this unexpected behavior. 

Angular and radial dimensions are already secured and are save for both sizes set to 0.

* create a drawing with some entities
* add some dimensions
* set arrow size and tick size to zero
* try to select one of the non dimension entities

all OS
latest master code

melwyncarlo:
* Solved issue #1329


dxli:
outdated by PR #2030

Issue #1288, opened 2020-11-01
tangential two circles, point circle drawing


orangecrow:
### Expected behavior
Using tan,tan,point on a line and a circle and the point on the line should produce a drawing such that the resulting circle should be tangent to both the line and the circle and touch the line at the point specified on it.

### Observed behavior
The resulting circle goes through the specified point and is tangent to the circle but it is not tangent to the line. In fact it is crossing the line perpendicularly.

### Steps to reproduce or sample file
Draw a circle then a line roughly perpendicular to the circle surface. than use the "tangential two circles, point" option, select the circle, the line, and the point in the middle of the line.

### Operating System and LibreCAD version info
**System:**
Windows 10
**LibreCAD version info:**
Version: 2.1.3
Compiler: GNU GCC 4.9.1
Compiled on: Sep 23 2016
Qt Version: 5.4.1
Boost Version: 1.60.0
System: Windows 8
### Second  Operating System and LibreCAD version info
**System:**
Linux Ubuntu (running on a choromebook thorugh chroot)
**LibreCAD version info:**
Version: master
SCM Revision: 2.0.9
Compiler: GNU GCC 5.3.1
Qt Version: 5.5.1
Compiled on: Mar 24 2016

DuncanLithgow:
Please provide a file or sketch. I can't see how what you want is possible. If it is tangential to the circle and the line what would define the radius of such a circle?

Also you might as well install 2.2rc2

lordofbikes:
The steps are clear, no need for a sample file.
And also the issue is clear.
What I can't say yet, if this behavior is a limitation of the equation solver
or if this special case, when the point is on the line, is intended by design.
My assumption is, that both is true. For any point on the line we have two solutions, how can the tool decide, on which side of the line the tangent circle should be.
For points apart the line, the solution is clear.

jbergengh:
When there are two solutions, as in this case, user should be allowed to choose which solution is desired.  That is how the "Tangent, 2 points" tool handles it.  For example, draw a circle, select the "Tangent, 2 points" tool.  Select two points outside the circle.  There will be two solutions.  LibreCAD shows the centers of both circles and allows the user to select which solution to use.  This issue is the same situation.

PR #1445, opened 2021-11-29
Solved issue #1288

melwyncarlo:
* Solved issue #1288

The line circle tangency problem has been solved by injecting appropriate formulae derived from the line and circle generic equations.

<hr>

Video link :  [https://youtu.be/PoRqCuHhvwA](https://youtu.be/PoRqCuHhvwA)

### DIFF
--- a/librecad/src/actions/rs_actiondrawcircletan2_1p.cpp
+++ b/librecad/src/actions/rs_actiondrawcircletan2_1p.cpp
@@ -28,6 +28,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "rs_graphicview.h"
 #include "rs_commandevent.h"
 #include "rs_arc.h"
+#include "rs_line.h"
 #include "rs_circle.h"
 #include "lc_quadratic.h"
 #include "rs_coordinateevent.h"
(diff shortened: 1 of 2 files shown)


pollywinters:
@melwyncarlo Doesn't work for me, either when merged into my current working copy, or when a direct pull of your Task_32 branch.
The case of your sample video, where the line is perfectly vertical, DOES work.
Using a circle with a non-vertical line passing through it gives no computed solutions for me.
Examining code changes, a check is made for whether one of the base entities is a line, in which case special calculations are performed. Within this section, a particular case is made when the line is perfectly vertical and one equation is used; for non-vertical lines a different equation is used.
I am presuming some error within the equation for the non-vertical line case - when I get enough time I'll examine the details and try to understand exactly how the method works (or fails to work...).

lordofbikes:
Many thanks for testing this @pollywinters.
As I already mentioned in #1288 it's not only a matter of the equation solver.
In the state when the issue occurs, the calculation is done with each mousemove event to show the solution in case of left click.
So moving around gives usually one solution when the mouse is beside the line.
I think therefore the line:
https://github.com/LibreCAD/LibreCAD/blob/f094ab36715d4beb36349ac551011264c6168ae8/librecad/src/actions/rs_actiondrawcircletan2_1p.cpp#L247

It accepts only one solution, it fails for 0, but for 2 also.

When the mouse is on the line we have 2 solutions and in this case a single click is not enough to create the wanted entity.

So as I mentioned in #1288, I'm not sure if the equation solver is the problem at all.
Before putting more effort in this patch, we have to know if `getCenters()` returns 2 for points on the line. Then the equation solver is okay and no need for an alternative calculation.
Having 2 solutions needs a different handling.
- E.g. show both solutions
- left click doesn't create the entities, but fix the tangent point
- then, when the mouse is moved again, show the single solution for the tangent point, depending on the mouse position, right or left of the line

dxli:
Commit 6349126 solves this issue using a simpler change: reject a candidate circle center, if the center is on a selected circle/line.

If we have numeric stability issues, we may resort to method of this pull request

PR #1446, opened 2021-11-29
multiple issues dragging text, fixes #1444

### LINKED ISSUE #1444: Multiple issues when dragging text around
I've been having multiple issues when dragging text to a different location in the drawing.  Sometimes the text disappears when moving it closer to the edge of the screen.  If I zoom out to the point where the text turns into a bounding box, then the bounding box moves twice as far as the text and geometry does.  The video I attached shows those conditions, and I've attached the test file.  Has anyone else experienced this?

Mint 20.2
LC 2.2.0-rc2-20

https://user-images.githubusercontent.com/55407804/143784833-fe89b9f7-ea57-403a-9865-0578e46b8056.mp4
[Text Drag.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/7614313/Text.Drag.dxf.zip)

melwyncarlo:
* Solved issue #1444

### DIFF
--- a/librecad/src/lib/engine/rs_entitycontainer.cpp
+++ b/librecad/src/lib/engine/rs_entitycontainer.cpp
@@ -1679,10 +1679,12 @@ void RS_EntityContainer::move(const RS_Vector& offset) {
         e->move(offset);
         if (autoUpdateBorders) {
             e->moveBorders(offset);
+            e->calculateBorders();
         }
     }
     if (autoUpdateBorders) {
         moveBorders(offset);
+        calculateBorders();
     }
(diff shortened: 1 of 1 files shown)


jbergengh:
Tested and working perfectly here.

lordofbikes:
Even when this patch solves the issue, it needs a deeper review.
The patch is in `rs_entitycontainer.cpp` which is base for other containers like polyline, dimensions or blocks. But these containers don't have the issue described in #1444.
So there may be possible side effects to other containers, maybe reduced performance with bigger containers.

PR #1460, opened 2021-12-18
current layer visibility, fixes #901

### LINKED ISSUE #901: Current Layer Visibility
I just noticed that you can hide a layer, but it still remains active as your current layer.  This means that you can be drawing objects without being able to see them.  I have done a search and have found references to this issue in versions prior to the active layer being displayed on the status bar, which would indicate that this issue might have received consideration and the decision was made to leave it as is?

melwyncarlo:
* Solved issue #901

<hr>

I solved this problem by disabling CAD tools and relevant menus when the currently active layer is either locked or invisible.

<hr>

https://user-images.githubusercontent.com/66683108/146648790-4d9c69bc-53f8-4d55-850a-57d465a88b3b.mp4


dxli:
fix in master and 2.2.1: PR #2746, PR #2747

Issue #1462, opened 2021-12-18
Remote Code Execution vulnerability in LibreCAD 2.2.0-rc3 (JWW CDataMoji)


eldstal:
### Vulnerable Products
* LibreCAD 2.2.0-rc3 and older
* Jw_cad 8.24a and older

### Steps to reproduce or sample file

1. Start LibreCAD 2.2.0-rc3 in a debugger
2. File/Open...
3. Unzip and open the attached [proof of concept file](https://github.com/LibreCAD/LibreCAD/files/7739792/jww_cdatamoji.zip)
4. Observe `ACCESS_VIOLATION` crash, with `eip=0x41414141` (AAAA)

Screenshot:

![x32dbg showing librecad.exe executing at 0x41414141](https://user-images.githubusercontent.com/158607/146652800-e1dfc005-d502-45f9-98e8-bfe09e5ca73b.png)

### Cause
The `CDataMoji` entity deserialization at `LibreCAD/libraries/jwwlib/src/jwwdoc.h` is vulnerable to
a stack buffer overflow.
`char buf[512]` declared in `CDataMoji::Serialize()` on [line 512](https://github.com/LibreCAD/LibreCAD/blob/768285653c46e734a75460928142cda1a33c2c53/libraries/jwwlib/src/jwwdoc.h#L512)
is of fixed size 512. Some varieties of `CDataMoji` provide their own size, e.g. MojiData2 on [line 523](https://github.com/LibreCAD/LibreCAD/blob/768285653c46e734a75460928142cda1a33c2c53/libraries/jwwlib/src/jwwdoc.h#L523)
and no bounds checking is performed. This allows an attacker to overflow `buf` and overwrite other stack variables,
including the return address.

The attached PoC file is tuned to trigger this behavior in the latest windows release of LibreCAD, but the same bug is also present
in older versions and on other platforms.

### Impact
An attacker can craft a JW-CAD input file and thereby gain control over execution flow (EIP controlled directly).

This allows an attacker to run arbitrary code on the system running LibreCAD, with the privileges of the current user.

### Proposed Mitigation
1. Perform bounds checking in `CDataMoji::Serialize()`, and refuse to load the file if it would overflow `buf`.
2. Enable stack smashing protection in the windows build of LibreCAD.

### Operating System and LibreCAD version info
Version: 2.2.0-rc3
Compiler: GNU GCC 7.3.0
Compiled on: Nov 29 2021
Qt Version: 5.12.4
[...]

PR #1463, opened 2021-12-18
Solved issue #1462

eldstal:
Added bounds checking to CDataMoji in jww parser.


lordofbikes:
many thanks :+1:

I merged this, because the fix is obvious.
But the lack of valid test files encourages me to cancel JWW support soon.
I wonder if that would be recognized at all.

Issue #1468, opened 2021-12-19
NULL pointer dereference in DXF parser, HATCH code 93


eldstal:
### Steps to reproduce or sample file
1. Unzip and load the [attached proof of concept file](https://github.com/LibreCAD/LibreCAD/files/7741210/dxf_hatch93.zip) in LibreCAD 2.2.0-rc3

### Cause
The `std::shared_ptr` `DRW_Hatch::loop` is written to when loading a `HATCH` entity with code 93. If this occurs before a code 92, the pointer is still `NULL`, leading to a crash.

### Impact
Denial of service.

### Proposed Mitigation
Ensure that `DRW_Hatch::loop` is not NULL before dereferencing at `drw_entities.cpp:1808`

### Operating System and LibreCAD version info
Version: 2.2.0-rc3
Compiler: GNU GCC 7.3.0
Compiled on: Nov 29 2021
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 10 (10.0)

lordofbikes:
fixed with #1469

eldstal:
This vulnerability has been assigned [CVE-2021-45343](https://www.cve.org/CVERecord?id=CVE-2021-45343).

PR #1469, opened 2021-12-19
Added NULL check for DXF hatch code 93, Fixing issue #1468

eldstal:
This fixes issue #1468

This patch is based on other similar code in the same function, e.g.
```c++
    case 73:
        if (arc) arc->isccw = reader->getInt32();
        else if (pline) pline->flags = reader->getInt32();
        break;
```

The end result is that if code 93 comes before the code 92 which would initialize the `loop` pointer, the code 93 is ignored.


lordofbikes:
many thanks :+1:

Issue #1474, opened 2021-12-20
Crash in DRW_LWPolyline::parseDwg due to inverted bounds check


eldstal:
### Steps to reproduce or sample file
1. Unzip and load the [attached proof of concept file](https://github.com/LibreCAD/LibreCAD/files/7744679/drw_entities_1265.zip) in LibreCAD 2.2.0-rc3

### Cause
The bounds check at [drw_entities.cpp:1264](https://github.com/LibreCAD/LibreCAD/blob/415f8fcbe4e8b256884c17da344b0fadc901a023/libraries/libdxfrw/src/drw_entities.cpp#L1264) is inverted. The `widthsnum` counter read from the DWG document controls this loop, and any `i` which is outside the bounds of the vector `vertlist` will be passed to `std::vector<>::at()` which performs its own bounds checking.

`vector::at()` throws an `std::out_of_range` exception.

The exception goes uncaught and the application crashes.

### Impact
Denial of service.

### Proposed Mitigation
Correct the bounds check on line 1264

### Operating System and LibreCAD version info
Version: 2.2.0-rc3
Compiler: GNU GCC 7.3.0
Compiled on: Nov 29 2021
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 10 (10.0)

lordofbikes:
fixed with #1475

PR #1475, opened 2021-12-20
Correct bounds check in DRW_LWPolyline, fixing issue #1474

eldstal:
This fixes issue #1474, where any use of the vertex list would cause a crash.


lordofbikes:
many thanks :+1:

PR #1486, opened 2021-12-29
glowing effect on mouse over entities, fixes #1480

### LINKED ISSUE #1480: Add glowing effect to lines / parts of drawings when mouse is hovered upon them.
We should have an option to enable/disable glow effect on lines or parts of drawings whenever mouse is hovered upon them. We should also be able to adjust the intensity of the glow. This will be somewhat helpful if we implement the line grids in next versions of LibreCAD.

melwyncarlo:
* Solved issue #1480

<hr>

Video link-1 :  [https://youtu.be/_rR_Ljp9XgM](https://youtu.be/_rR_Ljp9XgM)

Video link-2 : [https://youtu.be/T6yRYs3-PZA](https://youtu.be/T6yRYs3-PZA)


librecad/src/actions/rs_actiondefault.cpp
```diff
@@ -109,16 +109,67 @@ void RS_ActionDefault::keyReleaseEvent(QKeyEvent* e) {
...
+{
+    for (auto entity : *graphicView->getContainer())
+    {
+        if (entity != nullptr)
+        {
+            if (entity->isVisible() && ! entity->isLocked())
+            {
+                double hoverTolerance { 10.0 / graphicView->getFactor().magnitude() };
+
+                if (entity->rtti() == RS2::EntityEllipse)
```
melwyncarlo, line 127: Don't know why the tolerance value for `RS2::EntityEllipse` is interpreted differently in the `RS_Ellipse::isPointOnEntity` function.

That is why I had to set the `10.0` factor to `1.0`, else it detects a far-off mouse position as being _on_ the elliptic entity.

melwyncarlo, line 127: Don't know why the tolerance value for `RS2::EntityEllipse` is interpreted differently in the `RS_Ellipse::isPointOnEntity` function.

That is why I had to set the `10.0` factor to `1.0`, else it detects a far-off mouse position as being _on_ the elliptic entity.

librecad/src/actions/rs_actiondefault.cpp
```diff
@@ -109,16 +109,84 @@ void RS_ActionDefault::keyReleaseEvent(QKeyEvent* e) {
...
+                    hoverTolerance /= 10.0;
+                }
+                else
+                {
+                    if (hoverTolerance < 1.0) hoverTolerance = 1.0;
+                }
+
+                bool isPointOnEntity = false;
+
+                if (((entity->rtti() >= RS2::EntityDimAligned) && (entity->rtti() <= RS2::EntityDimLeader))
```
melwyncarlo, line 138: `RS2::EntityDim...`s and `RS2::EntityMText` don't have the `RS_Entity::isPointOnEntity` function on them. They use the `RS_EntityContainer::getNearestPointOnEntity` function instead.

librecad/src/lib/engine/rs_entitycontainer.cpp
```diff
@@ -229,6 +229,14 @@ bool RS_EntityContainer::toggleSelected() {
 }
 
 
+void RS_EntityContainer::setHighlighted(bool on)
```
melwyncarlo, line 232: There is the overridden function of the `RS_Entity::setHighlighted` function.
Without this, you cannot highlight groups of entities, like `RS_Polyline`, `RS_Dimension`, etc.

ajk2196:
@melwyncarlo , This is what I meant in my [thread](https://github.com/LibreCAD/LibreCAD/issues/1480). Posting here because I could not tag you there. I have attached a picture, modified in Inkscape.

ForLibrecad

melwyncarlo:
@ajk2196 I've replied in your thread.

melwyncarlo:
It officially works on all entities (being a bit presumptuous, here). 

The glow effect is experimental. It's variables can be adjusted easily. 
It's so-called _final state_ has yet to be discussed and decided.

But implementing the feature is now complete.

ajk2196:
Wow @melwyncarlo . Can you please share a video. Also, when will it be implemented into official LibreCAD releases? Because I have been trying, but cannot somehow get the codes to work.

melwyncarlo:
@ajk2196 Thank you very much! :smiley: I've already uploaded a video : [https://youtu.be/T6yRYs3-PZA](https://youtu.be/T6yRYs3-PZA)

Now it should work. I've tested the branch separately.

melwyncarlo:
@jbergengh Thank you very much for pointing out the error! The fix was basically appending a `return;` statement at the end of the _highlight_ operation. That way it no longer loops through deleted entities.

Below is a video where I'm just creating entities, hovering over them and selecting them. No SegFault encountered.
It is specifically the Task_45 branch.

<hr>

https://user-images.githubusercontent.com/66683108/149065447-95191cde-c9c3-456a-ac9d-43cc50c3d006.mp4

jbergengh:
The SegFaults are gone, but I am not seeing any highlighting when in a tool, such as when using the trim tool.

pollywinters:
@melwyncarlo Extra observations added to issue #1480. Basically with deep zooming to inspect small details of drawing / make small adjustments, if the entities on screen have non-zero line width, then the glow effect gets wider as zoom is increased. Width of the glow very quickly fills large areas of the drawing window, preventing any sensible viewing or drawing. Suggest that the glow width requires calculating as on-screen line width **plus** a number of pixels, rather than (what currently appears as) on-screen line width **multiplied** by some factor. Disclaimer - I haven't looked at the actual code yet! Am adding it to my "must have a look at this" list ;)

pollywinters:
@melwyncarlo With further testing and use of this feature, I have found that display of multi-line MText entities is broken. Only just found because I didn't have any drawings with multi-line MText's!
Only the first line of text of an MText is displayed, second and subsequent lines are invisible. The data for all text lines is saved into the dxf files, and does display if the drawing is opened in a version of LibreCAD without the hover/glow feature added.
Bug traced to file "librecad/src/lib/engine/rs_mtext.cpp", changes to RS_MText::draw() method code to make the text "glow".
Screenshot from 2022-05-20 14-07-38
Screen shot shows code with the before/after sections surrounded by an #if 0 / #else / #endif conditional section. When the new version of code is enabled, only the first line of any MText gets displayed, although it does change to highlighted colour and "glow" when hovered over. When the old version of code is enabled, all lines of the Mtext are displayed, they all change to highlighted colour when hovered over, but there is no "glow" effect.

pollywinters:
Testing hover / highlight / glow with other entity types:
NOTE: All in normal view mode, notes on draft mode lower down.

- Lines, rectangles, circles, ellipses = okay
- Polylines (open or closed) = okay
- Arcs, splines, freehand lines = okay
- Text = highlights on hover, does not glow
- MText = see my notes about MText glow bug above.
- Dimensions = okay
- Points (with my new point drawing code, PR #1532) = okay, however note that hover has to be close to the actual point position, while the glyph used to display the point (depending on user settings) can be much larger. Hover is not aware of the displayed glyph size, and takes the point as being conceptually zero size. Hover at the outer parts of the point glyph does not always get picked up as being within highlight distance of the point. I don't see this as a problem, just a bit of a user "gotcha".
- Images = no hover highlight/glow. I don't expect the image itself to highlight, but thought that a bounding frame might highlight. I can see that this might be impractical though.
- Hatches = does not highlight or glow. The boundary line sometimes glows up for a fraction of a second then goes out again. I suspect that the bounding line gets redrawn with the glow, then the hatch gets overdrawn with no glow, effectively hiding the boundary line.

Draft mode:

- No glow effect for any entities, but some do go highlighted on hover.
- Lines, circles, ellipses, arcs, points = seems to be reliable highlight on hover.
- Rectangles, polylines, splines, freehand lines = seems to be a tiny bit of brightening of the line, but not proper highlight. Could well be going to a mixed colour between the entity line colour and the highlight colour. Strangely, when zooming the view the colour becomes more biased towards the highlight shade. 
- Text, MText = text is replaced by a bounding box (as expected), bounding box gives reliable highlight on hover.
- Dimensions = the dimension arrows and extension lines go to the blended entity/highlight colour as per rectangles, polylines etc above. The bounding box that is shown instead of dimension text does not highlight.
- Images = image is replaced by a bounding box (as expected), which does highlight on hover.
- Hatches = hatch is replaced by a bounding box (as expected), but no highlight. When zoomed the bounding box seems to flash to the mixed base/highlight colour for a fraction of a second, but for much shorter time than when in normal view mode.

Note that in both normal view and in draft mode, I have occasionally seen some entities get stuck at the highlight colour, but cannot reliably reproduce or diagnose cause yet. I suspect it is something to do with interaction between entity select and the hover glow while drawing new entities. View/redraw (F5) causes proper redraw of the screen with colours being shown correctly again.

librecad/src/actions/rs_actiondefault.cpp
```diff
@@ -109,16 +111,152 @@ void RS_ActionDefault::keyReleaseEvent(QKeyEvent* e) {
...
+                    if (nearestDistanceTo_pointOnEntity <= hoverTolerance_adjusted) isPointOnEntity = true;
+                }
+                else
+                {
+                    isPointOnEntity = entity->isPointOnEntity(currentMousePosition, hoverTolerance_adjusted);
+                }
+
+                if (isPointOnEntity)
+                {
+                    RS_Color whiteColor(QColor(Qt::white));
```
yekose, line 157: In here the compiler gives an error says qualified name is not allowed, when i changed it to QColor(255,255,255) it worked.

dxli:
merged by d3e13ea00

PR #1487, opened 2021-12-30
New Selection Cursor, and Option to Switch Selection Cursor

melwyncarlo:
Attempted to mimic AutoCAD. Open for suggestions.

<hr>

AutoCAD_Select_Cursor_Small

<hr>

Video link :  [https://youtu.be/T6a51yhDZaE](https://youtu.be/T6a51yhDZaE)


dellus:
As far as I know, the square in the AutoCAD cursor also represents the area of snap search radius. And it's size in pixels can also be set.
In current LibreCAD it's not that obvious where that limit is, seems to vary with zoom factor. You just notice it when snap jumps to a point.
For only selecting an entity I find the current arrow cursor more clear.
The issue has to be considered more thoroughly.

melwyncarlo:
I created it just for selection purpose; though it can be re-purposed. 
I've set it as an option; so cursors may be switched.
I've never used AutoCAD, so I will never know.
Glad to know about the resizing feature; is it overall or just the square?

dellus:
I don't have access to AutoCAD, but I have a German booklet about AutoCAD 12. Here they say you can set the size of the snap window from 1 to 50 pixels. It seems in this old version only a full screen crosshair is available. They also say that the snap window only appears when any snap mode is activated.
I have NanoCAD V5 installed for comparison purposes, available for free. It seems to be a close clone of AutoCAD. Here you can set the crosshair size in %, whereas 100% means fullscreen size, out of the box it is 5%.
They also have a settable  "pickbox", not sure what it is, with default size 4 pixels.
Snap window size is by default 10 pixels, and snap marker size 5.
So quite a lot of possible separate settings.

melwyncarlo:
It can't be 10 pixels; 10 pixels is very very small; perhaps it might be 10 mm? 1 pixel is practically a faint dot.
I can't use nanoCAD either, because I'm on Linux.

<hr>

On this [AutoCAD webpage](https://knowledge.autodesk.com/support/autocad/learn-explore/caas/CloudHelp/cloudhelp/2020/ENU/AutoCAD-Core/files/GUID-363698CF-C3DC-4770-81EF-CB09D86B3D3A-htm.html), at the bottom of the below image, it says `device independent pixels`. 
So, it probably has its own definition for pixels.

image

<hr>

I deduce that the cross-hair and pickbox are designed programatically, and not via PNG or SVG images.
As other custom cursors in LibreCAD used PNG images, I took the same route.

However, to do it the right way, the objectives must first be clearly defined.
E.g., the various cross-hair modes (their description and when they should be activated), 
the various related options to be implemented, etc.

dellus:
I have made screenshots of the 4 pixel pickbox and the 10 pixel aperture (snap window) on my main PC with a 1920x1280 monitor and a very old 1024x768 laptop. Then I cut them out so I could enlarge them and count the pixels. They were the same pixel size on the small and the large screen. The nominal pixel number stretched to both sides of the cross-hair axis, so effectively double size.
The pickbox is the equivalent to the LibreCAD arrow cursor, so what you have provided as an alternative.
The aperture box (snap widow) appears instead when a command using snap (like line from A to B) is activated. Displaying the aperture box can be switched off.
10 pixels might look tiny on a modern high resolution screen, but not on an older one. So good to have the pixel size adjustable.
Regarding crosshair modes, I would support removing the spiderweb if it collides with improvements. I don't see what is is good for, I think it just was a queer fun idea of former developer ravas. I don't believe he will cry. Maybe also remove isometric type.

GSStnb:
The isometric crosshairs should stay.  It is useful when doing an isometric drawing within an orthogonal drawing and when the isometric grid cannot be used. (it's just annoying they are in two different places; App & Drawing Preferences, but that is a different problem.)

melwyncarlo:
@dellus I'm very glad to now know the basic idea behind it. 

The crosshair modes can be kept as options, for those who might want to use them.

dxli:
moved to #2825

PR #1491, opened 2022-01-04
FIx various source comment typos

luzpaz:
Found via `codespell -q 3 -S *.ts,./libraries,*.dxf,*.lff,*.desktop -L ans,atleast,ba,bload,ded,inout,lightyear,recources,trun`

### DIFF
--- a/librecad/src/lib/engine/rs_system.cpp
+++ b/librecad/src/lib/engine/rs_system.cpp
@@ -572,9 +572,9 @@ QStringList RS_System::getDirectoryList(const QString& _subDirectory) {
 #ifdef Q_OS_UNIX
     RS_DEBUG->print( RS_Debug::D_ERROR, "RS_System::getDirectoryList: %s", appDir.toStdString().c_str());
     // for AppImage use relative paths from executable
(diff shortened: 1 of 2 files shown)


librecad/src/lib/engine/rs_system.cpp
```diff
@@ -572,9 +572,9 @@ QStringList RS_System::getDirectoryList(const QString& _subDirectory) {
 #ifdef Q_OS_UNIX
     RS_DEBUG->print( RS_Debug::D_ERROR, "RS_System::getDirectoryList: %s", appDir.toStdString().c_str());
     // for AppImage use relative paths from executable
-    // from packet manager the executable is in /usr/bin
+    // from package manager the executable is in /usr/bin
     // in AppImage the executable is APPDIR/usr/bin
-    // so this should work for paket manager and AppImage distribution
+    // so this should work for package manager and AppImage distribution
```
luzpaz, line 577: Looks like it should be 'package manager' not 'packet manager', right?

lordofbikes:
many thanks :+1:

PR #1498, opened 2022-01-08
copy/paste rotated block, fixes #1284

### LINKED ISSUE #1284: Copy & paste block with angle, duplicate the angle
### Expected behavior
Copy & Peste a block from one drawing to another, copy the block with its attributes, scales, angle, rows and columns and spacing.

### Observed behavior
It dupicates the angle in the paste drawing.

1.- Draw a rectangle in drawing "A", convert to block, rotate 30°
2.- Create a new drawing, "B".
3.- In Drawing "A", select block, hot ctrl-C to copy and select reference point.
4.- In drawing "B", hit ctrl-V to paste and select reference point.
5.- The block is paste, the angle in properties tool is 30°, but is seen as 60°

melwyncarlo:
* Solved issue #1284

<hr>

It rotates it twice, once in the `paste` function, and then again in the `pasteContainer` function.

<hr>

https://user-images.githubusercontent.com/66683108/148631983-de5c2e27-b44d-46fb-88e2-68363da35686.mp4


lordofbikes:
This solution is not acceptable.
Showing properties for the copied block shows the rotation angle.
But properties of the pasted block shows 0° even when it's rotated.
The problem has another cause.

dxli:
I added a similar quick fix for #1616 : 4657adbe2001c2ee9a819e37373a7b015eeb9a25 . The fix has the same drawback of leaving the pasted insert at rotation angle 0. I am still searching for a better fix to leave the insertion rotation angle as the original.

The regression was introduced by d6d5b037ccfd02a

PR #1503, opened 2022-01-08
fillets caused spurious lines, fixes #1496

### LINKED ISSUE #1496: Fillets cause spurious line(s)
### Expected behavior

Clean drawing with rounded corners

### Observed behavior

New lines appear when changing line style

Attached a sequence of operations which cause the effect:
1) Start with a 10x10mm rectangle
2) Add fillets (in my case 4mm)
3) Change properties - line style - to medium dash - new lines appear
4) Change props back to continuous - lines remain (plural - there seem to be a double line)

Linux Slackware64-current
LibreCAD 2.2.0alpha.
Screenshot_2022-01-07_04-50-50

melwyncarlo:
* Solved issue #1496

<hr>

A very amusing fact is that you might think that the problem arises from the function that deals with setting the line type, but no, not at all. It comes from the function that deals with closing the polyline.

This change fixes the mentioned issue; though I am unaware of its implications elsewhere.
But, as it only deals with the extrinsic properties of lines, and _not_ their physical shape, it shouldn't pose any problem.

### DIFF
--- a/librecad/src/ui/forms/qg_dlgpolyline.cpp
+++ b/librecad/src/ui/forms/qg_dlgpolyline.cpp
@@ -81,7 +81,7 @@ void QG_DlgPolyline::setPolyline(RS_Polyline& e) {
 
 
 void QG_DlgPolyline::updatePolyline() {
-    polyline->setClosed(cbClosed->isChecked(), 0.0);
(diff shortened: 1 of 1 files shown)


lordofbikes:
many thanks :+1:

lordofbikes:
many thanks :+1:

PR #1505, opened 2022-01-09
improve arc SVG export, fixes #1493

### LINKED ISSUE #1493: Implement a robust SVG writer / converter.
The current SVG exporter does not export `RS_EntityDim...`s.

Also, it needs to provide option to set image width and height, as well as margins.

Earlier, I thought that the `RS_EntityLine` was not being generated, but it is. 
The problem was that it was created at the left-most point `(0,0)`, even though 
I had drawn the line at position `(80,0)`.

<hr>

Screenshot from 2022-01-06 12-58-19

<hr>

Screenshot from 2022-01-06 13-00-46

<hr>

Screenshot from 2022-01-06 13-00-55

melwyncarlo:
* Solving issue #1493 

<hr>

I wonder if it works on Mac as well.

Still got to look into fixing the broken text.
The text looks broken because the text is basically drawn using `RS_Polyline` entities;
hence, linear graphics like `1` and `7` look great, but other curved graphics look incoherent.

<hr>

A_Crisp_SVG_Arc

### DIFF
--- a/librecad/src/lib/gui/rs_painterqt.cpp
+++ b/librecad/src/lib/gui/rs_painterqt.cpp
@@ -263,19 +263,30 @@ void RS_PainterQt::drawArc(const RS_Vector& cp, double radius,
  * @param a2 Angle 2 in rad
  * @param reversed true: clockwise, false: counterclockwise
  */
-void RS_PainterQt::drawArc(const RS_Vector& cp, double radius,
-                           double a1, double a2,
-                           bool reversed) {
-    if(radius<=0.5) {
+void RS_PainterQt::drawArc( const RS_Vector& cp, 
+                            double radius,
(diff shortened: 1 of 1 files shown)


dxli:
This is a good point: painter should be relying on the native, _i.e._ the QPainter, as much as possible for performance and avoiding issues like #1493.

Similar problem also exists for Ellipses. The current QPainter has an API for complete ellipses, but not for elliptical arcs. Conceptually, we can scale an elliptic arc into an arc, and scale the drawing back to an elliptic.

For the time being, let's fix Arcs first.

This PR doesn't handle reversed arcs properly yet. I'm troubleshooting to locate the reversed part.

dxli:
merged by 2634581

PR #1506, opened 2022-01-09
mouse wheel zoom didn't center on mouse position, fixes #1504

### LINKED ISSUE #1504: Mouse wheel zoom doesn't center at mouse position
### Expected behavior:
Zooming in and out using the mousewheel should center at the mouse position.

### Observed behavior:
Zooming in centers at mouse position only if mouse is at screen center. The more the mouse is off the screen center the zooming behavior gets more unpredictable. Still worse zooming out does behave in a different way so you never get back to your start view. 

### Operating System and LibreCAD version info:
V 2.1.3, V 2.2.0

Also see this discussion: https://forum.librecad.org/Zoom-in-out-with-mouse-wheel-doesn-t-perform-the-same-as-zoom-in-out-with-CTRL-tp5721198.html
In older versions the mouse wheel zoom centers at mouse position and CTRL +/- centers at screen center. Makes sense, doesn't it? Even if the current way is by purpose I propose to return to the previous procedure as I consider it the better solution.

melwyncarlo:
* Solved issue #1504

<hr>

https://user-images.githubusercontent.com/66683108/148685884-7f0e4e20-ba45-419b-a1b9-6d02ba9f04b9.mp4

### DIFF
--- a/librecad/src/ui/qg_graphicview.cpp
+++ b/librecad/src/ui/qg_graphicview.cpp
@@ -634,38 +634,29 @@ void QG_GraphicView::wheelEvent(QWheelEvent *e) {
 		bool invZoom = (RS_SETTINGS->readNumEntry("/InvertZoomDirection", 0) == 1);
 		RS_SETTINGS->endGroup();
 
-		if ((e->delta()>0 && !invZoom) || (e->delta()<0 && invZoom)) {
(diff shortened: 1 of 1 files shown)


melwyncarlo:
Tested Task_55 separately, and it works!

Issue #1476, opened 2021-12-26
Need for Non-scaling dashes' and dots' lines. (Or Proportionately Scaling Dash Lines and Dot Lines)


ajk2196:
In current versions of Freecad, when we choose to make or draw dashed/dotted lines, the dashes and dots are drawn, but they appear to be screenspace scaled (the scale of dashes/dots and space between them remains constant irrespective of actual zoom of the entire drawing), which means that the distances between dots and dashes remains same when scaling the drawing in or out. This visual creates a confusion when small areas represented by dashes are placed on extremely large sized plans. I had an experience where I was working on a project with an area of 122m X 249m. It contained some detached buildings. I had set the arc of door openings to 'dashes'. When I zoomed out to see the entire drawing, the dashes appeared as a continuous arc curve. I panicked, but it did go away when I zoomed and found out about this thing. However, wanting to have perfection in work just like all other architects, I had to check all the doors in the drawing which took up considerable amount of time.

Therefore, kindly please set the dash lines and dot lines to scale proportionately with the drawings and remove the feature of screenspace scaling of dashes and dots. Or atleast give the option to the users to whether have screenspace scaling or proportionate scaling.

Version: 2.2.0-rc3
Compiler: GNU GCC 7.3.0
Compiled on: Nov 29 2021
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 10 (10.0)

dellus:
Linetype is fixed for printout like linewidth, also regardless of scale. It is not reliable in Print Preview either. 
You have to use your imagination to estimate the appearance in the resulting plan print.
 This may help: https://forum.librecad.org/Dash-dot-line-appear-continuous-tp5718594p5718595.html
Look for the link for linetypes.pdf. I could not upload it here.
It is very unlikely your request will be fulfilled for 2.2.0, as the coding effort would go too deep. Maybe V3.

ajk2196:
Having it implemented in V3 would be okay; but if possible, kindly please do it as soon as you good people can. I, posting these many issues feel a bit of guilt: sitting here and just complaining; wish I knew how to code.

[...]

PR #1508, opened 2022-01-12
unit conversion when switching between drawings, fixes #1482

### LINKED ISSUE #1482: Unit Conversion and Implementation: The most serious issue/bug that LibreCAD should never have
First of all, we need to remove the setting to set units of drawings from two options menu and keep it in any one of them. As of now, we can set the unit of drawings (say foot, meter, millimeter) from both Application Preferences and Current Drawing Preferences. A new comer to LibreCAD will only set it from one of them, because that is the case in every application: You don't have to save a same setting at more than one place; once its done, it takes the effect.

**Secondly, we need to make it work**. This is the only setting that is found more than twice in the combined settings options of LibreCAD, yet doesn't work. When I set the units to meters and draw a square of 2x2 units, it gives me the same-dimensioned square that I draw after selecting 'foot' as the default drawing unit from both menus(Application Preferences and Current Drawing Preferences); that is, a 2 sq.ft. square and a 2 sq. mtr. square overlap each other; while in real world case, the 2 sq. ft. square should be smaller than a 2 sq. mtr. square.

**Steps to reproduce:**

Set default drawing unit to meters in both Application Preferences and Current Drawing Preferences.
Draw a 2x2 square using either line tool or rectangle tool.
Now set default drawing unit to foot in both Application Preferences and Current Drawing Preferences.
Again draw a 2x2 square using either line tool or rectangle tool.
You will see both the squares overlapping.

melwyncarlo:
* Solved issue #1482

When switching between drawings, the mouse position and dimensions would, thereafter, adjust themselves accordingly to suit the current drawing sheet's units, subject to toggling off the 'Unitless grid' option in the 'Application Preferences' menu.

Note that the foci of the video are the dimension text label and the coordinate widget.

<hr>

Video link : https://youtu.be/24wm4jFrCDA

### DIFF
--- a/librecad/src/lib/engine/rs_dimaligned.cpp
+++ b/librecad/src/lib/engine/rs_dimaligned.cpp
@@ -34,6 +34,7 @@
 #include "rs_constructionline.h"
 #include "rs_math.h"
 #include "rs_debug.h"
+#include "rs_settings.h"
 
 RS_DimAlignedData::RS_DimAlignedData():
 	extensionPoint1(false),
@@ -106,7 +107,11 @@ RS_VectorSolutions RS_DimAligned::getRefPoints() const
  * measurement of this dimension.
  */
 QString RS_DimAligned::getMeasuredLabel() {
-	double dist = RS_Units::convert(edata.extensionPoint1.distanceTo(edata.extensionPoint2) * getGeneralFactor());
+	double dist = edata.extensionPoint1.distanceTo(edata.extensionPoint2) * getGeneralFactor();
+
+    RS_SETTINGS->beginGroup("/Appearance");
+    if (RS_SETTINGS->readNumEntry("/UnitlessGrid", 1) != 1) dist = RS_Units::convert(dist);
+    RS_SETTINGS->endGroup();
 
     RS_Graphic* graphic = getGraphic();
     QString ret;
--- a/librecad/src/lib/engine/rs_dimaligned.cpp
+++ b/librecad/src/lib/engine/rs_dimaligned.cpp
@@ -106,7 +106,7 @@ RS_VectorSolutions RS_DimAligned::getRefPoints() const
  * measurement of this dimension.
  */
 QString RS_DimAligned::getMeasuredLabel() {
-	double dist = edata.extensionPoint1.distanceTo(edata.extensionPoint2) * getGeneralFactor();
+	double dist = RS_Units::convert(edata.extensionPoint1.distanceTo(edata.extensionPoint2) * getGeneralFactor());
 
     RS_Graphic* graphic = getGraphic();
     QString ret;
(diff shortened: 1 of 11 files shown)


melwyncarlo:
I have tested this branch separately, and it works all well!

ajk2196:
Thanks for the video link. Liked it a lot that you have added the glowing effect. However, if you can post one more video where you carry out the same experiment with more other units like feet, yards and so on; it would be amazing; for we don't know when this feature will be added officially and experiment it out on our own.

Moreover, I have three more suggestions; but cannot raise an issue as your feature hasn't been implemented yet; but adding these next features will be helpful to architects.

1. When adding dimensions; include the units as well. In your demonstration video, it only shows 100 units or 3.9370. It would be great if it said 100 mm or 3.9370 in. Now while adding the units, it would be helpful if we add the boxing to the dimensions as shown in the graphic below. Below showed way of dimensioning will be helpful when there is a great clutter of dimensions in a large drawing. I saw one of my seniors do this; he only worked with hand-drawn drawings.
delforgit

2. In the main drawing window, display in top right corner the units we are working on.

3. Make one single drawing contain multiple dimensional units. When implementing this; we will have a bug where upon reopening a saved file with multiple dimensions - will open up with different dimensions converted to one default-set unit. So please take care of that.

And @melwyncarlo ; can't thank you enough. Like always, you are a blessing to this team.

jbergengh:
Placing a box around a dimension like that has a special meaning.  It is part of the GD&T specification.  It means the dimension is a "basic" dimension, having no tolerance.  The tolerance is then added using tolerance zones via a feature control frame.  It's not something that is used to reduce clutter!!!

dellus:
@ ajk2196: We should not ask too much from melwincarlo, we can be glad for what he has already done.
Also LibreCAD should not become too complicated and bloated.
Adding the unit to the dimension text is not customary in architectural drawings, at least not in Europe (I am an architect). It can even be a problem as it uses space when it is cramped.  
I have never seen those boxes in architectural drawings. Also introducing this feature might be a not so trivial task.
melwincarlo's efforts should better be spent on more important issues benefitting a wider circle of users.

GSStnb:
@dellus @ajk2196 It is not something (the boxes) I've seen on architectural / construction / fabrication drawings in North America either.  Nor having the unit of measure inline with the dimensions.  It would be determined from the title block.

ajk2196:
Ok then, its good. Maybe we haven't come across a problem which requires this . I will try finding the document in which we did this, and post it here after getting the permission. It may belong to GD&T, but it is a good workaround to stay free of confusions when we have a lot of dimensions tucked at one side of the drawing. In cases like those and in the drawings, if you take in human error in consideration from the architects' side, can do blunders.

In our case, as far as I remember, it was the drawing of a facade or some type of distorted triangular grid. Can't be sure because I could only have a glance over it for a few seconds and didn't know then that it would be helpful here. 😅 Let the efforts be invested somewhere else; I too am of that same opinion. Hope some of the efforts are put into this:

[Non scaling dashes](https://github.com/LibreCAD/LibreCAD/issues/1476)

dxli:
merged

PR #1516, opened 2022-01-20
don't scroll layer widget after toggling layer visibility, fixes #1515

### LINKED ISSUE #1515: Toggling layer visibility scrolls layer list to active layer
Currently layer list scrolls back to the bottom of the list (the active layer). In my opinion it would be better to leave the scroll position as-is. If a drawing contains large number of layers (twenty or more), than it is hard to explore the drawing, because if wrong layer was hidden, then the user needs to scroll back to its position in layer list and to toggle its visibility again. If the scroll position stayed as is, than it would be much quicker.

### Expected behavior
If layerwidget contains many layers (scroll bar is visible), then simple toggling of layer visibility shouldn't change the scroll position.

### Observed behavior
Toggling layer visibility (or lock, print, construction layer) resets the scroll position to the active layer. The behaviour is irritating especially for drawings containing large number of layers (e.g. maps) opened for exploration.

Steps to reproduce:
1. Add enough layers for layerwidget to show scroll bar.
2. Scroll to the bottom of the layer list.
3. Select (activate) last layer.
obraz
4. Scroll the list to first item.
obraz
5. Toggle first item visibility (toggling lock, print and construction has the same effect).
obraz

Fedora Linux 35 (x86_64), LibreCAD commit 5d331c8bf51031f46989a5ce3350c59508505289 (almost certain that issue is also present in 2.2.0-rc3).

michalg0x5a:
Before the change, toggling layer visibility (or in essence updating
contents of layerwidget) caused it to scroll to the active layer. Such
behaviour is problematic especially for drawings with large number of
layers.

Fix is simple - don't set active layer during layerwidget update.

### DIFF
--- a/librecad/src/ui/qg_blockwidget.cpp
+++ b/librecad/src/ui/qg_blockwidget.cpp
@@ -256,7 +256,6 @@ QG_BlockWidget::~QG_BlockWidget() {
 void QG_BlockWidget::update() {
     RS_DEBUG->print("QG_BlockWidget::update()");
 
-    int yPos = blockView->verticalScrollBar()->value();
     RS_Block* activeBlock;
 
     if (blockList) {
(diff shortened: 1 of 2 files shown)


lordofbikes:
many thanks :+1:

Issue #1525, opened 2022-03-06
Mirror of an image works incorrectly


adamzimny:
### Expected behavior
Image is mirrored correctly:
![image_mirror_correct](https://user-images.githubusercontent.com/3201093/156923104-ba109870-03bd-458c-b5a8-a48533d9c4c7.png)

### Observed behavior
Image is rotated and offset instead of mirrored. Also, when selected, the selection border is offset from the image itself.
![image_mirror_bug](https://user-images.githubusercontent.com/3201093/156923130-84776018-ec49-40de-a2d4-75f480211a94.png)

### Steps to reproduce or sample file
Add an image and make a mirror of it by using "Modify" toolbox and selecting "Mirror" action.

### Operating System and LibreCAD version info
Linux,  2.2.0-rc3-24-g4602ec6a

adamzimny:
Fix proposed here: https://github.com/LibreCAD/LibreCAD/pull/1526

PR #1526, opened 2022-03-06
Fix image mirroring

adamzimny:
Image was rotated and offset instead of mirrored.
Also, when mirrored image was selected, the selection border was
offset from the image itself.

Fixes issue #1525

### DIFF
--- a/librecad/src/lib/engine/rs_image.cpp
+++ b/librecad/src/lib/engine/rs_image.cpp
@@ -369,11 +369,18 @@ void RS_Image::scale(const RS_Vector& center, const RS_Vector& factor) {
 
 
 void RS_Image::mirror(const RS_Vector& axisPoint1, const RS_Vector& axisPoint2) {
+    RS_Vector vp0(0., 0.);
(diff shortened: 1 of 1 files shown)


lordofbikes:
This solution doesn't withstand a close/open cycle.
The images mirror correct, but on reopen, even when they are on the correct position, the are no longer mirrored.
A quick look into the DXF reference showed no flag for mirrored images. So this seems not possible to me yet.
Maybe I've overlooked something, I'll check this later again.

adamzimny:
True, it did not withstand close/open cycle :(
Yes, DXF does not have a mirrored flag. It is not needed though, the orientation of U-V vectors (or rather their "handedness") is enough to tell if image is original or mirrored:
- original image - if V is pointing upwards then U is pointing to the right
- mirrored image - if V is pointing upwards then U is pointing to the left

This "handedness" can be detected when you do a cross product of U and V and check the sign of the Z component.

New attempt is taking a different approach - instead of modifying how image is mirrored (which was actually ok) is rather modifying how image is drawn. 
Close/open works fine now.

PR #1528, opened 2022-03-13
renaming of nested blocks, fixes #1527

### LINKED ISSUE #1527: Rename nested block not work
### Expected behavior
Rename nested block rename all block name in dxf file

### Observed behavior
Rename nested block rename block name in dxf file
In dxf file the last occurence of block is not renamed
Rename last occurence of block directly in dxf file will work as expected

Import a dxf to block with a nested block
Rename a nested block 
Save and quit
Re-open and the nested block is empty

Version: 2.2.0-rc3
Compiler: GNU GCC 7.3.0
Compiled on: Nov 29 2021
Qt Version: 5.12.4
Boost Version: 1.65.1
System: Windows 7 SP 1 (6.1)

adamzimny:
Block list fix:
When block A that is nested within block B is renamed, then rename
also all inserts of block A withing block B in the blocklist.

Entity container fix:
RS_Insert is a subclass of RS_EntityContainer. Inserts within
inserts (in entity tree of a graphic) should be renamed as well.

Fixes issue #1527

### DIFF
--- a/librecad/src/lib/engine/rs_blocklist.cpp
+++ b/librecad/src/lib/engine/rs_blocklist.cpp
@@ -173,8 +173,15 @@ void RS_BlockList::remove(RS_Block* block) {
 bool RS_BlockList::rename(RS_Block* block, const QString& name) {
 	if (block) {
 		if (!find(name)) {
+			QString oldName = block->getName();
 			block->setName(name);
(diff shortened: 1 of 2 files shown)


lordofbikes:
many thanks :+1:

PR #1537, opened 2022-04-10
Fix typos

luzpaz:
Found via `codespell -q 3 -S *.ts,./libraries,*.dxf,*.lff,*.desktop -L ans,atleast,ba,bload,ded,inout,lightyear,recources,trun`


lordofbikes:
many thanks :+1:

PR #1432, opened 2021-11-25
Update tab order qg_dlgtext.ui

MarcelloTheArcane:
The tab order of elements in the text dialog box was a bit counterintuitive, so I've added a tab order.  Hopefully it's logical:

image

PR #1540, opened 2022-04-24
Ref PR #1432 - Update tab order in qg dlgtext.UI - also add "cbFont" selector into the tab order

pollywinters:
Addition to PR #1432 - Update tab order in qg dlgtext.UI
PR #1432 by [MarcelloTheArcane](https://github.com/MarcelloTheArcane) did not include the Font selector widget in the tabbing order. Hence not possible to change the base font by tabbing through the dialog options.
This PR includes the Font selector (dialog itemname "cbFont") at tabbing order position 9:
Screenshot from 2022-04-24 22-09-40
.


pollywinters:
Further update to tab order in Text dialog, also set the correct tab ordering for the MText dialog. Previous changes for the Text dialog omitted the tab ordering for the fields within the Pen compound widget, which appears within the Text dialog when using the Properties tool on an existing Text entity.
When including the tab ordering for the Pen compound widget it becomes sensible to also fix up the MText dialog within this same PR, because both of the dialogs share the use of the Pen widget, with a common change for both dialogs.
Note that it seems that the only reliable way of ensuring the tabbing order is correct is to set it programmatically by code within the respective .cpp files, as when it is just set by <tabstop> entries within the .ui files, the entry for the Pen compound widget tab keeps getting erased whenever the Edit Tab Order tool in Qt Designer is used on these files. The .ui files can be manually edited, but then if Qt Designer is used again to alter the layout, the Pen widget gets dropped out of the order again.

pollywinters:
Note that my rework on the .ui files for the Text and MText dialogs includes some minor tidying of layout in order to improve the vertical alignment of fields when using the Properties tool on existing Text and MText entities, where the Layer and Pen fields become visible. No major change, I suspect that few would notice unless directly comparing "before" and "after" screen shots.
Screen shots of modified dialogs attached:
1. Text dialog when creating a new Text entity.
Screenshot from 2022-05-16 23-51-31
2. Text dialog when changing Properties of an existing Text entity.
Screenshot from 2022-05-16 23-50-19
3. MText dialog when creating a new MText entity.
Screenshot from 2022-05-16 23-51-52
4. MText dialog when changing Properties of an existing MText entity.
Screenshot from 2022-05-16 23-49-54

PR #1484, opened 2021-12-29
new DimArc tool, fixes #1255

melwyncarlo:
* **DimArc tool** (similar to its AutoCAD counterpart)
* Solved issue #1255

<hr>

DimArc_Icon

<hr>

Video link on the DimArc tool :  [https://youtu.be/JE0oe5TS-ds](https://youtu.be/JE0oe5TS-ds)

Video link on DimArc Transformations :  [https://youtu.be/NAHlhFgaJ9E](https://youtu.be/NAHlhFgaJ9E)

<hr>

The `DimArc` feature saves in various image formats, as well as the PDF format.

Unfortunately, for now, it does not save in the `DXF` format, so when you save the file, 
the `DimArc` feature will _not_ get saved alongside other entities.

I've posted an issue regarding this matter in the `libdxfrw` repository. 
Link :  [https://github.com/LibreCAD/libdxfrw/issues/57](https://github.com/LibreCAD/libdxfrw/issues/57)

Issue #1255, opened 2020-07-02
DimArc does not support?


zhengf312:
### Observed behavior

I can't find dim Arc function in LibreCAD.  Does it not support?

zhengf312:
dimension Arc length

DuncanLithgow:
You are correct that this is missing.

PR #1551, opened 2022-06-26
Change DXF Export to use ext Style handle ID for 340 Dimstyle field.

computergeek1507:
Dimstyle should use the Text Style handle ID not the name.

http://docs.autodesk.com/ACD/2011/ENU/filesDXF/WS1a9193826455f5ff18cb41610ec0a2e719-7a53.htm

### DIFF
--- a/libraries/libdxfrw/src/libdxfrw.cpp
+++ b/libraries/libdxfrw/src/libdxfrw.cpp
@@ -314,15 +314,17 @@ bool dxfRW::writeLayer(DRW_Layer *ent){
 
 bool dxfRW::writeTextstyle(DRW_Textstyle *ent){
     writer->writeString(0, "STYLE");
+    //stringstream cause crash in OS/X, bug#3597944
+    std::string name=ent->name;
+    transform(name.begin(), name.end(), name.begin(), toupper);
     if (!dimstyleStd) {
-        //stringstream cause crash in OS/X, bug#3597944
-        std::string name=ent->name;
-        transform(name.begin(), name.end(), name.begin(), toupper);
-        if (name == "STANDARD")
+        if (name == "STANDARD"){
             dimstyleStd = true;
+        }
     }
     if (version > DRW::AC1009) {
         writer->writeString(5, toHexStr(++entCount));
+        textStyleMap[name] = entCount;
     }
 
     if (version > DRW::AC1012) {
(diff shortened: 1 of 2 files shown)


pollywinters:
Interesting. Would you mind if at some time I pick your brain for other details surrounding the DXF storage of dimensioning info? I have been independently doing some work around issue #1255, to support the Dim Arc dimension. 
In pull request #1484 @melwyncarlo has got as far as a trial implemention of a Dim Arc tool and display of the entity, but does not yet implement storage of the data for this as the DXF library does not currently support it. I have got about 99% of a storage implementation done, but still only privately, not yet at a state where I feel happy to release a pull request for it. Part of the problem is that the Autocad released spec for DXF format does not document this feature at all, and I have had to do a bunch of deduction by cross reference to the DWG spec which does include minimal info about it, and manual inspection of some sample DXF files that include Dim Arc entries. Some of the stuff I've done is pure guesswork!
My implementation seems to work on my local build, but it has involved quite a lot of enhancement to the DXF library code. It stores Dim Arc entities in drawings, and loads them back again in my LC build, but that only means that the changes are self-consistent. It's more than likely that my implementation of the feature will not be compatible with Autocad (which I don't have any access to for testing) or other DXF based CAD packages. And possible that I will have broken compatibility even when not using Dim Arc...
Unfortunately I am at present short of time for doing LC work (pesky real life getting in the way!) but hope to be back onto it maybe mid-July.
It is very likely anyway that these changes will never get released in LC version 2, where stability is more important than implementation of new features, but my hope is that they will provide a useful demonstration towards being able to add them to the new LC version 3 code.

computergeek1507:
I use the netDxf C# library to based my code off. I was creating DXF with LibreCAD and it was causing issues with netDxf. In LibreCAD, currently a string is bring written with the 340 code and the spec says it should be a Hex ID to the style handle.

https://github.com/haplokuon/netDxf/issues/228
image


I have one DXF from AutoCad and the 340 is a Hex ID.
image

pollywinters:
@computergeek1507 thank you, I'll take a look at how that library handles some of the dimensioning issues.

computergeek1507:
can this be fixed for v2 or does it have to be pushed to v3?

lordofbikes:
Many thanks @computergeek1507 for this contribution.
I removed V3 and added pending, as it is not decided yet.

Also, this is pure `libdxfrw`, so when fixed upstream, it will be available for both, V2 and V3.

I did not a detailed review yet, but what I see is that it lacks modifications in the reading part. When we change the output we must assume, that we will open both cases in future, strings and hex handles. This must be detected and handled correct on opening files.
We have to check if this can be done from DIMSTYLE record itself or if we must change the DXF version number for a safe detection.

We are close to release 2.2.0, so I only want to merge things which are considered to be complete and save. But there are lots of other pull requests which will be merged for a 2.2.1 release later, so I think yours too.

dxli:
I added the change to libdxfrw ( 0212aaa6d54d47330881206a078c8683be01e3c5 )

Let's merge this one also (no merge conditions).

Issue #1007, opened 2018-08-11
Provide a Flatpak as software package — for Linux systems


x80486:
[Flatpak](http://docs.flatpak.org/en/latest/introduction.html) is a technology for building, distributing, installing and running applications – primarily targeted at the Linux desktop and desktop applications.

gerardbm:
Related: https://github.com/LibreCAD/LibreCAD/issues/858

tinywrkb:
[Here's an initial Flatpak manifest](https://github.com/tinywrkb/flatpaks/blob/master/org.librecad.librecad/org.librecad.librecad.yaml) to start from.  
Packaging adapted from the [Arch Linux package](https://github.com/archlinux/svntogit-community/blob/packages/librecad/trunk/PKGBUILD).  
I don't use the app so if the devs are not interested in this then the package needs a maintainer that is familiar with the app.

The package can be built easily with flatpak-builder.

Cliveptr:
Supporting Flatpak could also resolve #894 (no PPA maintainer) - everyone would have access to the latest release without having to manually build it every time.

@tinywrkb how functional is your manifest?

tinywrkb:
@Cliveptr maybe try it? you just need flatpak and flatpak-builder packages.

```
# install the KDE SDK to the user or system Flatpak installation 
flatpak install flathub org.kde.Sdk//5.15

# get the sources
git clone https://github.com/tinywrkb/flatpaks.git

# build and install the app to the user Flatpak installation
cd org.librecad.librecad
flatpak-builder --user --force-clean --install build org.librecad.librecad.yaml

# run the app
flatpak run org.librecad.librecad.yaml
```

MartinIIOT:
Please, publish LibreCAD via Flatpak.
Thank you.

hfiguiere:
Flatpak submitted.

Let use know who shall be granted write permission as upstream contributor. (it's not a committment from you to maintain it)

lordofbikes:
Many thanks!
That's probably me.

x80486:
`LibreCAD` is now in `Flathub`: https://github.com/flathub/org.librecad.librecad

PR #1580, opened 2022-12-19
Issue #1007 - Update appadata to pass validation

hfiguiere:
- this is required for flathub submission
- this is in no way specific to flathub

(this doesn't close the issue)


lordofbikes:
many thanks :+1:

PR #1584, opened 2022-12-30
fix versioning to reflect 2.2.0 final

spotrh:
Signed-off-by: Tom Callaway <spot@fedoraproject.org>

I noticed in updating to 2.2.0 final that LibreCad was still self-reporting "2.2.0-alpha" (Help->About). This fix to librecad/src/src.pro should resolve that.

### DIFF
--- a/librecad/src/src.pro
+++ b/librecad/src/src.pro
@@ -10,7 +10,7 @@ DISABLE_POSTSCRIPT = false
 DEFINES += DWGSUPPORT
 DEFINES -= JWW_WRITE_SUPPORT
 
-LC_VERSION="2.2.0-alpha"
+LC_VERSION="2.2.0"
 VERSION=$${LC_VERSION}
 
 # Store intermedia stuff somewhere else


lordofbikes:
The modified line just represents the fallback and it is intended, that it shows a label that does not represent a current tag.
It is intended, that `LC_VERSION` is overwritten here:
https://github.com/LibreCAD/LibreCAD/blob/f55f341e69b70c2c8ceaf73ad594261406507349/librecad/src/src.pro#L43
The requirements are a `git` executable somewhere in `PATH` and the folder `../../.git` and that the repo was pulled with `--tags`.
Then a valid version label is created from `git describe`, which shows exact source code version including git hash in case it is not a tagged commit.

Is this possible in the Fedora build system?

Otherwise we should think about a better solution to set a version label and distinguish if it is only fallback or a valid source code version.

spotrh:
Hm, okay. It seems odd to depend on git for that, I'm not sure how useful the git hash is to an end-user, but ... okay, that's a style choice and I won't quibble over that.

I could certainly add git as a BuildRequires...however, the GitHub source tarball for Librecad that we're using (https://github.com/LibreCAD/LibreCAD/archive/4b91d9b0f919be41f7e7568c87c5c67dfac189aa/librecad-4b91d9b.tar.gz) does not seem to have the ".git" folder in the topdir (LibreCAD-4b91d9b0f919be41f7e7568c87c5c67dfac189aa).

Fedora tries not to make its own tarballs when we can either pull directly from upstream/GitHub.

dxli:
we should keep releasing 2.2.x maintenance releases, unless our version 3 branch picks up momentum.

dxli:
Not needed with 2.2.1 and 2.2.2 releases planned

PR #1622, opened 2023-06-07
Hide zero foot

dxli:
This is PR #1390 by [Nick Deubert] (https://github.com/ndeubert)

Hide zero foot in the architecture format for zero foot.


ndeubert:
Thanks!

PR #1626, opened 2023-06-08
Enable MSVC compiler

madmaxoft:
These changes are needed in order to compile using MSVC 2022:
- Library dependencies need to be specified as `*.lib` files, rather than `lib*.a`
- `-std=c++11` flag is not understood by MSVC
- `struct` and `class` are not interchangeable between declarations and definitions, it results in function signatures being different in the header than the cpp file, ending in "undefined symbol" errors from the linker
- `char **` and `char* []` are not the same, again resulting in "undefined symbol" errors from the linker.
- Fixed a qmake warning when `BOOST_DIR` environment variable is not set

Also corrected the discrepancy with custom.pro, this file should contain local settings only and so it shouldn't be in the repo at all.


librecad/src/lib/modification/rs_modification.h
```diff
@@ -44,7 +44,8 @@ class RS_GraphicView;
 /**
  * Holds the data needed for move modifications.
  */
-struct RS_MoveData {
+class RS_MoveData {
+public:
```
dxli, line 48: is this necessary?

Following C++ core guide lines: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-struct

Using struct is reasonable here, each member is logically independent and this no clear class/struct wide invariant

librecad/src/main/console_dxf2png.cpp
```diff
@@ -79,7 +79,7 @@ bool slotFileExport(RS_Graphic* graphic,
 /// \param argv
 /// \return
 ///
-int console_dxf2png(int argc, char* argv[])
+int console_dxf2png(int argc, char** argv)
```
dxli, line 82: `char** argv` is equivalent to `char* argv[]` as a function parameter

librecad/src/lib/modification/rs_modification.h
madmaxoft: Yes, the code needs to choose either a `class` or a `struct` for these and stick to it everywhere. As noted in the description, MSVC treats `class` and `struct` as different things in function signatures, so a function declaration in a header file that sees only the `class RS_MoveData;` forward declaration will not match the function definition in cpp file which sees the `struct` definition.
This actually had me headscratching for an hour, I couldn't find out why the linker was complaining about undefined symbols when "clearly" they were defined.

libraries/muparser/muparser.pro
```diff
@@ -17,8 +17,6 @@ TARGET = $$DLL_NAME
 GENERATED_DIR = ../../generated/lib/muparser
 # Use common project definitions.
 include(../../common.pri)
-QMAKE_CXXFLAGS_DEBUG += -std=c++11
-QMAKE_CXXFLAGS += -std=c++11
```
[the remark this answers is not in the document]
madmaxoft: WHatever the plan is, the flags are already set in c../../common.pri, so there's no need to set them again here.

librecad/src/lib/engine/rs_patternlist.h
```diff
@@ -61,7 +61,7 @@ class RS_PatternList {
 	void init();
 
 	int countPatterns() const {
```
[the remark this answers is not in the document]
madmaxoft: Most Qt classes have an `int`-returning `count` method, so I didn't question this at all, it's just the path of least resistance to get rid of a warning.
This change is not essential and can be dropped from the PR, if needed.

librecad/src/lib/modification/rs_modification.h
dongxuli-concord-design: Could you share some exact error messages to show the issue? Struct and class are always to distinct types in function signatures. There could be some incorrect forward declarations, which should be fixed instead.
The current design pattern of entity classes is by composite a class based data struct, which is POD, with no internal logic or invariant, therefore, as structs.
If you find error in linking to a class of such a data type, say, "class FooData". please search the source code for "class FooData", and replace all forward declarations to "struct FooData;". I do expect this would be the actual fix.

librecad/src/lib/engine/rs_patternlist.h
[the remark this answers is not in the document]
dxli: This is good. Just wondering whether we add some protection against signed integer our of range assigning (undefined behavior).

librecad/src/main/console_dxf2png.cpp
dxli: (Msvc main signature)[https://learn.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=msvc-170]

This is puzzling. `int foo(char **)` is the same as `int foo(char** const)`.

Again, could you share the exact linker error messages.

I may still have msvc on some old laptops. I can give it a try also.

librecad/src/lib/modification/rs_modification.h
madmaxoft: Sure. First there's several warnings:
```
[...]\LibreCAD\librecad\src\lib\gui\rs_dialogfactoryinterface.h(55): warning C4099: 'RS_MoveData': type name first seen using 'struct' now seen using 'class'
[...]\LibreCAD\librecad\src\lib\modification\rs_modification.h(47): note: see declaration of 'RS_MoveData'
```
And then there's the linker error:
```
qc_dialogfactory.obj : error LNK2001: unresolved external symbol "public: virtual bool __cdecl QG_DialogFactory::requestMoveDialog(class RS_MoveData &)" (?requestMoveDialog@QG_DialogFactory@@UEAA_NAEAVRS_MoveData@@@Z)
..\..\windows\LibreCAD.exe : fatal error LNK1120: 1 unresolved externals
```
This results just from changing `class RS_MoveData` to `struct RS_MoveData`.

The problem is that there are forward declarations using `class`, while the definition is using `struct`. Based on what @dxli is saying, I believe the proper fix is to change the forward declarations to `struct`, rather than changing the definition to `class`; I'll change my PR to that if you agree.

librecad/src/main/console_dxf2png.cpp
madmaxoft: This is the error I was getting before this change:
```
main.obj : error LNK2019: unresolved external symbol "int __cdecl console_dxf2png(int,char * *)" (?console_dxf2png@@YAHHPEAPEAD@Z) referenced in function main
  Hint on symbols that are defined and could potentially match:
    "int __cdecl console_dxf2png(int,char * * const)" (?console_dxf2png@@YAHHQEAPEAD@Z)
..\..\windows\LibreCAD.exe : fatal error LNK1120: 1 unresolved externals
```
And in any case, I believe the function declaration should match the function definition.

Weirdly enough, `console_dxf2pdf` seems to be the same, the definition doesn't match the declaration, but in this case the compiler / linker do not mind. I have no idea what's going on here, but since fixing the mismatch has fixed the error, I just went with it.

madmaxoft:
So what's the plan, should I rebase my PR onto current master, or are you manually picking stuff to fix?

dongxuli-concord-design:
Hi Matte,

I have cloned your branch and rebased to the master.

I will keep all data classes as struct, and fix any incorrect forward declarations of structs as classes. 

Since the change is not risky to me. I am thinking about pushing it from my branch. Please test to make sure that msvc works with my commit(not ready yet).

Thanks a lot. 

Dongxu

dxli:
Hi Mattes,

Could you test commit 20c59a7 for MSVC support.

Thanks!

Dongxu

madmaxoft:
The commit builds without errors, only some 645 warnings are left, most of them related to 32-bit vs 64-bit (I'm compiling for 64-bit) or complaining about using `sprintf` and `strcpy`.

However, the produced executable seems to run into heap corruption upon start. That is most likely introduced by a recent commit, since that didn't happen at the tip of my PR.
Here's the log I get in the console:
```
[20230609 17:00:16.324 W 40936] QLayout::QLayout - kernel\qlayout.cpp:148 - QLayout: Attempting to add QLayout "" to QG_LibraryWidget "Library", which already has a layout
HEAP[LibreCAD.exe]: HEAP: Free Heap block 00000016B478ABB0 modified at 00000016B478ABF0 after it was freed
```
Here's the stack trace at the time:
```
12 _q_resolveEntryAndCreateLegacyEngine_recursive       qfilesystemengine.cpp    150 0x7ffe21985cd4 
13 QFileSystemEngine::resolveEntryAndCreateLegacyEngine qfilesystemengine.cpp    196 0x7ffe219853be 
14 QFileInfoPrivate::QFileInfoPrivate                   qfileinfo_p.h            105 0x7ffe2193ac77 
15 QFileInfo::QFileInfo                                 qfileinfo.cpp            347 0x7ffe21933f0e 
16 QPixmap::load                                        qpixmap.cpp              738 0x7ffe2248072f 
17 QPixmap::QPixmap                                     qpixmap.cpp              205 0x7ffe2247dadb 
18 QG_ColorBox::addColor                                qg_colorbox.cpp          226 0x7ff7de9f8978 
19 QG_ColorBox::addColors                               qg_colorbox.cpp          86  0x7ff7de9fa2a6 
20 QG_ColorBox::init                                    qg_colorbox.cpp          131 0x7ff7de9f9894 
21 QG_ColorBox::QG_ColorBox                             qg_colorbox.cpp          78  0x7ff7de9f886f 
22 QG_PenToolBar::QG_PenToolBar                         qg_pentoolbar.cpp        40  0x7ff7dea2b8b2 
23 LC_WidgetFactory::createStandardToolbars             lc_widgetfactory.cpp     395 0x7ff7deb36f1a 
24 QC_ApplicationWindow::QC_ApplicationWindow           qc_applicationwindow.cpp 241 0x7ff7deb7c7ab 
25 QC_ApplicationWindow::getAppWindow                   qc_applicationwindow.cpp 374 0x7ff7deb7a557 
26 main                                                 main.cpp                 276 0x7ff7deba5511 
```
The failure is reproducible, it happens on each start, but not always at the same place.

madmaxoft:
I've made a lucky guess and reverted 716bf87315480da7f58c2c24f74d80d1f7636a16. That fixed the heap corruption.

dongxuli-concord-design:
> I've made a lucky guess and reverted 716bf87315480da7f58c2c24f74d80d1f7636a16. That fixed the heap corrupt

Does the corruption still happen with the latest master head. I am pretty sure, it's fixed. The corruption was due to invalid usage of initializer_lists. I just modified the functions to use vectors instead

madmaxoft:
I can't build because of 4fa20f7db90705ef30705857d404756e6ae48ad0; after fixing the build problems, the heap corruption is still present. I've created a PR against an older version of the code that fixes the corruption: #1628.

PR #1673, opened 2023-08-03
feat: Shortcut "." to specify current point coordinates on command line

gwhitney:
In using the LibreCAD command line, I find that far and away the most common point that needs to be entered is simply the current point (the origin for relative coordinates). I am aware that "0..0" specifies this, but that is four fairly awkward keystrokes. Hence, I suggest that a single-character shortcut is well worth it. This commit implements "." to mean the current point (i.e. same as "0..0") when coordinates are expected on the command line. The shortcut was chosen because a bare "." has no other meaning, and it could be read "point" anyway, as a convenient mnemonic. Happy to revise in favor of any other single character if desired. Another possibility that came to mind was a bare "@", chose "." mostly because it's easier to type and because of the cute mnemonic.

Note that as documentation is (currently?) separate from the source repository, I was sure whether you preferred a parallel PR to the docs repository, or just want me to do a PR there if this is merged. Happy to do either as you prefer, just let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [2.2.1-alpha] - unreleased
 
 ### Added
+- shortcut "." to specify the current point coordinates on command line
 
 ### Removed
 - importshp plugin, see issue #1481
--- a/librecad/src/lib/gui/rs_eventhandler.cpp
+++ b/librecad/src/lib/gui/rs_eventhandler.cpp
@@ -278,6 +278,15 @@ void RS_EventHandler::commandEvent(RS_CommandEvent* e) {
                     }
                 }
 
+                // handle quick shortcut for current point:
+                if (!e->isAccepted()) {
+                    if (cmd == ".") {
+                        RS_CoordinateEvent ce(relative_zero);
+                        currentActions.last()->coordinateEvent(&ce);
+                        e->accept();
+                    }
+                }
+
                 // handle absolute polar coordinate input:
                 if (!e->isAccepted()) {
                     if (cmd.contains('<') && cmd.at(0)!='@') {


lordofbikes:
I'm a bit in doubt about `.`, because it's meaning depends on language settings.
It's used with `,` in reversed sense in many languages, and therefore reflected in the num-pad layout.
Means, some layouts have `.`, when others have `,` keys on the num-pad.
If we add this for convenience, I think the num-pad should be considered and we might use an universal shortcut, which is available on all num-pads. I think `+` or `*` are universal, where `+` is a single keystroke on the alpha keyboard too, but `*` needs to hold `SHIFT` there.

Something what fits here is a frequent, one person request from IRC. He's not concerned about relative position, but the origin at `0,0`. So just 3 keystrokes but still too much.
So adding a shortcut for relative might be worth to add an origin shortcut too.
E.g. `+` for relative and `*` for origin, or a standalone `0` for origin?

lordofbikes:
Concerning the manual, we first have to tag a release now to separate 2.2.0 from master.
So I recommend to just open an issue there.
Adding this feature description in the release version will be just confusing.

gwhitney:
Hmm I don't have any keyboards with number pads. So that leaves me unsure if there's any clear win from the point of view of "simplest to type." So maybe we should go purely with mnemonic ease, in which case I'd suggest a bare `@` for the relative origin (since `@` is the usual flag for relative) and a bare `0` for the absolute origin (the "0-point" of the plane)?

In the end, just let me know what characters you select and I will be happy to update this PR.

GSStnb:
How about an asterisk { * ) as it is on the keyboard number row and number pad?

gwhitney:
'*' for which, the relative or absolute origin? I am gathering from @lordofbikes that the idea now would be for this to provide single-character abbreviations for both.

GSStnb:
' * ' in lieu of the '@' to avoiding having to 'leave" the the number pad for one character (meets the requirement for "simplest to type") .  '0' and  ' . ' is a coin toss, heads; '0', tails ' . '  :-)  

I lean to '0' for origin and ' * '.  A  period ( . ) is a overloaded as a decimal point and a delimiter.  Using an ' * ' forces a different key to avoid mis-typing (Maybe, perhaps, sorta?)

gwhitney:
Maybe a consensus is developing for `0` for the absolute origin? As for the relative origin, if it needs to be on the number pad, is there a reason to prefer `*` over @lordofbikes suggested `+`? Of the number pad keys I personally like `+` best, as LibreCAD currently indicates the relative origin on-screen with a red plus-in-circle icon.

GSStnb:
The red plus-in-circle icon looks like a splat if I take my reading glasses off.  

 ' + ' is also used for indicating a positive value (not that is really used as such since a positive value is assumed) where as ' * ' currently has no other purpose outside of a math operator.  From a documentation perspective, trying to explain the multiple uses of a single overloaded character is awkward.

gwhitney:
OK. At this point I am just standing by until @lordofbikes feels enough opinions have been expressed and provides guidance on how I should modify the PR. Thanks!

fa201:
I agree with @GSStnb (even if I did not understand the joke) knowing that some people have different locale settings: I personnally have `.` as a delimiter even if  the official is `,`  in France since programming with `,` is a pain in the neck with international and national data...

I vote for `*` for relative. I am sorry but `0` should mean zero.

Would it mean that double-dot notation is cancelled for relative ?
Could we have a radio button offering the legacy `@` and `*` as options? There may be people with scripts...

If there is a specific character for relative, why should we have another specific for absolute? I makes the code and the behavior of LibreCAD more difficult to understand and it is error-prone.

gwhitney:
Sorry, maybe the conversation became hard to follow. The concept of this PR is to *add* two single-character abbreviations specifically for the absolute and for the relative origin, when the command line is expecting a coordinate, since those appear to be very commonly-needed coordinates. It would not remove any other functionality, and the abbreviations should not conflict with any other valid coordinates.

Hence, `0` for the absolute origin is OK, because when the command line is expecting coordinates, just entering a `0` is currently not valid. And it seems to be an easy-to-remember shortcut for the point 0,0.

The debate seems mostly about what single character should abbreviate the "current point," i.e. the origin of the relative coordinate system. The leading candidates seem to be `+`, `*`, or `@` (in no particular order).

Nothing that is being proposed here would invalidate any current behavior of LibreCAD.

GSStnb:
# = 'hash'
! = 'bang'
* = 'splat'
...
The origin looks like (to me without my reading glasses) a " * ", or a 'splat'...

Perhaps to ensure everyone is on the same page, the PR introduces (and Glen - correct me if I am wrong):
   - A single character to replace having to type "0,0" for the origin.  Thus drawing a vertical line via the command line from the origin would be entered as li, 0 (first point - replacing the need to type 0,0) to 10..6 (second point)
    - Another single character to continue from the 'relative zero' (shown with the small red circle in the drawing space).  Thus, continuing from the above example to draw a circle at the end of the line, enter ci , * (centre - replacing the need to type 0..0), 5..0 (point on circle, radius of 5).

This example shows why the ' * ' is preferred for me - my fingers never left the number pad.  Having to type an '@' is moving back the the alpha-numeric keys and introducing a [Shift]

gwhitney:
Yes, thanks for the comprehensive examples. I agree. 

And I will throw one last possibility into the mix: `0` for absolute origin, and allow either `.` and `,` for current point (relative origin) -- in other words, both work as synonyms. That way, anyone can use the current point either on the number pad (since all number pads have either a `.` or `,`) and/or as an unshifted key from the regular keyboard (as opposed to all of `@#!+*` that are shifted keys on the regular keyboard) for those of us that have no number pads.

GSStnb:
Yes, that seems to be a the solution for those with number pads (extended keyboards) and for those without a number pads (laptops). '0'rigin and ' . ' | ' , ' - from the relative point.

lordofbikes:
Great, it seems we've come to an agreement.
I'm fine with that too. :+1:

gwhitney:
Oh I thought it was clear I needed to update the PR to conform to the conversation above. I just hadn't had a moment to do it yet. I definitely will this week. Should I make a new PR on top of this merged partial implementation? Or push another commit to this PR? (Sorry have never had a PR merged in the "middle" so to speak.) Thanks for guidance

lordofbikes:
As it is merged now, please do the implementation as discussed, based on new master, and create a new pull request.
Many thanks again! :+1:

PR #1677, opened 2023-08-08
feat: Extend one-character coordinate shortcuts

gwhitney:
This commit adds a shortcut of `,` to specify the current point
  (relative origin) when coordinates are expected on the command line,
  on top of the recently-added `.` shortcut with the same meaning. (This
  addition means that all number pads have a key which can be used for
  the current point.) In addition, it adds a shortcut of `0` to mean
  `0,0` so that there is a parallel one-character way of specifying
  the absolute origin.

(As agreed upon in the discussion re PR #1673.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## [2.2.1-alpha] - unreleased
 
 ### Added
-- shortcut "." to specify the current point coordinates on command line
+- shortcut "0" to specify coordinates "0,0" on command line
+- shortcut "." or "," to specify the current point coordinates on command line
 
 ### Removed
 - importshp plugin, see issue #1481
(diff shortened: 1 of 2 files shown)


librecad/src/lib/gui/rs_eventhandler.cpp
```diff
@@ -235,9 +235,28 @@ void RS_EventHandler::commandEvent(RS_CommandEvent* e) {
...
+                // handle quick shortcuts for absolute/current origins:
+                if (cmd.length() == 1) {
+                    RS_Vector at = relative_zero;
+                    switch (cmd[0].toLatin1()) {
+                        case '0':
+                            at.set(0,0);
+                            /* FALL THROUGH, to be replaced with c++17 [[fallthrough]] */
+                        case '.':
+                        case ',':
+                        {
```
dxli, line 247: Some minor coding style changes added.

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 ## [2.2.1-alpha] - unreleased
 
 ### Added
-- shortcut "." to specify the current point coordinates on command line
+- shortcut "0" to specify coordinates "0,0" on command line
```
dxli, line 6: wording

"the origin of the absolute space, _i.e._ the coordinates (0, 0)"

Issue #1207, opened 2020-03-18
Exiting print preview displays wrong drawing


xanderdin:
### Expected behavior
Exiting print preview mode should return to previewed drawing document.

### Observed behavior
Exiting print preview mode brings to another drawing document.

### Steps to reproduce or sample file
1. Run LibreCAD (default 'unnamed document 1' will be opened).
2. Open another document with some drawing.
3. Click Print Preview icon in toolbar to enter preview mode.
4. Click Print Preview icon again to exit preview mode.

Result: instead of returning to the opened in step 2 document the empty 'unnamed document 1' is displayed.

### Operating System and LibreCAD version info
Version: 2.2.0-rc1-190-g51b21cc7
Compiler: GNU GCC 8.3.0
Compiled on: Mar 18 2020
Qt Version: 5.11.3
Boost Version: 1.67.0
System: Debian GNU/Linux 10 (buster)

~~I've also build @baranovskiykonstantin's version from https://github.com/baranovskiykonstantin/LibreCAD/tree/closeprintpreview for comparison and it does not have this issue. Obviously the rejected https://github.com/LibreCAD/LibreCAD/commit/2bd7b074ecced9545fa99940ab20d13b44567472 code is somehow responsible for this.~~

~~Related to #1169~~

lordofbikes:
I've merged the rejected commit, but the issue remains for me.
Also, the same issue applies when a block is edited and closed.
- start a new session in tab mode, which opens an unnamed document
- open a drawing with blocks
- edit any block
- close the block editor tab
- the first tab, which is the unnamed document is activated

I assume there is no and never was any logic behind the tab handling. So when any tab is closed, the index for the active tab is reset to 0 which simply is the first tab.
This would need a kind of stack to implement, which could also be used for CTRL-Tab and CTRL-SHIFT-Tab handling.
Currently these hot keys iterate through the tabs ascending or descending. With some logic behind we can implement the known functionality from other software, where CTRL-(SHIFT)-Tab iterates the tabs in historical order. This allows to fast toggle between two drawings over and over again.

xanderdin:
I confirm, the issue is still there.

Strange, but it looks like I've mistaken when was comparing builds and built not from baranovskiykonstantin:closeprintpreview  but rather from his master branch which was not having the issue. :man_facepalming:

tonischranz:
[...]

PR #1696, opened 2023-10-05
Exiting print preview displays wrong drawing #1207

israelmsnts:
Hello,

This is my first PR, so I picked up a simple issue to start with. As described in Issue #1207, after performing certain actions, LibreCAD always reverts to the first window. I tested this on the 2.2 branch, and the issue persists.

Upon inspecting the code, I noticed that even though there is access to parentWindow, the active window always defaults to the first one in the window list. Since parentWindow is used to determine whether the current window is printPreview or a Block, I utilized this variable to set the focus when closing the current window.

I use LibreCAD for woodworking drawings as a hobby, and while I may not have experience with C++ or QT, I was eager to contribute.

### DIFF
--- a/librecad/src/main/qc_applicationwindow.cpp
+++ b/librecad/src/main/qc_applicationwindow.cpp
@@ -547,8 +547,15 @@ void QC_ApplicationWindow::doClose(QC_MDIWindow * w, bool activateNext)
 	actionHandler->set_view(nullptr);
 	actionHandler->set_document(nullptr);
 
-	if (activateNext && window_list.count() > 0)
+    if (!parentWindow && activateNext && window_list.count() > 0)
+    {
 		doActivate(window_list.front());
+    }
+    else
+    {
+        RS_DEBUG->print("QC_ApplicationWindow::active parent window");
+        doActivate(parentWindow);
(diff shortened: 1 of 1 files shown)


lordofbikes (changes requested):
The static analyzer will throw a warning here, because of the theoretical chance, that the `if()` condition fails while `parentWindow` is `nullptr`. Then `doActivate()` will be called in the `else` branch with `nullptr` parameter.
So I think it should be:
```
    if (activateNext && window_list.count() > 0) {
        if (nullptr != parentWindow) { 
            doActivate(parentWindow);
        }
        else {
	    doActivate(window_list.front());
        }
    }
```

lordofbikes:
Many thanks for your efforts!
If you plan more contributions, please use the `master` branch for your changes and pull requests.
When the changes are safe and worth a back port to the stable `2.2` branch, we can cherry-pick the commits from `master` to stable branch.
For this fix, I have tested it, we can merge it to `2.2` as an exception.

lordofbikes (approved):
Many thanks! :+1:

lordofbikes:
Merged to `master` too.

PR #1702, opened 2023-10-13
Icons

flywire:
Three icon issues are:
1. some icons contain #130000 (Diesel) color which is generally perceived as #000000 (Black)
2. the zoom in/out icons have long been reversed
3. icons were created with 90 DPI rather than the css standard of 96 DPI

This change fixes the first two issues by changing Diesel to Black, and swapping the contents of the zoom icons. The third issue has never been considered and is discussed in https://inkscape.org/release/inkscape-0.92/:

> Release Notes
Inkscape 0.92
-=  Released 2017-01-01  =-
>
> • The default resolution was changed from 90dpi to 96dpi, to match the
CSS standard. For more background information, please see the Wiki
article about handling of units in Inkscape. Inkscape 0.92 will
attempt to identify 'legacy' Inkscape files that need to be
converted. If such a file is detected, the user will be offered
three options:
    1. Set 'viewBox'. Inkscape will add an appropriate 'viewBox' which
will do a global scaling of the document. It will also adjust
the document width and height if necessary.
    2. Scale elements. Inkscape will scale each internal element.
    3. Ignore. Do nothing. This is an appropriate choice for documents
meant for screen display.

I suggest trying option 2. Scale elements. I don't expect any significant difference to svg files.

Can anyone provide comments in terms of the programming and OS environments?

### DIFF
--- a/librecad/res/icons/angle_line_to_line.svg
+++ b/librecad/res/icons/angle_line_to_line.svg
@@ -144,6 +144,6 @@
        sodipodi:cx="60"
        sodipodi:type="arc"
        id="path5091"
-       style="fill:none;fill-opacity:1;fill-rule:evenodd;stroke:#130000;stroke-width:7.99999857;stroke-linecap:square;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:8;stroke-opacity:1" />
+       style="fill:none;fill-opacity:1;fill-rule:evenodd;stroke:#000000;stroke-width:7.99999857;stroke-linecap:square;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:8;stroke-opacity:1" />
   </g>
 </svg>
--- a/librecad/res/icons/arc_concentric.svg
+++ b/librecad/res/icons/arc_concentric.svg
@@ -72,7 +72,7 @@
        sodipodi:cx="9.0311108"
        sodipodi:type="arc"
        id="path4394"
-       style="fill:none;fill-opacity:1;fill-rule:evenodd;stroke:#130000;stroke-width:2.25777793;stroke-linecap:butt;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1" />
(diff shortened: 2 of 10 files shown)


lordofbikes:
Many thanks :+1:

1. fine
2. is probably an issue of point of view, similar to mouse wheel scrolling direction
3. is something Inkscape complains about for years and is basically correct, but this matters mostly, when it comes to printing, because the icons are created in pixel size in GUI, the DPI  
setting is not relevant, it is no problem to change it, to silence Inkscape, but it doesn't affect LibreCAD

lordofbikes:
many thanks :+1: 

1. fine
2. is probably an issue with point of view, similar to scroll direction with mouse wheel
  zoom in means the section in the magnifier becomes smaller, thus the document expands and vice versa for zoom out
  from this point of view the icons are correct and probably the intention from the creator
  so I think they are correct, and I'm probably not alone, I have found no issue or forum post about this
3. Inkscape is complaining about this for years and it is basically correct
  but in GUI, the icons are created in pixel size, thus DPI doesn't matter, DPI is mainly an issue when it comes to printing
  so changing DPI to silence Inkscape is fine, it should not affect LibreCAD

A good example, why it is recommend to not mix up issues in pull requests, even when they are small. 1. could be merged, but 2. blocks it now. 
Every single issue, no matter how small, deserves its own branch and pull request :wink:

flywire:
> 2. is probably an issue with point of view, similar to scroll direction with mouse wheel
>    zoom in means the section in the magnifier becomes smaller, thus the document expands and vice versa for zoom out
>    from this point of view the icons are correct and probably the intention from the creator
>    so I think they are correct, and I'm probably not alone, I have found no issue or forum post about this
> 3. Inkscape is complaining about this for years and it is basically correct
>    but in GUI, the icons are created in pixel size, thus DPI doesn't matter, DPI is mainly an issue when it comes to printing
>    so changing DPI to silence Inkscape is fine, it should not affect LibreCAD
> 
> A good example, why it is recommend to not mix up issues in pull requests, even when they are small. 1. could be merged, but 2. blocks it now. Every single issue, no matter how small, deserves its own branch and pull request 😉

lordofbikes:
many thanks :+1:

PR #1708, opened 2023-11-05
Adding support for C++17 in project files.

marko-radojcic:
Small change in project files to enable C++17 compiling on GCC on Linux Mint Victoria. Compiles and runs. Needs testing on other platforms.

### DIFF
--- a/common.pri
+++ b/common.pri
@@ -27,7 +27,7 @@ win32 {
 
 # use c++ only
 QMAKE_CC = g++
-QMAKE_CFLAGS = -std=c++1y
+QMAKE_CFLAGS = -std=c++17
 
 # Windows compiler settings
 win32 {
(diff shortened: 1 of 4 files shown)


dxli:
I think we can update Qt dependency in README.md from Qt-5.2 to Qt-5.15 (LTS).

I'm pretty sure Qt-5.2 doesn't support c++17.

marko-radojcic:
I did test on 5.15. The reason I didn't push my changes that utilized Qt6 is that I had to undo them because, the task of modifying is getting bigger and bigger as you progress. Qt6 removed or deprecated many Classes and/or methods that are used in LibreCAD. I read about the solutions but they require a more significant rewrite of the code which I can't do at the moment.

dongxuli-concord-design:
We can create a qt6 branch. Better to do it now than later

marko-radojcic:
I agree with your solution, it is probably for the best.

dxli:
Hi Marko,

I have created a qt6 branch.

All I can say is that it builds with qt6, but obviously, need to first verify all changes.

Then, we need to get readme/howto docs updated, and get CI builds ready as well.

If interested, you may help from here.

Regards,

Dongxu

PR #1712, opened 2023-11-19
Fix loading of splines that falsely defined by closed

### LINKED ISSUE #863: Closed spline problem
It seems that when a spline has the "closed" flag the knot vector is not respected resulting in a wrong result. Here is a simple circle converted to spline.
https://www.dropbox.com/s/hbewagpqk94po8m/circle.dxf?dl=0

When opened in the latest librecad it results in a rounded square rather a circle.

circle

If you open the properties and untick the "**Closed**" spline then it is rendered correctly as a circle.

Looking at the code it seems that in the **rs_spline.cpp**, method **RS_Spline:update()** when the spline is flagged as **closed** it appends to the end of the **tControlPoints** the first n-degree points. However it doesn't adjust accordingly the **data.knotlist**, as a result the data.knotslist in the "RS_Spline::knotu" doesn't have a length of "num + order", so instead of using the dxf knotlist it creates a new **knotVector**.

baranovskiykonstantin:
Fixes #863.

[Closed splines](https://pages.mtu.edu/~shene/COURSES/cs3621/NOTES/spline/B-spline/bspline-curve-closed.html) have the `p` identical control points in the beginning and at the end of the control points list (where `p` is the degree of the spline).
Some DXF-files contains splines that are defined as closed but actually are open (an example is attached) that causes the incorrect rendering by the LibreCAD. We should verify the closure of the spline by checking the conrol points.

screenshot

[open-circle-spline-defined-as-closed.dxf.zip](https://github.com/LibreCAD/LibreCAD/files/13403914/open-circle-spline-defined-as-closed.dxf.zip)

### DIFF
--- a/librecad/src/lib/filters/rs_filterdxfrw.cpp
+++ b/librecad/src/lib/filters/rs_filterdxfrw.cpp
@@ -666,6 +666,9 @@ void RS_FilterDXFRW::addSpline(const DRW_Spline* data) {
         for (auto const& vert: data->fitlist)
             spline->addControlPoint({vert->x, vert->y});
     }
+    // ensure that the spline is really closed
+    if (spline->data.closed and !spline->hasWrappedControlPoints())
+        spline->data.closed = 0;
 
     spline->update();
 }


dongxuli-concord-design:
What's the expectation based on the dxf specifications?

baranovskiykonstantin:
> What's the expectation based on the dxf specifications?

I did not found the explanation in the specification but I checked the attached file in the AUTODESK online viewer and it draws the circle:
https://autode.sk/3R6jjse

PR #1714, opened 2023-11-21
Support snapping for leaders

### LINKED ISSUE #1686: Leader: line below text should be oriented with 15° increment as per regular line
### Expected behavior
I would like to have the line below text be oriented with SHIFT key to get 15 degree step on line angle as per a regular line.
Actually it would make it also easier to draw perfectly horizontal line.

I twa also requested in the forum: https://forum.librecad.org/Leader-Line-Angle-Constraints-td5721146.html

Earlopain:
* Moved snap angle to the header file. Everywhere it was 15 degrees
* Add snapping to leaders (:

Closes #1686

### DIFF
--- a/librecad/src/actions/rs_actiondefault.cpp
+++ b/librecad/src/actions/rs_actiondefault.cpp
@@ -58,9 +58,6 @@ struct RS_ActionDefault::Points {
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


Earlopain:
Thank you!

PR #1730, opened 2024-01-26
Export 2 csv plugin

quimperval:
This plugin helps to export the coordinates of points, lines and polylines  to csv files (but separated by semicolon).


plugins/pointstocsv/pointstocsv.cpp
```diff
@@ -0,0 +1,282 @@
...
+#include <QTextEdit>
+#include <QDialogButtonBox>
+#include <QVBoxLayout>
+#include <QPushButton>
+#include <QComboBox>
+#include <QLabel>
+#include <QRect>
+#include <QFileDialog>
+#include <QMessageBox>
+#include <cmath>
```
dxli, line 25: On header order:

we mostly keep c++ standard headers before Qt headers. Once there were some Qt bugs which were triggered if Qt headers are included before c++ standard headers, but in most time, the order is not an issue.

quimperval:
Hello @dxli 
I've made the changes according to your comments. 
Please let me know any other comment.

dxli:
Hi @quimperval ,

It's good enough for merging.

One thing might be worth doing: create a unified interface for entity data exporting/importing. Something like:

```C++
struct EntityDataInterface
{
    virtual QString ExportCsv() const = 0;
    virtual void ImportCsv(const QString&) = 0;
};

struct LineData : public EntityDataInterface
{
....
    QString ExportCsv() const override {
        return getLineFormatedText(*this);
    }
};
```
Regards,

dxli

PR #1744, opened 2024-02-28
Snap Middle by Melwyn PR #1424

dxli:
This PR is to merge PR #1424 by Melwyn Francis Carlo <melwyncarlo@gmail.com>

To test this new "Snap middle" feature:

1, click to activate this tool;
2, select point 1 and 2;

The middle point of 1 and 2 is snapped

### DIFF
--- a/librecad/res/icons/icons.qrc
+++ b/librecad/res/icons/icons.qrc
@@ -51,6 +51,7 @@
         <file>snap_grid.svg</file>
         <file>snap_intersection.svg</file>
         <file>snap_middle.svg</file>
+        <file>snap_middle_manual.svg</file>
         <file>lock_rel_zero.svg</file>
         <file>set_rel_zero.svg</file>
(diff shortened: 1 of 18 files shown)


dellus:
Wonderful.
I have tried to express the percent value, for instance 25 %, as 100/4. But it is not accepted, it takes the middle instead. Could this be achieved? Maybe better even to replace percentage by fraction from the onset. Middle then would be 1/2, but not necessary to type as it is the default.

What about the virtual line intersection in the same manner?

dxli:
it should work by design, to specify 25% by typing in command widget:

100/4
or
12+13

dellus:
Yes, it does work. I must have been confused about the syntax, probably not closing the input in command widget with Enter.
Then I struggled with how to specify anything larger than 1/2 (50%), for instance 2/3. Now I have got it, it's 100*2/3.
All is fine as it is.

PR #1745, opened 2024-03-04
Pens palette widget

sand1024:
a small yet nice widget for a bit more convenient way for working with pens.

It allows managing of persistent pens, select entities based by pens and apply pens to selection, current layer and so on.

Demo video for the widget is there - https://youtu.be/YSU0K3-PfbI
Feature branch - https://github.com/sand1024/LibreCAD/tree/pens_palette_widget
Readme - https://github.com/sand1024/LibreCAD/blob/pens_palette_widget/PEN_PALETTE_README.md


dxli:
On copyright and licence: for new additions, we tend to use GPLv2 (or later versions); Only for legacy code, we keep the old GPLv2.

Could you also assign copyright to LibreCAD.org, so giving the possibility of relicensing to other free software licenses. For example we can share code with other open source CAD, or simply join LibreCAD v3.

Regards,

dxli

sand1024:
Yep, no problem - I suppose I've picked a license from one of existing files and probably one was with the old license. 
Could you please point me some file with right license? I can update files accordingly - or you can do this during the merge, if it's simpler

dongxuli-concord-design:
Lc_ prifix indicates our own addictions, so free to choose GPL v2 or later.

New methods in old files could be indicated by an "author" comment.

In past, we got some complaints saying we didn't respect copyright of the original qcad, so we would like to keep some records in the git codebase. It's a good thing in my eyes, as free software developers, we would like to honor contributors by keeping their names in codebase.

BTW, I think I ran into technical difficulties in my hyperbola approach. I think I should have a little bit more time to help getting your work merged faster.

Regards,

dxli

PR #1761, opened 2024-04-23
Additional actions

sand1024:
This pull request includes several handy actions for entities editing and modification (see below).

Actions are:
1) Draw Cross (https://youtu.be/H9Lul0J02a0)
2) Draw Star (https://youtu.be/bAD7kkAR5wA)
3) Draw Rectangle - 1 Point (https://youtu.be/eOscpTiWJMY)
4) Draw Rectangle - 2 Points (https://youtu.be/9bjBZAcEmVc)
4) Draw Rectangle - 3 Points (https://youtu.be/IJhG3Rz9U5o)
5) Draw Line of Points (https://youtu.be/8kDSSKg-vqU)
6) Draw Snake Line (https://youtu.be/bXom0IWw4R4)
7) Draw Snake (X) Line (https://youtu.be/bXom0IWw4R4)
8) Draw Snake (X) Line (https://youtu.be/bXom0IWw4R4)
9) Draw Angle From Line (https://youtu.be/ZcnIUXZdSBI)
10) Draw Orthogonal From Line (https://youtu.be/sfO2aUeQ_2E)
11) Draw Line from Point To Line (https://youtu.be/3bEWIryP6jQ)
12) Slice/Divide Line (https://youtu.be/z0I8N73aWFM)
13) Slice/Divide Circle/Arc (https://youtu.be/ILACvThv6I4)
14) Draw Circle by Arc (https://youtu.be/xSJnC-Mvczw)
15) Duplicate entity (https://youtu.be/xSJnC-Mvczw)
16) Lines Join (https://youtu.be/NCqodd0i_gE)
17) Break/Divide (line, arc, circle) https://youtu.be/_UHawUJXZe0
18) Line Gap (https://youtu.be/0Y0YsNSZIJA)

+ Generic basic class for action and action option widgets
+ customizable column count for left toolbar
+ settings based ability to re-open last opened files

More detailed description for actions and features - located here https://github.com/sand1024/LibreCAD/blob/relative_point/ACTIONS_README.md


sand1024:
Hi! 
I've noticed your changes, yet not sure that this change is correct: 
https://github.com/LibreCAD/LibreCAD/commit/c2268b55174e19240147ceb7a41a71c20a0c083c

more specifically - this correction: 

RS_Vector LC_LineMath::getNearestPointOnLine(RS_Line* line, const RS_Vector& coord, bool infiniteLine){
    if (infiniteLine){

        RS_Vector point1 = line->getStartpoint();
        RS_Vector point2 = line->getEndpoint();
        return getNearestPointOnInfiniteLine(coord, point1, point2);
    }
    else{
        return line->getNearestPointOnEntity(coord, true, nullptr);
    }
    // For infinite lines, find the nearest point is not limited by start/end points
    bool onEntity = ! infiniteLine;
    return line->getNearestPointOnEntity(coord, onEntity, nullptr);
}

Probably I'm incorrect, but line->getNearestPointOnEntity(coord, onEntity, nullptr);  will return one of endpoints if of the line 
if the line is not constructional (please refer to RS_Line: 153-159). 

However, the function getNearestPointOnInfiniteLine actually should return projection of the point on ray on which given line is placed (so basically, that may be similar to considering given line to the construction line and checking projection on construction line). 

Thus I suppose such change may affect the logic of the function - please check this.

I'll check this too, but from the first glance it seems that such a change may affect original logic. 

Thank you

librecad/src/lib/math/lc_linemath.cpp
```diff
@@ -355,28 +339,9 @@ RS_LineData LC_LineMath::createParallel(RS_Vector &start, RS_Vector &end, double
...
-        double ys = s1.y + u * (e1.y - s1.y);
-        ret = RS_Vector(xs, ys);
-    }
-    else {
-        // lines are parallel
-        ret = RS_Vector(false);
-    }
-
-    return ret;
+RS_Vector LC_LineMath::getIntersectionLineLine(const RS_Vector& s1, const RS_Vector& e1, const RS_Vector& s2, const RS_Vector& e2) {
```
sand1024, line 340: yes, you're correct regarding the code duplication there. Yet  the reason why there is separate implementation is just a minor performance optimization - the original implementation does not create additional temporary objects and there might be only one intersection point (so one solution). So it just may be slightly faster. 
For sure, these implementations are equivalent, but I think it might be practical to add that specific case for lines intersections based on points to RS_Information class?

librecad/src/lib/math/lc_linemath.cpp
```diff
@@ -150,15 +152,9 @@ RS_Vector LC_LineMath::calculateEndpointForAngleDirection(double angleValueDegre
...
-        RS_Vector point1 = line->getStartpoint();
-        RS_Vector point2 = line->getEndpoint();
-        return getNearestPointOnInfiniteLine(coord, point1, point2);
-    }
-    else{
-        return line->getNearestPointOnEntity(coord, true, nullptr);
-    }
+    // For infinite lines, find the nearest point is not limited by start/end points
+    bool onEntity = ! infiniteLine;
+    return line->getNearestPointOnEntity(coord, onEntity, nullptr);
```
sand1024: please take a look to my comment to pen request - I'm not sure that  getNearestPointOnEntity will not return edge endpoints as line is not constructional. Yet buy generic logic of the function, if infiniteLine parameter is true, projection to the ray where the line is located (or projection on construction line) is needed.

dongxuli-concord-design:
Before my change, your implementation was using onEntity=true, for non-infinite lines, and the infinite line implementation is equivalent to onEntity=false. 
That's the reason of the change. Logically, equivalent

dongxuli-concord-design:
Before my change, your implementation was using onEntity=true, for non-infinite lines, and the infinite line implementation is equivalent to onEntity=false. 
That's the reason of the change. Logically, equivalent 

Also, conceptually, there are only onEntity=true/false two cases. If you had two cases initially, the implementation should be equivalent.

Issue #1759, opened 2024-04-17
'Properties' doesn't show 'By layer' for an entity if so assigned but the end result


dellus:
Expected behavior: 'Properties' for an entity drawn with 'by layer' should show 'By layer'.

Observed behavior: 'Properties' for an entity drawn with 'by layer' shows the attributes assigned to the layer.

LibreCAD 2.2.1 alpha 412 appimage or earlier (since introduction of Layer Tree and Pen Palette).
Puppylinux Fossapup64

sand1024:
mmm - not sure I understand. In which place so you observe such a behavior? In properties dialog? In active pen buttons? 
Please add more details.

dongxuli-concord-design:
By layer
By block

I do remember there's something logic cleanup needed.

sand1024:
mmm - I think I finally got it - if properties dialog is invoked for the entity, "By layer/By Block" label is not shown anymore, only actual color name. 
Yes, it seems it's my fault related to setup of properties dialogs, I'll fix it. Actually, the intent was to show not a black/white icon for the color in dialogs, but actual color (that same logic as in active pen toolbar).  
Anyway, I'll correct this with separate pull request.

dongxuli-concord-design:
We can keep the color icon and the text "by Layer" for example. Hope it works out for all languages

sand1024:
yep, that was exactly my original intent - yet there is a bug in implementation I didn't notice.

dxli:
I disabled the PenPallet by default.

You can remove the setting, or enable it by:

```
[CustomToolbars]
UsePenPallet=1
```

dxli:
Tools > Modify > Attributes, color is shown as "Unchanged";

Tools > Modify > Properties, color is shown as the current value.

Two behavior is good (for different usages). Closing this one.

PR #1763, opened 2024-04-23
Pen toolbar - sync pen by layer action now works if PenPalette widget is disabled

sand1024:
Original implementation relied on PenPalette widget for applying active layer's pen to pen toolbar, so if PenPalette is disabled - that action didn't work. 
Now separate action was added, so functionality is restored and works if PenPalette is disabled.


dongxuli-concord-design:
Hi @sand1024

Did you do something to reduce the sizes?

sand1024:
Hi!
not yet, I'm going to start adjusting sizes now and hope I'll commit them in nearest couple of days. 

That commit is for a bit different situation - previously, the action for updating pen by layer's pen used invocation of method from PenPalette - and so that was the reason why the action (2 green triangles) was not executed if PenPalette is disabled. 

Now I've added separate action, so PenToolBar and PenPalette are independent.

sand1024:
I think I'll fix "By Layer" issue in properties first (#1759) as it is faster, and after it will switch to sizes adjustments.

sand1024:
committed fix for #1759

PR #1785, opened 2024-05-14
fix for the issue with drawing snapper and coordinates in coordinates…

sand1024:
a small fix for recently added actions

In my initial implementation, coordinates widget was not updated and snapper (if enabled) was not shown for cases where preview is not drawn (i.e - for initial states of action). 

Now these actions works in more consistent way with other actions. 

Fix is applicable both for master branch and 2.2.1

### DIFF
--- a/librecad/src/actions/lc_abstractactionwithpreview.cpp
+++ b/librecad/src/actions/lc_abstractactionwithpreview.cpp
@@ -20,21 +20,24 @@ along with this program; if not, write to the Free Software
 Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 **********************************************************************/
 
+#include <QList>
+#include <QMouseEvent>
+
+#include "lc_abstractactionwithpreview.h"
+#include "lc_actiondrawslicedivide.h"
+#include "rs_arc.h"
+#include "rs_circle.h"
+#include "rs_commandevent.h"
+#include "rs_commands.h"
+#include "rs_coordinateevent.h"
+#include "rs_debug.h"
+#include "rs_dialogfactory.h"
+#include "rs_graphicview.h"
+#include "rs_line.h"
 #include "lc_linemath.h"
 #include "rs_point.h"
-#include "QMouseEvent"
-#include <QList>
-#include "rs_circle.h"
-#include "rs_arc.h"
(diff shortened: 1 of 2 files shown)


librecad/src/actions/lc_abstractactionwithpreview.cpp
```diff
@@ -149,6 +149,23 @@ bool LC_AbstractActionWithPreview::isAcceptSelectedEntityToTriggerOnInit([[maybe
...
+    RS_Vector mouse = snapPoint(e);
+}
+
+/**
+ * Explicitly updates coordinate widget by current mouse position.
+ * Method is useful for actions states that do not call snapPoint() method on mouse move (which, in turn, updates the widget)
+ * @param e
+ */
+void LC_AbstractActionWithPreview::doUpdateCoordinateWidgetByMouse(QMouseEvent* e){
+    RS_Vector mouse = graphicView->toGraph(e->x(), e->y());
```
dongxuli-concord-design, line 163: Please use e->position ()

X, y methods are deprecated

dongxuli-concord-design (approved):

librecad/src/actions/lc_abstractactionwithpreview.cpp
dxli: One more thing, we should try to minimize the header dependency of base classes.

Decoupling headers is good in long run.

To remove the "rs_vector" dependency, we can keep pointer members in the base class and use accessors, like:
```
class RS_Vector;
class Base {

public:
const RS_Vector& getPoint() const;
RS_Vector& getPoint();
void setPoint(const RS_Vector&);

private:
std::unique_ptr<RS_Vector> m_aPoint;
};

// in cpp

const RS_Vector& Base::getPoint() const 
{
     return m_aPoint != nullptr ? *m_aPoint: RS_Vector{};
}

RS_Vector& Base::getPoint()
{
    assert(m_aPoint != nullptr);
    if (m_aPoint!=nullptr)
        return *m_aPoint;
    static RS_Vector shouldnotUse{};
    return shouldnotUse;
}

void Base::setPoint(const RS_Vector& newValue)
{
    if (m_aPoint == nullptr)
        m_aPoint = std::make_unique<RS_Vector>();
    *m_aPoint = newValue;
}
```

PR #1798, opened 2024-06-02
fix compilation with clang and llvm's libcxx

dpronin:
librecad/src/lib/debug/rs_debug.h requires cstdio to be included because of FILE type being used

log of compilation failure:

```bash
clang++ -c -pipe -g -O2 -std=gnu++11 -Wall -Wextra -D_REENTRANT -fPIC -DDWGSUPPORT -DQC_APPDIR="librecad" -DLC_VERSION="2.2.0.2-5-g3aa2c41c" -DQT_NO_DEBUG -DQT_SVG_LIB -DQT_PRINTSUPPORT_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -I. -I/usr/include -I../../libraries/libdxfrw/src -I../../libraries/jwwlib/src -Icmd -Ilib/actions -Ilib/creation -Ilib/debug -Ilib/engine -Ilib/fileio -Ilib/filters -Ilib/generators -Ilib/gui -Ilib/information -Ilib/math -Ilib/modification -Ilib/printing -Iactions -Imain -Imain/console_dxf2pdf -Itest -Iplugins -Iui -Iui/forms -Iui/generic -I../res -I/usr/include/qt5 -I/usr/include/qt5/QtSvg -I/usr/include/qt5/QtPrintSupport -I/usr/include/qt5/QtWidgets -I/usr/include/qt5/QtGui -I/usr/include/qt5/QtCore -I../../generated/librecad/moc -I../../generated/librecad/ui -I/usr/lib64/qt5/mkspecs/linux-clang -o ../../generated/librecad/obj/rs_arc.o lib/engine/rs_arc.cpp
In file included from lib/debug/rs_debug.cpp:28:
lib/debug/rs_debug.h:91:20: error: unknown type name 'FILE'
   91 |     void setStream(FILE* s) {
      |                    ^
lib/debug/rs_debug.h:99:5: error: unknown type name 'FILE'
   99 |     FILE* stream;
```

### DIFF
--- a/librecad/src/lib/debug/rs_debug.h
+++ b/librecad/src/lib/debug/rs_debug.h
@@ -28,6 +28,7 @@
 #ifndef RS_DEBUG_H
 #define RS_DEBUG_H
 
+#include <cstdio>
 #include <iosfwd>
 #ifdef __hpux
 #include <sys/_size_t.h>
--- a/librecad/src/lib/debug/rs_debug.h
+++ b/librecad/src/lib/debug/rs_debug.h
@@ -29,7 +29,6 @@
 #define RS_DEBUG_H
 
 #include <cstdio>
-#include <iosfwd>
 #ifdef __hpux
 #include <sys/_size_t.h>
 #endif


dpronin:
The patch probably has to be propagated to the versions utilizing qt5, at least all ~2.2, i.e. >=2.2.0 and <2.3

dxli:
We should replace the iosfwd header with the actual cstdio

dxli (approved):

dpronin:
> The patch probably has to be propagated to the versions utilizing qt5, at least all ~2.2, i.e. >=2.2.0 and <2.3

I have checked compilation of 2.2.1 and it completes without any changes, so we could leave these changes only for 2.2 version

dpronin:
@dxli who is supposed to merge the changes in?

dongxuli-concord-design:
It's still blocked. 

I will see whether I can manually push. 

Btw, looks like we don't need this for 2.2.1 and and master(2.2.2), where we already include qt headers.

dpronin:
> It's still blocked. 
> 
> I will see whether I can manually push. 
> 
> Btw, looks like we don't need this for 2.2.1 and and master(2.2.2), where we already include qt headers. 

No, we don't. 2.2.1 and 2.2.2 compile successfully since there are qt headers included how you have told

Nevertheless, I prefer to follow the rule **include what you use**

dongxuli-concord-design:
Yes, you are right on including all used. 


I am wondering whether we can rely on pimp pattern here. It's a header essentially used by all, so better to be lean. 

I mean to remove studio along with qt headers by using a pImp pointer.

dpronin:
@dongxuli-concord-design what you did is something weird, it is not pimpl idiom. Moreover, you have redesigned the class by removing a pointer to stream from the fields and making this stream be a static field within the compilation module

take a look at [the answer](https://stackoverflow.com/questions/71104545/constructor-and-destructor-in-c-when-using-the-pimpl-idiom/71104867#71104867) how to implement pimpl correctly

dongxuli-concord-design:
> @dongxuli-concord-design what you did is something weird, it is not pimpl idiom. Moreover, you have redesigned the class by removing a pointer to a stream from the fields and making this stream be a static field within the compilation module
> 
> take a look at [the answer](https://stackoverflow.com/questions/71104545/constructor-and-destructor-in-c-when-using-the-pimpl-idiom/71104867#71104867) how to implement pimpl correctly


The goal is to make the header lean, not necessarily pImp. I didn't use pImp, so including <memory> or using raw pointers are avoided. 

It's a Singleton, so compiling unit wide static fields just work as well as members. 

The inner interface class was initially designed to be an interface, but removing smart pointers made the class a concrete one, so its name still needs to be corrected.

dpronin:
> mean to remove studio along with qt headers by using a pImp pointer.

> ...

> The goal is to make the header lean, not necessarily pImp

Please, decide what you like to do, you confuse clever people



> It's a Singleton, so compiling unit wide static fields just work as well as members.

I am very happy that you probably recently learnt 2-3 design patterns but you implement them as messy and clumsy as it gets. If you implement class as singleton then do not leave one half of fields in the class declaration and the second in the module, you violate everything by this move. Either you use a C++-oriented singleton and declare everything within the class or you make a C-oriented singleton and declare and define everything with the compilation module, decide

dongxuli-concord-design:
> > I mean to remove studio along with qt headers by using a pImp pointer.
> 
> > ...
> 
> > The goal is to make the header lean, not necessarily pImp
> 
> Please, decide what you like to do, you confuse clever people
> 
> 
> 
> > It's a Singleton, so compiling unit wide static fields just work as well as members.
> 
> I am very happy that you probably recently learnt 2-3 design patterns but you implement them as messy and clumsy as it gets. If you implement class as singleton then do not leave one half of fields in the class declaration and the second in the module, you violate everything by this move. Either you use a C++-oriented singleton and declare everything within the class or you make a C-oriented singleton and declare and define everything with the compilation module, decide

If you can improve it or redesign it, you are more than welcome to do so.

PR #1828, opened 2024-08-16
"Warning:: Paper size less than 10mm" warning text updated

iqqmuT:
**Current Drawing Preferences** is under **Options** menu, not **Edit**.


lordofbikes:
many thanks :+1:

PR #1834, opened 2024-08-21
Two fixes of unicode.lff from SolveSpace

ruevs:
，
	(U+FF0C) Fullwidth Comma - Was invalid, fixed
 
�
	(U+FFFD) REPLACEMENT CHARACTER - Changed to a nicer looking one

https://github.com/solvespace/solvespace/blob/master/res/fonts/unicode.lff.gz

Fro cross reference:
https://github.com/solvespace/solvespace/issues/1042 https://github.com/solvespace/solvespace/pull/1478

Issue #1835, opened 2024-08-21
LibreCAD ignores font defined U+FFFD symbol shape


app4soft:
### Expected behavior

LibreCAD should use shapes of all symbols set defined in LFF.

If font has custom shape for `[fffd] �` (U+FFFD) symbol LibreCAD should use it.

### Observed behavior

Actually LibreCAD uses hardcoded shape for `[fffd] �` for all fonts, ignoring font defined shape.

Discovered after https://github.com/LibreCAD/LibreCAD/pull/1834 has been merged.

### Steps to reproduce or sample file

Here is a testcase LFF-font sample (copy code and save as LFF file, then open in LibreCAD).

There is `[fffd] �` symbol copied shape from `[003f] ?` symbol, and then itself copied to `[00bf] ¿` symbol.

```
# Format:            LibreCAD Font 1
# Creator:           LibreCAD
# Version:           2.2.2_alpha1-138-g515194f8
# Name:              LFF
# Encoding:          UTF-8
# LetterSpacing:     3
# WordSpacing:       6.75
# LineSpacingFactor: 1
# Created:           
# Last modified:     2024-08-21
# License:           Public Domain

[fffd] �
C003f

[003f] ?
3,0;0,3
0,3;0,6
0,6;3,9
3,9;6,6
6,6;6,3
6,3;3,0
4.2,5.4;1.8,5.4,A1
4.2,5.4;3.96822,4.75982,A-0.17545
3.3,4.1;2.99957,3.40001,A0.19789
3.97,4.76;3.3,4.1
3,2.8;3,2.2,A1
3,2.2;3,2.8,A1

[00bf] ¿
Cfffd
```

And here is how it looks in LibreCAD

![Screenshot_2024-08-22_01-37-20](https://github.com/user-attachments/assets/f0142edc-1326-4ff7-90e8-e3a81005f5fe)

### Operating System and LibreCAD version info

- Version: 2.2.2_alpha1-138- g515194f8
- Compiler: GNU GCC 9.4.0
- Compiled on: Aug 21 2024
- Qt Version: 6.6.3
- Boost Version: 1.71.0
- System: Debian GNU/Linux 12 (bookworm)

dxli:
fffd is the "replacement character"

It's intended to be an unknown character for the font. Should a font define its own unknown? Or the software should assign the unknown character if the font doesn't know a character?

Let a font know about the unknown is the issue. How do I know whether the shape is unknown to the font?

[...]

PR #1839, opened 2024-08-22
Add `�` to the U+FFFD (REPLACEMENT CHARACTER) in unicode.lff

ruevs:
To address the issue raised here:
https://github.com/LibreCAD/LibreCAD/pull/1834#issuecomment-2302869925


dxli:
Let's merge this one, but please note the unknown character is supposed to be handled by logic (not following the font specification), so the shape may not be acutally used

ruevs:
> but please note the unknown character is supposed to be handled by logic (not following the font specification), so the shape may not be acutally used

[Understood.](https://github.com/LibreCAD/LibreCAD/issues/1835#issuecomment-2303855086)

Issue #1846, opened 2024-08-23
Performance regression in text rendering


u2fly:
### Expected behavior

There should not be performance regressions after new features implemented.

### Observed behavior

Actually there is a huge performance regression in text rendering, affected by merged https://github.com/LibreCAD/LibreCAD/pull/1840 and https://github.com/LibreCAD/LibreCAD/pull/1843.

### Steps to reproduce or sample file

1. Load latest LibreCAD `2.2.2_alpha`;
2. Open document (or create new one) with few Text and MText labels and some dimensions;
3. Try pan and zoom.

Sample DXF: [A4_Landscape_ISO7200TD_dxf.zip](https://github.com/user-attachments/files/16576202/A4_Landscape_ISO7200TD_dxf.zip)

### Operating System and LibreCAD version info

- Version: 2.2.2_alpha1-141-g0e0b66e8
- Compiler: GNU GCC 9.4.0
- Compiled on: Aug 22 2024
- Qt Version: 6.6.3
- Boost Version: 1.71.0
- System: Debian GNU/Linux 12 (bookworm)

sand1024:
That's really strange.. I can't see the regression you've mentioned on the file you've provided. Both zoom and pan work fine - yet, I'm on Windows, and probably such behavior is OS-specific. 
What is regression? Drawing speed? Zoom step? etc. 
I'll check this, yet I think we'll review rendering flow soon.

app4soft:
> What is regression? Drawing speed? Zoom step?

Drawing speed on Zoom in/out and Scroll up-down/left-right (as on pan/drag LibreCAD simplifies text rendering to just rectangle until pan/drag stops).

> That's really strange.. I can't see the regression you've mentioned on the file you've provided. Both zoom and pan work fine - yet, I'm on Windows, and probably such behavior is OS-specific.

Probably, because your PC is modern and has more powerfull CPU and more RAM.

Try to create drawing with a huge text sample (like _"Lorem ipsum dolore..."_ with few hundreds or thousands of characters) and try to zoom in/out and scroll left-right/up-down with LibreCAD builds before and after new featers merged.

I suppose development builds of LibreCAD should be always tested on low-end PC or on virtual machine with minimal settings (CPU <= 2GHz; RAM <= 2Gb), where its easy to see and detect regressions.

app4soft:
> That's really strange.. I can't see the regression you've mentioned on the file you've provided.

[...]

PR #1922, opened 2024-10-15
Rendering performance

sand1024:
### Rendering pipeline optimization
It possible to find mentions that LibreCAD is not suitable for large drawings due to slow rendering speed.  
It was not clear for me why LibreCAD renders drawing slower than other similar programs, so I've decided to dig into this deeper. 

Rendering pipeline ("paintEvent") was reviewed and optimized (by eliminating not needed/duplicated checks, pre-calculating values needed for paining, avoiding extra calls and so on).     

The primary goal was to achieve good/acceptable performance, especially for large drawings.  

While internally the code in some places now is less elegant, the reason for that was optimization of painting speed. 

As result, depending on specific entities included into drawing, the performance of rendering (on Windows machine) as increased. 

In my environment, rendering is up **to 7 times faster** comparing to previous AppImage (measurements for painting time may be enabled by `DEBUG_RENDERING` define in _rs_graphicview.h:42_).  

On other machines/operating systems, of course, results may be different.

NOTE: As changes in rendering are sensitive, despite that various cases were tested, it still might be that some painting issues/glitches may occur (as they may depend on actual data). 

Files for testing (opening time for them is long): 

lines_large.dxf - 190084 entities in drawing (mostly lines)
circles_large.dxf - 720409 circles
arcs_large.dxf - 180100 arcs
ellipse_large.dxf - 165747 ellipses 
polyline.dxf - 16402 polylines (each polyline has 16 lines/arc)
ellipse_solid - 39631 entities, solid fill
spline_points_large.dxf - 162020 spline point entities
splines.dxf - 3474 splines
solid.dxf - 60024 leaders 
lorem_one.dxf - 10 large MText entities

Archive with samples are attached

#### Other changes

1) Better preview for MText/Text entities (baselines for text lines are shown instead of boundary rectangle on preview) 
2) Additional button that allows to render entities with default length in non-draft mode. 
3) Restored snapping to grid if grid is not shown. 
4) New version check improvement

##### General Settings
1) Added settings to control colors/linetype/fill for selection overlay box
2) Added settings for minimum values that affects rendering 
3) Added "no grid" default value for new drawings
4) Added setting for default zoom factor used for scroll zoom

##### Drawing Settings
1) Added settings for specifying line joins and line caps for rendering. 

### Issues addressed: 

LibreCAD#1916

LibreCAD#1917

LibreCAD#1921

LibreCAD#1912

[LC_Tests.zip](https://github.com/user-attachments/files/17384612/LC_Tests.zip)
[LC_Tests_1.zip](https://github.com/user-attachments/files/17384632/LC_Tests_1.zip)


dxli:
Just wondering, whether bounding boxes can be cached as well. In cases of MText, lots of time consumed by calculating bounding boxes.

sand1024:
Hm... not sure I understand - you mean bounding boxes for the entire MText or individual lines/letters?  

Anyway, that's good point, I'll check the code deeper. 

For text lines, there is a caching of baselines (which are drawn instead of text for preview, panning and if the height of text is less than specified amount of pixels). 

Actually, for MText I suppose there is still a room for improvements exists - just due to the current implementation (as mtext is just a container of inserts with polyline(s) that defines single letter), but I suppose deeper rework is required. 

So far, the rendering of MText should be faster due to generic optimizations, as well as for caching pens (it was surprising, yet defining pen for entity as well as setting pen for QPainter are quite time consuming operations).

app4soft:
To note, there is still high drop of performance when scaling down to the level where LibreCAD starts to draw many lines with real width in scale (other than if zoom out to see all the lines 1px wide).

Related issue on many text objects rendering still might be in case here:

- https://github.com/LibreCAD/LibreCAD/issues/1846

Also, here is another testcase file: [pr-dxf.zip](https://github.com/user-attachments/files/17451308/pr-dxf.zip)

Here is screencast showcasing the performance drop in fast zoom in-out: [librecad-pull-1922.webm](https://github.com/user-attachments/assets/99e17c0b-fbe6-4271-8366-82e401cfd803)

sand1024:
hm.... I've done a quick test of the file provided... 

After quick investigation, I would say that performance drop for such lines with >1px screen width occurs if anti-aliasing is enabled. 

If one is disabled in the Application Preferences, on my environment I can't see such performance drop. 

If it is so - that's bad, as we got a performance limit of QT antialiasing algorithm, and it could not be improved rewriting painting pipeline in such way that it will use OpenGL -based painting. 

And even in such case, I still not sure whether performance will be better on systems without descent graphic cards and native GL support. 

So yes, I suppose the only one workaround at least for now for such cases is just to disable anti-aliasing.

app4soft:
> the only one workaround at least for now for such cases is just to disable anti-aliasing.

Could be then "Anti-aliasing" button added to "View" toolbar? (for quick toggle it if needed, instead of opening preferences dialog each time)

Such button exists in QCAD.

qcad-aa

sand1024:
@app4soft well, adding such button is good idea and it will be available in the next update.

However, I was wondering why lines in your drawing were looking with such rough steps even with antialiasing, and found interesting thing there. 

In the area where you've noted performance degradation (crossed polylines) there are 1080 overlapping polylines (one on top of another).  

While that potentially explains the reason for such drawing effect as well as the performance degradation there, I think the question is deeper - is it intentional to use such a big amount of polylines in drawing there (for testing or so)?  

I'm actually concerned that some not needed duplicates of entities may be created as result of some some action execution.

app4soft:
> I was wondering why lines in your drawing

I'm not the author of this DXF-file, just got it from somewhere.

Another testcase: [lorem-ipsum_dxf.zip](https://github.com/user-attachments/files/16741593/lorem-ipsum_dxf.zip)

- https://github.com/LibreCAD/LibreCAD/issues/1846#issuecomment-2308956015

PR #1963, opened 2024-11-16
compare Master Developer

emanuel4you:
compare Master Developer new updates


librecad/src/lib/information/rs_information.cpp
dxli, line 1: This file is older than the master branch.

It should be removed from the pull request, since the changes here only revert a valid bug fix in the master branch.

dxli:
Please rebase to the latest master, to avoid overwriting latest fixes.

I am not sure how it happens, as the version control should have taken care of this types issues.

dxli:
I have created a PR to your branch, so rs_information.cpp is excluded from this PR.

dxli:
Sorry, my mistake.

I didn't get the actual issue before my suggestion.

That actual issue:

we are building the releases using the oldest ubuntu available at github actions, and qt6 was not available in apt then. Our Qt is actually installed by github actions, not by apt.

I have added the changes directly to the master branch, so rebasing this PR to the master should fix the qt5compat issue.

sand1024 (changes requested):
I suppose PR should be reworked before merge to master. 

At least, there is a copy/paste code that is identical in different classes and one should be eliminated. 

For example, qg_lsp_commandedit.cpp and qg_py_commandedite.cpp - they are basically just copies of the existing commandedit from command widget. 

Also, are "qg_" prefix is correct for new clasess? 

Plus, it's not clear whether this pull request contains complete piece of functionality or just intermediate initial set of features. 

So I suppose it might be early to merge this.

PR #2007, opened 2025-01-06
add pixi

looooo:
Would be nice if LibreCAD can be build with pixi. Pixi is a cross-platform multilanguage package-manager [1].

I think there is not too much missing to make this work. Currently I am stuck with defining the boost-directories. Please let me know if anyone knows how to set them with qmake. I need to do something like this:

```
qmake6 INSTALL_PREFIX=$CONDA_PREFIX 
       BOOST_LIB_PATH=$CONDA_PREFIX/lib
       BOOST_INCLUDE_PATH=$CONDA_PREFIX/include/boost
```

[1] https://github.com/prefix-dev/pixi

### DIFF
--- a/.github/workflows/run_pixi.yml
+++ b/.github/workflows/run_pixi.yml
@@ -8,7 +8,7 @@ jobs:
       max-parallel: 3
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest, macos-13]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/run_pixi.yml
+++ b/.github/workflows/run_pixi.yml
@@ -18,4 +18,4 @@ jobs:
       with:
         pixi-version: v0.40.2
         cache: true
-    - run: pixi run install
+    - run: pixi run build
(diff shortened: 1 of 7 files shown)


looooo:
not sure why the compilation is not finishing locally. Would be nice if you can enable the ci's, so we can do further debugging.

looooo:
linux is failing with:
```
ui/main/lc_releasechecker.cpp:33:118: error: class 'LC_TagInfo' does not have any field named 'gnu_dev_major'
   33 | LC_TagInfo::LC_TagInfo(int major, int minor, int revision, int bugfix, const QString &label, const QString &tagName):major(major), minor(minor), revision(
      |     
```

dxli:
I will look into this error.

dxli:
Could you switch to cmake from qmake?

our snapcraft is using cmake.

looooo:
I switched to cmake but needed to make some changes for find_package(qt6). Also I added a install command at the end of the cmake. Seems to work fine on osx-arm64 and linux-64.

looooo:
windows fails with:
```
D:\a\LibreCAD\LibreCAD\librecad\src\lib\math\lc_quadratic.h(28): fatal error C1083: Cannot open include file: 'boost/numeric/ublas/matrix.hpp': No such file or directory
```

looooo:
this is ready for review.

looooo:
thank you, great

looooo:
ref: https://github.com/conda-forge/staged-recipes/pull/28941

PR #2056, opened 2025-03-03
Fix Weblate conflicts

qwilvove:
Hello,

Sorry, my last PR made conflicts in Weblate. Let's try to correct this with this PR.

Regards


lordofbikes:
I had to reset #2055, it should not have been merged at all, over 1600 files changed, this is for sure not the right solution.
This PR doesn't look better, I close it.
I'll take a look at this, but it may take some time.

PR #2163, opened 2025-05-27
Fix spelling error in dialog box

officialpranav:
Fix spelling error in Dialog box after opening DWG file


lordofbikes:
https://www.scribbr.com/commonly-confused-words/a-vs-an/

Issue #2210, opened 2025-06-22
Freeze on snap (infinite cycle in RS_EntityContainer)


sand1024:
@dxli 
Hi!
I've found quite a strange issue that leads to freeze when making a snap (on master branch).   The freeze is reproduceable on [LibreCAD-2.2.2_alpha1-429-gbb314fad1-win64.exe](https://github.com/LibreCAD/LibreCAD/releases/download/2.2.2_alpha-latest/LibreCAD-2.2.2_alpha1-429-gbb314fad1-win64.exe), so it's not related to my local modifications. 

The freeze is due to infinite cycle. That infinite cycle occurs in RS_EntityContainer::getNearestIntersection, here:  

![Image](https://github.com/user-attachments/assets/181f14df-60a8-461e-ae22-ad6026708a8b)

I suspect it's related to your recent changes in RS_EntityContainer, and the logic for the cycle termination is broken somewhere. 

Actually, that infinite cycle is hard to reproduce reliably - it seems to be floating that occurs just from time to time  (at least, I've got it several times).  
Potentially, active moving of the mouse may help.  I've got one on several different drawings, so it seems that it's not related to specific drawing data.  

However, it seems that this feeze is reproduceable on the attached drawing - just move mouse around. 

[lc_freeze.zip](https://github.com/user-attachments/files/20852045/lc_freeze.zip)

Snap modes that are enabled are as following:

![Image](https://github.com/user-attachments/assets/836c4f67-7d00-4d8d-bec6-7ee8e082d573)

dxli:
@sand1024 

I haven't been able to reproduce the issue. However, there's a design issue with RS_EntityContainer::first() and next():

the traversing is controlled by member variables "subcontainer" and "entIdx", it means any mult-threading traversing would trigger undefined behaviors. I feel we should redesign the first()/next() mechanism based on clean graph DFS traversing, and the traversing should not modify the target container.

My previous change was triggered by troubleshooting of #2179. Incorrect RS_EntityContainer copy ctor/assigner led to corrupted data models. The hatch corruption was only a symptom.

sand1024:
@dxli 
Oh, the situation is even worse indeed.... There is quite a serious design issue, and I can't believe it was not detected earlier. 

1) The current implementation means that it's not possible to use nested first/next() iterations (as **entIdx** and subcontainer are part of container)

2) what is worse - there is a find method in the RS_EntityContainer
```
[...]

PR #2217, opened 2025-07-08
EntityContainer Traverser

dxli:
Motivation: issue #2210: next to traverse the same entity container in parallel

1. Added a new class LC_ContainerTraverser;
2. The traverser holds traversing information, so the same container can be traversed in parallel;
3. Not resolved, if the container is modified between creation of a traverser and traversing, undefined behavior is expected;
4. Code updated to use LC_ContainerTraverser.


dxli:
@sand1024 

This is a PoC of the traverser concept.

Please note, more work is needed to simulate RS_EntityContainer::prevEntity(). Probably, it's not actually needed.

Feel free to improve or even redo it.

sand1024:
@dxli 
Hi!
Thank you for tagging me. I've really quickly checked the implementation (so far directly on the GitHub). 

If I'm correct, the overall idea is something like this—make a traversal, select relevant entities into the internal list (one is returned by LC_ContainerTraverser::entities()) and later iterate over that list, right? 

Of course, it will work fine for full traversal, but probably it might be improved a bit?  Or, probably, use different implementations: 
1) one is for one-directional traveral (just iterate over all entities) 
2) one that is needed for algorithms where previous/next semantics should be supported;

The points that are in my mind are as foollows:

1) The initial traversal might be a kind of overhead if the overall loop over entities is used just to find some entity (and break the cycle earlier). 

2) Plus there might be additional costs for collected vector management/resizing for large sets.  For simple iterations (say, set some property of matched entity) that might be avoided. 

I suppose using some kind of policy may be promising there. 

For example, potentially it might be reasonable to pass the instance of some policy to the taveral (as a functional interface) to the traverser? 

Such a policy, very roughly, might be as something like 

```
class LC_EntityTraversingPolicy{
  virtual  int accept(RS_Entity*, RS_EntityContainer* parent) = 0; // and based on the result of inspection of the entity and returned value, either call further proces() of it, or continue and go to the next entity or just complete traversal;
  virtual void process(RS_Entity*) = 0;   
}
```  
So the traversal will do generic traversal of entities tree, while the policy will decide whether specific entitiy or container should be included into traversal process, and the how the accepted entity should be processed. 

With such approach, collecting entities to the list might be achieved just by specific implementation of such a policy... 

3) From the other hand, with intial collecting entities, the processing of entities may be simplified indeed, so, for example, 
```
bool LC_ActionPolylineArcsToLines::hasArcsSegments(RS_Polyline *p) {
    lc::LC_ContainerTraverser traverser{*p, RS2::ResolveAll};
    for(RS_Entity* entity = traverser.first(); entity != nullptr; entity = traverser.next()) {
```

My be just replaced by 

`for(auto entity:  traverser.entities()`) {`

So I think more investigations and considerations needed there...  Probably it worth to analize all places of such traversal and check what for it's performed, and how traversed entities are processed.  

I fill that in most cases the approach with first/next might be replaced by straight iteration, yet need to dig deeper into this. 

Unfortunately, right now it will be tough for me to play with this, as I have to complete and commit the current feature I'm working on. As ussually, it took more time than I've initially planned :(

Yet I hope I'll be able to commit something in consistent state in 1-2 weeks, and so I'll be glad to pay more attention to this case.

dxli:
@sand1024 

There's actually also implementation for generator (as in python) type traversing, so it only traverses one node each step.

Actually, that's the majority of the implementation.

PR #2131, opened 2025-04-25
Ordinate Dimensions

sand1024:
This PR adds support of Ordinate dimension. 

Demo Video: https://youtu.be/56kGlFMSZlg

#### Generic support of the Ordinate Dimension

ordinate_dims

1) Ordinate Dimension entity added (for x and y ordinates)
2) Support of ordinate dimension entity was added to DXF reading/writing
3) Ordinate dimension properties dialog added
4) Support of ordinate dimension was added to informational cursor and entity info widget. 

Ordinate dimensions supports both WCS and UCS. 
ordinate_dims_ucs

#### Related actions

Several actions related to ordinate dimensions were added (located in Dimensions group): 
ordinate_dims_menu

1) Ordinate - the text command is **"dimord"**. Allows to create new ordinate dimension. 
The origin/definition point of the dimension - zero (0,0) point is located in the origin of the current Coordinate System (UCS or WCS).

Therefore, before creation of ordinate dimensions, a proper UCS should be set.  

2) Select Ordinates by base - allows the user to pick one ordinate dimension and select all ordinate dimensions within drawing with the same defining point and direction of horizontal.  

3) Ordinates Re-base - (command is **"dimordrebase"**). Allows to set the defining point for selected ordinate dimensions to the origin point of the current UCS. 

This command is useful for editing ordinate dimensions, if either base point was changed, or if dimension entities were moved. 

4) "Set UCS by Ordinate Dimension" - this action allows the user to set the UCS that corresponds the defining point and horizontal angle of the selected Ordinate Dimension. 

Information about Ordinate dimension is also shown by Informational Cursor and Entity Info Widget

ordinate_dims_info

### Other changes

2) Very initial support for DIMSTYLES (this is work in progress)
3) Very initial support of other dimension-related entities (also WIP)

In general, the next PR will include more changes for dimension entities and actions are planned.

### Small Fixes
1) A fix for styling of linear dimensions line (if the text is inline of dimension line) 
2) Selection of dimensions using entity type filter in Select Window was improved.

PR #2281, opened 2025-08-31
drw_entities: add missing return in DRW_Tolerance::parseCode()

jirislaby:
The below commit added DRW_Tolerance::parseCode() without actually returning anything from all except the default case. Return `true` in that case.

The compiler correctly complains:
```
src/drw_entities.cpp:1165:1: error: control reaches end of non-void function [-Werror=return-type]
```

Fixes: 944e6340fe5a (Ordinate Dimensions (#2131))

Noted there:
https://github.com/LibreCAD/LibreCAD/commit/944e6340fe5a1e17a4e44d9de25c78131bfb4cb5#r164908446


dxli:
Tecnically, dxfrw fixes should go directly to https://github.com/LibreCAD/libdxfrw/

Issue #1946, opened 2024-10-27
Garbled text in fonts containing eliptic arcs


SilverGreen93:
When opening an AutoCAD exported DXF, some of the text is shown as garbled, the arcs used are not drawn correctly.

### Expected behavior

This is how the text looks in AutoCAD

![autocad](https://github.com/user-attachments/assets/ad6ad5c8-46a0-4847-8c77-e137e1c6d668)

### Observed behavior

This is how the text looks in LibreCAD

![librecad](https://github.com/user-attachments/assets/d042c4d0-432b-424f-a106-bdac5a8b6ad6)

### Steps to reproduce or sample file

Use this file to reproduce:
[demo.zip](https://github.com/user-attachments/files/17533282/demo.zip)

### Operating System and LibreCAD version info

Version: 2.2.0.2
Compiler: GNU GCC 13.1.0
Compiled on: Jul 29 2023
Qt Version: 5.12.11
Boost Version: 1.75.0
System: Windows 11 23H2

lordofbikes:
based on https://forum.librecad.org/Text-missing-or-garbeled-in-DXF-file-tp5725500.html

dxli:
Sorry, pushed to a wrong issue.

https://github.com/LibreCAD/LibreCAD/commit/8f76d523855ce6bf8b931144c704474656f9fb5b

The idea, create all polylines with ellipses to support scaling.

This is clearly experimental, since the change affect all polyline creation.

If this is not acceptable, we will try to limit the changes to fonts only.

dxli:
![Screenshot_20241222_085936](https://github.com/user-attachments/assets/28994b36-6599-416f-be15-9a75cd05f349)

SilverGreen93:
> ![Screenshot_20241222_085936](https://github.com/user-attachments/assets/28994b36-6599-416f-be15-9a75cd05f349)

I see a commit referencing this issue. Is this fixed/pushed?

dxli:
@SilverGreen93 ，

The issue: polyline support for scaling (non-uniform, x-scale different from y-scale).

The fix: [0bd4277](https://github.com/LibreCAD/LibreCAD/commit/0bd4277ec3ef9b23306a0d9299cc7d5f5ecc6cb4) was reverted due to official file format (dxf) specifications, i.e., a polyline can only contain line and circular arcs. Elliptic arcs cannot be supported in polylines.

[...]

PR #2344, opened 2025-10-28
Issue1946: Text scaling improvement

dxli:
Issue #1946: Text Scaling Improvement

    The root cause: DXF polyline does not support elliptic arcs, causing
    distortion for anisotropic scaling of arc containing fonts, since the
    expected geometries after scaling are elliptic.

    The fix: added support for RS_Ellipse in RS_Polyline, but only arcs are
    used for saving. A new vertex extdata field is created for each elliptic
    arcs, to indicate the field is used by "LibreCAD" and keep the elliptic
    ratio.

    This fix may cause DXF compatibility issues.

### DIFF
--- a/librecad/src/lib/engine/rs.h
+++ b/librecad/src/lib/engine/rs.h
@@ -106,6 +106,7 @@ namespace RS2 {
     enum FormatType {
         FormatUnknown,       /**< Unknown / unsupported format. */
         FormatDXF1,          /**< QCad 1 compatibility DXF format. */
+        FormatDXFRW2018,           /**< DXF format. v2018. */
         FormatDXFRW,           /**< DXF format. v2007. */
         FormatDXFRW2004,           /**< DXF format. v2004. */
         FormatDXFRW2000,           /**< DXF format. v2000. */
--- a/librecad/src/lib/filters/rs_filterdxfrw.cpp
+++ b/librecad/src/lib/filters/rs_filterdxfrw.cpp
@@ -2559,9 +2559,13 @@ bool RS_FilterDXFRW::fileExport(RS_Graphic& g, const QString& file, RS2::FormatT
         exportVersion = DRW::AC1018;
         m_version = 1018;
         m_exactColor = true;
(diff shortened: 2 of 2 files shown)


sand1024:
@dxli 
Hi!
Probably I've missed something, yet hotestly I'm not sure I understand why you've extended quite common class RS_Polyline just to support specific case related to fonts.  

As far as I see - polyline that includes ellipses should be hardly persistent. The issue with elliptic segment of polyline occurs just due to hacky implementation of fonts rendering  - where paths for characters are modelled via polylines. And of course, during scaling arc segment should be transformed into ellipse.. 

Yet anyway - it seems such polyline is rather runtime utility entity used for rendering fonts. If, for example, fonts are rendered similarly to SHX or so - there will be no need for such entity, correct? 

You intent is quite clear, yet with extending polyline I suppose there are several new issues and questions occurs. 

Just several quick one that comes into my mind: 

1) Such polyline with elliptic segments becomes incompatible with normal polylines that are supported by DXF and lots of CAD programs (at least, AutoCAD does not support elliptic segments, AFAIK). 

2) It's not clear how it existing commands that works with polylines should interact with such extended polylines. Should they also be expanded to support ellipses? Should they ignore them? Which meta information should be shown for such polylines? How to make them consistent? 

3) What happens if some character is exploded? And such polyline that was part of the character will become normal polyline? 

As far as I understand, the only real case for existing of such polyline is just rendering of texts. 

But what I don't understand - why this quite a specific functionality added into the basic polyline. 

For me, it's much more natural just to create separate polyline-like entity that will include support of elliptic arcs - and that entity will be used just for character rendering - as part of symbol block.  

Of couse, such entity may be inherited from RS_Polyline, if necessary (yet in general it's just a description of path and some methods from polyline may be just not needed). 

However, using different entities will deliver proper separation of consern and will not mix different logic there (as well as will not mix the code and affect DXF compatibility). 

Of course, it might be that I've missed some points and current approach is ok - yet please consider my considerations above.

PR #2418, opened 2025-12-29
Implement case-insensitive command matching.

### LINKED ISSUE #2407: Feature Request: Make Command Line input case-insensitive
Feature Request: Make Command Line input case-insensitive

Description Currently, the "Command Line Widget" appears to be case-sensitive. All main internal commands (e.g., angleline, arc, circle) are defined in lowercase.

If a user types a command in uppercase or mixed case (e.g., ANGLELINE, Circle, ARC), the widget fails to recognize the command, even though it matches a valid long-form command.

Expected Behavior The Command Line Widget should normalize input to lowercase. If the user types ANGLELINE, ARC, CIRCLE, or CiRcLe, the widget should evaluate it as the lowercase main command (angleline, arc, circle) and execute it successfully.

The proposed implementation is to append the toLower() function to the command string processing within QG_CommandWidget::handleCommand.

librecad/src/ui/dock_widgets/command_line/qg_commandwidget.cpp Location: Around line 199

Current Code:

```
void QG_CommandWidget::handleCommand(QString cmd) {
    cmd = cmd.simplified();
    // ...
}
```
Proposed Change:

```
void QG_CommandWidget::handleCommand(QString cmd) {
    // Convert input to lowercase to match internal command definitions
    cmd = cmd.simplified().toLower(); 
    // ...
}
```
@dxli @lordofbikes

Issues related #1796, #1807, #1975

jonathankrocha:
Updated the CLI logic to support case-insensitive inputs.

- Normalizing command line input to lowercase.

- Ensuring the internal keyword list is stored in lowercase to guarantee a match.

Resolves #2407

### DIFF
--- a/librecad/src/cmd/rs_commands.cpp
+++ b/librecad/src/cmd/rs_commands.cpp
@@ -160,49 +160,40 @@ RS_Commands::RS_Commands() {
     for(auto const& [fullCmdList, aliasList, action]: g_commandList){
         //add full commands
         for(auto const& [fullCmd, cmdTranslation]: fullCmdList){
-	    // Create lowercase versions for map keys
-            QString fullCmdLower = fullCmd.toLower();
-            QString cmdTransLower = cmdTranslation.toLower();
-
             if (fullCmd == cmdTranslation)
                 continue;
             // use translated commands first
-	    // Use lowercase for keys in m_cmdTranslation and m_mainCommands
-            if (isCollisionFree(m_cmdTranslation, fullCmdLower, cmdTranslation))
-                m_cmdTranslation.emplace(fullCmdLower, cmdTranslation);
-            if (isCollisionFree(m_mainCommands, cmdTransLower, action, m_actionToCommand.count(action) ? m_actionToCommand[action] : QString{})) {
-                m_mainCommands.emplace(cmdTransLower, action);
+            if (isCollisionFree(m_cmdTranslation, fullCmd, cmdTranslation))
+                m_cmdTranslation.emplace(fullCmd, cmdTranslation);
+            if (isCollisionFree(m_mainCommands, cmdTranslation, action, m_actionToCommand.count(action) ? m_actionToCommand[action] : QString{})) {
+                m_mainCommands.emplace(cmdTranslation, action);
                 m_actionToCommand.emplace(action, cmdTranslation);
             }
         }
         for(auto const& [fullCmd, cmdTranslation]: fullCmdList){
-	    QString fullCmdLower = fullCmd.toLower();
-            if(isCollisionFree(m_mainCommands, fullCmdLower, action, m_actionToCommand.count(action) ? m_actionToCommand[action] : QString{})) {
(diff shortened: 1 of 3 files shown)


librecad/src/cmd/rs_commands.cpp
```diff
@@ -160,40 +160,49 @@ RS_Commands::RS_Commands() {
     for(auto const& [fullCmdList, aliasList, action]: g_commandList){
         //add full commands
         for(auto const& [fullCmd, cmdTranslation]: fullCmdList){
+	    // Create lowercase versions for map keys
+            QString fullCmdLower = fullCmd.toLower();
+            QString cmdTransLower = cmdTranslation.toLower();
```
dxli: QString::toLower() uses the "C" locale. Recommended to use QLocale::toLower().

Just a suggestion. I'm not aware of usage cases when the conversions are different.

dxli (commented):
I still prefer using a user preference to use case-insensitive commands.

This PR remove the capability of case-sensitive commands, so it's more restrictive than adding the possibility of case-insensitive support.

In long run, I suggest keep the case-sensitive design, because I can imagine when most commands are generated by programs, and when case sensitivity may not be an issue itself.

Thanks!

jonathankrocha:
I agree that keeping the case-sensitive design is the right call. I have reverted the previous changes to restore that capability.

Instead, I’ve taken a different approach: case-insensitivity is now applied specifically to Tab-completion logic. This allows users to find available commands via Tab even if they have already typed uppercase characters, while still maintaining the case-sensitive integrity of the commands themselves.

dxli:
Thanks for working on this, @jonathankrocha, and for reverting the first version when case-sensitive aliases came up.

I'm closing this in favour of #2865. Here's why:

- **Tab completion already ignores case on master.** `RS_Commands::complete()` and `tabPressed()` in `qg_commandwidget.cpp` already use `Qt::CaseInsensitive` (the lines this PR replaces use it too). So after the revert, this PR doesn't change how commands are matched.
- **What's left is a restyle, plus one removal.** The diff mostly reorganises `tabPressed()`. It also drops the `line`/`rectangle` entries from `QG_ActionHandler::getAvailableCommands()`, which is unrelated. The branch now conflicts with master as well.
- **Issue #2407 is still open.** Typing `LINE` or `Circle` and pressing Enter still does nothing.

#2865 fixes that while keeping the case-sensitive aliases you wanted to preserve. The exact match is still tried first. If it finds nothing, the command is matched ignoring case, but only when every match names the same action. For example, aliases `Zq` and `zQ` still run their own actions, and `ZQ` stays unmatched.

Reviews on #2865 are welcome.

sand1024:
@dxli 
Well, I think support of case sensitive commands may complicate the user and I still not sure that it delivers some practical benefits, yet anyway - that will work, so I'm fine with it.
 
The only thing - I work now on UI editor for commands aliases, so just please update/tag me if you'll modify something in RS_Commands, QG_CommandWidget, so I'll be able to incorporate related changes.
Thank you.

PR #2433, opened 2026-01-10
To Qt6

PushKK:


dxli (changes requested):
It appears to be outdated to use QT_REQUIRE_VERSION():

Modern Alternatives
In 2026, most developers handle version requirements during the build process rather than at runtime. This prevents the application from even being compiled if the environment is unsuitable:
CMake: Use find_package(Qt6 6.4 REQUIRED) in your CMakeLists.txt.
Compile-time Check: Use the QT_VERSION macro and #if guards to verify the version during compilation.
Runtime Version Check: If you need to check the version at runtime without terminating the app immediately, use qVersion() to get the current version string.

Issue #2431, opened 2026-01-09
Version not 2.2.2


PushKK:
Version LibreCAD from this GIT not show version number:
<img width="566" height="428" alt="Image" src="https://github.com/user-attachments/assets/d59093db-39de-4633-9e93-581e056619ab" />

sand1024:
I think version shown there is defined by LC_VERSION (that may come from build script or not) - main.cppp:61

dxli:
Just checked the builds for the master branch: both linux and windows-arm64 show versions, instead of the commit hash.

PushKK:
<img width="566" height="428" alt="Image" src="https://github.com/user-attachments/assets/20cf1a01-df8c-4d57-bfdc-878783561e2e" />

lordofbikes:
see #2441

PR #2441, opened 2026-01-14
"LC_VERSION" bug

PushKK:
For https://github.com/LibreCAD/LibreCAD/issues/2431

### DIFF
--- a/librecad/src/src.pro
+++ b/librecad/src/src.pro
@@ -13,7 +13,8 @@ DEFINES -= JWW_WRITE_SUPPORT
 LC_VERSION="2.2.2.6-alpha"
 LC_PRERELEASE = "true";
 
-VERSION=$${LC_VERSION}
+DEFINES += LC_VERSION=\"$${LC_VERSION}\"
+DEFINES += LC_PRERELEASE=\"$${LC_PRERELEASE}\"
 
 # Store intermedia stuff somewhere else
 GENERATED_DIR = ../../generated/librecad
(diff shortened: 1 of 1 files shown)


lordofbikes:
This is not the correct solution, if there is an issue at all.
`LC_VERSION` is modified with `git` system calls to show an unique version, that gives the exact code base for the binary.
Thus moving these two lines up ignores the `git` hash modifications.
This is part of the automated build process and a valid version tag will return the release name in a release build.
All other builds will have the `git` hash in the version string, which is essential for support and debugging.

lordofbikes:
To be correct, `git describe` returns a tag name, which might have a hash appended and a counter, if the latest commit has no tag.

dxli:
@PushKK 

Could you investigate this issue more?

Also, needs to make the LC_VERSION concistent among our CI builds: as both qmake and cmake are used.

Ideally both cmake and qmake should read the same file to get LC_VERSION to use, and CI building should only modify a single file to set the version.

PushKK:
In LibreCAD (master branch) I see LC_Version in:
```
- CMakeLists.txt:  project(LibreCAD VERSION 2.2.2)
                               LC_VERSION=2.2.2.5-alpha;
- src.pro:               LC_VERSION="2.2.6-alpha";
- main.cpp:           QCoreApplication::setApplicationVersion("2.2.2");

```
And it work fine.
It's only digits and letters. And can be any?
Ubuntu have version "25.10" (year and month). And work fine too.

`LC_VERSION is modified with git system calls to show an unique version, that gives the exact code base for the binary. Thus moving these two lines up ignores the git hash modifications.`
Yes, but this string "2.2.2.5-alpha" for users only. Git use own hash modification.

lordofbikes:
Recommendation for users is to use release versions, where `git describe` returns **2.2.2** or whatever the maintainer tagged the latest commit for the release build.

Users who use nightly builds or build librecad themselves are considered as tester or developer. Then it is important, e.g. if they report an issue, that the version string contains the `git` hash. Thus the team can check if an issue is probably already solved in a younger commit. This helps much to avoid wasting the team's time.

dxli:
@PushKK 

The goal is to get correct and consistent versions for our CI builds, and developer builds(if possible).

Of course, it's not a hard-coded string, but setting during CI building. Have a look at our continuous builds.

PR #2446, opened 2026-01-18
 I18n: Update .ts files in unix

PushKK:
I add 'lupdate' for update all *.ts files.


PushKK:
Yes, this patch insert tr() to *.ts from 'librecad/src/actions/not_used/'...
May be it not used and can be deleted?

dxli:
outdated. closing now.

Issue #2399, opened 2025-12-08
Widget Right dock very wide by default


Questions2000:
### Observed Behavior
I notice this/experience more of an issue with this because I am running tests as oppose to general user who will only experience this rarely.

If you delete your `LibreCAD` folder from `.config` folder (Linux users) and then launch LC AppImage, it opens up LC with a fresh start, showing Welcome dialog.

I noticed that the default Right dock width chosen seems very wide.
For me it opens 50-75% the width of LC window.
Then have to keep adjusting this setting every time I am running tests.

Maybe would adjust to make more like 25-30% instead for the default, if anyone agrees

Thank You

### Operating System and LibreCAD Version
LibreCAD-2.2.2_alpha1-506-g7c221562c.AppImage
Compiler: GNU GCC 13.3.0
Compiled on: Nov 29 2025
Qt Version: 6.4.2
Boost Version: 1.83.0
System: Debian GNU/Linux 12 (Bookworm) (X11)

PushKK:
Right dock created with maximum size.
It needed for different display size - 1280 to 2440 pixel.

For change window size may be you can do it:

lc_widgetfactory.cpp:
```
127    QDockWidget* LC_WidgetFactory::createDockWidget(const QString& horizontalTitle, const char *name, const QString& verticalTitle) const {
128        auto result = new LC_DockWidget(m_appWin, horizontalTitle, verticalTitle);
129        // auto result = new QDockWidget(horizontalTitle, m_appWin);
130        result->setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Maximum);
```

lopestom:
Well, I didn't want to write, since I no longer have daily user experience.
But I liked LibreCAD so much that I needed to comment on some parts that could improve its use, mainly considering that it's for the user and the drawing itself, in short, productivity.
<details><summary>I really thanks</summary>
<p>

[...]

PR #2459, opened 2026-01-30
UI: Fix Qt_DockWidget

PushKK:
For this issue:
https://github.com/LibreCAD/LibreCAD/issues/2399

### DIFF
--- a/librecad/src/ui/dock_widgets/ucs_list/lc_dlgucslistoptions.ui
+++ b/librecad/src/ui/dock_widgets/ucs_list/lc_dlgucslistoptions.ui
@@ -26,14 +26,14 @@
          <string>If сhecked, the tooltip with additional information about UCS will be shown for table item.</string>
         </property>
         <property name="text">
-         <string>Toolip with UCS details for table items</string>
+         <string>Tooltip with UCS details for table items</string>
         </property>
        </widget>
       </item>
       <item row="0" column="0">
        <widget class="QCheckBox" name="cbShowTypeIcon">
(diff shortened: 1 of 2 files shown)


dxli:
@sand1024 

Do you have a better idea to get this one fixed?

dxli:
I pushed a commit: f4a85644b5b10532d6ebcec02827c61c560e0eb9

The idea is to set a default width for the dock widgets, if no saved geometry is found.

This idea works better:

 it only affects initial opening, but does not limit user ability to adjust the size of docking area;

sand1024:
> I pushed a commit: [f4a8564](https://github.com/LibreCAD/LibreCAD/commit/f4a85644b5b10532d6ebcec02827c61c560e0eb9)
> 
> The idea is to set a default width for the dock widgets, if no saved geometry is found.
> 
> This idea works better:
> 
> it only affects initial opening, but does not limit user ability to adjust the size of docking area;

Yes, I think this is more than reasonable - after initial start, the user will adjust sizes accordingly. 

In general, there is related thing - existing "initial dialog" for the first run seems to be quite outdated. 

There are lot's of UI (and UI layout) customization options, yet I'm sure most of users are not aware of them (like docking, graphic view appearance settings/color and so on). 

So it may be reasonable to either expand initial dialog or creates separate one (like "Tip of The Day"). 

I'm thinking about a kind of wizard with series of hints or tips that educates the user regarding features of LC. Potentially - such tips may be even suggested by users (yet in this case I'm not sure how to handle localization).

PR #2471, opened 2026-02-10
feat: show the command next to menu item

silasb:
When learning librecad it was difficult for me to know what menu item mapped to what command.  My goal is to make myself more productive and I'm hoping this will help others as well.

This change makes it very obvious by updating the menu item to include the command name, secondary command name (not sure what to call it), and the alias, if one exists.

I understand that this could be considered noisy.  Perhaps if this becomes noisy enough it could be a preference item to enable/disable this, but default it to enabled?

I can backport this to 2.2.1 as well.

image

image

### DIFF
--- a/librecad/src/cmd/rs_commands.cpp
+++ b/librecad/src/cmd/rs_commands.cpp
@@ -438,6 +438,24 @@ QString RS_Commands::command(const QString& cmd) {
 
 
 
+QStringList RS_Commands::aliasesForAction(RS2::ActionType a) const {
+    QStringList res;
+    for (auto const& p : m_shortCommands) {
+        if (p.second == a) {
+            res << p.first;
+        }
+    }
+    auto it = m_actionToCommand.find(a);
+    if (it != m_actionToCommand.end()) {
+        // ensure main command present and first
+        if (!it->second.isEmpty() && !res.contains(it->second)) {
+            res.prepend(it->second);
+        }
+    }
+    return res;
+}
+
+
 /**
  * Checks if the given string 'str' matches the given command 'cmd' for action
  * 'action'.
--- a/librecad/src/cmd/rs_commands.h
+++ b/librecad/src/cmd/rs_commands.h
@@ -61,6 +61,8 @@ public:
     RS2::ActionType cmdToAction(const QString& cmd, bool verbose = true) const;
     RS2::ActionType keycodeToAction(const QString& code) const;
 
+    QStringList aliasesForAction(RS2::ActionType a) const;
+
     static QString command(const QString& cmd);
 
     static bool checkCommand(const QString& cmd, const QString& str,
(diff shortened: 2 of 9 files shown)


sand1024:
@silasb 
Well, the generic intention there is clear, yet from the first glance it seems that current implementation is noisy indeed. I'm sure it's necessary to let the user to control whether commands are shown or not there (so appropriate settings UI and options are needed), plus displaying all variants of commands may be just overhead - so that's my be an option too. 

There is another questionable thing - it's consistency. The action may be shown not only in menu - but in toolbar or in other places - so the good question is whether the command alias should be shown in tooltip of action or not. 

Please note that I had no chance to review your changes in commit yet, so above notes are just very quick comments. 

Actually, I plan to rework the entiring approach for command line widget as well as commands support, plus UI for management of commands will be helpful - so I suppose that approach will be mixed anyway.

silasb:
Thanks @sand1024. The current structure is definitely noisy because I’m looping through multiple data sets to normalize menu items into commands. From a UI perspective, the clutter stems from how many string aliases are mapped to a single command.

Refactoring this would streamline both the UI and the CLI, but it would likely involve breaking changes—such as removing shorthand aliases (e.g., losing c for 'circle') that users rely on.

If we wanted to make the UI more streamlined we could pick the shortest alias/command that exists for said menu item, but this might be confusing for some folks as well.

I've attached another image showing the context menu, recent menu, and command line both reflecting the change here.  Not pictured is the widget, but those also reflect the change.

I can create a new preference to support this if this path makes.

dxli (approved):

dxli:
@silasb 

also, implementation of aliasesForAction() is not perfect due to the linear searching time. I suppose we can reorganize the RS_Command data organization to get better. I understand this is not a bottleneck yet, but we prefer a clear logic design: simple mapping lookup is clearer than manually searching through the map.

Some minor coding style: C++17 structured binding helps for map iteration.

I have no trouble with merging this PR, as we can always improve with a second PR.

sand1024:
@dxli 
Well, I suppose we may merge this PR as it - I'll integrate it, as I'll also add an additional feature to educate the user and promote commands. 

As for implementation of aliasesForAction() - well, that's hardly an issue for now, as actually the amount of commands is small. Yet from the generic design point of view - yes, I suppose RS_Commands needs a facelifting as well as the function of obtaining commands by action type. I plan to review cmd widget with time, and probably there will be a chance to refresh. 

But as a generic note - are many aliases for the same command reasonable? It seems like a kind of overkill for me and I really doubt that some user actualy use, say 3 or 4 (or even 2) forms of the same commmand. Most probably, only one form is always invoked. 

Most probably, even for the most demanding cases, support of two forms (like full command and short alias) should be enough for majority of users. 

And from the implementation point of view - having just 2 names for one command will definitely allow simpler and cleaner implementation. 

What do you think of this?

sand1024:
This feature will be supported, yet in different way (a bit less visually noisy), so it will look something like

image

So I think this PR may be just closed without necessity to merge - yet anyway, it was a good inspiration.

PR #2489, opened 2026-03-05
add a new action: dual

dxli:
New action: RS2::ActionDrawDual

Command to trigger: dual

1. Select entities;
2. Command: dual;
3. dual geometries of selected will be created.

This action is not ready for general usages, so not added to menu entries yet.


sand1024:
@dxli 
well, I'm sure there were some serious reasons to merge uncompleted functionality into master instead of keeping it in separate branch until completion... Yet in general, I'm not sure that this is a good practice - let's keep master in operational state without at least "known" issues.

Issue #2501, opened 2026-03-13
Request: Configure alternative Spanish translations to default to ES_ES


JuloWaks:
I've discovered this project while studying in an institute that provides profesional formation.
I wanted to start by thanking for making posible to draw CAD without privative licensing schemes.

I came to realize that some of the students would struggle because the institute configured ES_AR and according to weblate the translation is at 30% 

Would it be possible to configure, where required, that the "unfinished" chains would default to ES_ES for ES_*

For what I understand it is possible to configure some sort of hierarchy in weblate for this kind of cases.

Could you take a look or let me know what other information I can provide so we can make this happen?

dxli:
@JuloWaks 

I'm aware of the issue of overdued translation updating.

weblate appears to be a little bit outdated. Nowadays, free AI translation should be good enough for any language we would like to support.

[libretranslate](https://libretranslate.com/) supports free self-hosted translation, but not for all languages.

For languages not supported by libretranslation, using local LLM models to complete our translation.

While we do have enough budget for paid services, it's much easier to rely on free services for developers to obviate accounting troubles.

I will quickly update the 2.2.1 branch to add translations for all languages. Please report translation errors here, so the procedure could be improved.

If you could help training local LLM models for translation, it helps to speed up translation also.

dxli:
@JuloWaks 

I have merged PR #2502 to update translations for 2.2.1

Please try the latest 2.2.1 builds and evaluate the AI translations

JuloWaks:
Hey, I will be testing translations over the weekend, I can also review the PR if you want

dxli:
@JuloWaks ,

Could you review PR #2512 ?

dxli:
@JuloWaks 

I have merged PR #2512.

This is a large PR, and I tested on a few platforms: win64, linux-arm64, etc.

PR #2502, opened 2026-03-14
Issue #2501: update translations

dxli:
1. added a script to translate ts files;
2. updated translations in ts files.

Since the translations are by libretranslate, feedbacks are welcome and will help us improve the translation procedure.

### DIFF
--- a/scripts/ts_libre.py
+++ b/scripts/ts_libre.py
@@ -1,9 +1,23 @@
-### translate ts files by local hosted libretranslate
-### to start: 
-###   pip install libretranslate
-###   libretranslate --port 5000
-###   python ts_libre.py 
-### the python script will translate ts files in the subfolder translations/
+"""
+PURPOSE:
+    Automates the translation of Qt Linguist (.ts) files using a local or remote 
+    LibreTranslate API instance. It specifically handles regional locale codes 
+    (e.g., zh_TW, es_MX) by mapping them to API-compatible ISO codes and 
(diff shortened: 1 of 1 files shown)


lopestom:
> 1. added a script to translate ts files;
> 2. updated translations in ts files.
> 
> Since the translations are by libretranslate, feedbacks are welcome and will help us improve the translation procedure.

Unfortunatelly the LibreCAD-v2.2.1.4-4-gc28b342fc-x86_64.appimage not show ALL pt_BR language. Since with LibreCAD-v2.2.1.4-3-g027b93da3-x86_64.appimage is Okay.
I'm trying to translate it in a technical and simplified way (when possible) into Brazilian Portuguese (I didn't bother to find out if there are others here already doing this). But it still requires time and testing for all strings vs. spell checking vs. visual correction.
The idea of ​​AI is interesting, but it's better when we have users who want to contribute in a real way to a better understanding of the options. I understand that there aren't many of them...........

Edit
Okay. I believe the translation in pt_BR with AI is not displayed due to significance errors.

```
radius=%1 is invalid (negative)
```
```
raio=% 1 é inválido (negativo)
```
In many translations there is a space between `%` and `1` (and `2` as well) as well as a lack of `\n`.

dxli:
@lopestom 

The basic idea is to use AI as a start point, and users can always send in corrections.

The current translations are from libretranslate, and for particular languages, we can also try other AI models

PR #2510, opened 2026-03-19
UI: Change controls order and size

PushKK:
Change:
- order of controls,
- size of Label,
- text in Label.


sand1024:
this is obsolete, no need to merge

PR #2513, opened 2026-03-27
Reverting key CLI commands for CAD agnostic compatibility

### LINKED ISSUE #1683: Commands not working
### Expected behavior

command is typed in command bar,
the tool is run as enter is pressed

### Observed behavior

Unknown command: xxxx 

is displayed 
where xxx is
fillet
offset
o
bevel

Windows 11
2.2.1_alpha-276

### LINKED ISSUE #1833: Offset & Parallel conflict with one another
### Expected behavior

Offset should allow offset creation for any shape, not just straight lines.

### Observed behavior

Observe the commands that appear in command line when you select the following options (using toolbar buttons):

- Modify > Offset = Offset: offset, o (this command will allow you to offset a rectangle only when selected using the button)
- Lines > Parallel = Command: o (offset) (this command will only allow you to add parallel lines to straight lines when using the button)
- The key commands pa or o are overlapping/shared and also only activate the Parallel version of the offset command
- When trying to then use the tool that is activated using key commands on a rectangle, the program does not recognize it as a rectangle and does not allow offset to be used, only Parallel on individual lines in the rectangle
- Thus it is not possible to activate the true Offset tool using key commands

Mac OS
LC 22.0.2

pgmorgan:
Somewhere between 2.2.0 and 2.2.1 key CLI commands in LibreCAD were changed. These represent a serious shift for long time CAD users. Previously the CLI had major overlap with AutoCAD CLI which goes back decades and has a strong historical user base. This platform agnostic CLI was a major draw for users coming to and from LibreCAD or AutoCAD.

This was discussed in Issue #1683 and other users chimed in about this matter.

Fixes #1683
Fixes #1833

## Update: reworked to add, not remove

The original patch simply reverted the `mod*`-prefixed names back to their pre-2.2.1 equivalents, which meant removing the newer names. That traded one set of broken muscle memory for another. It's been reworked around a different rule instead, so nobody loses a command they're already used to:

- the most reasonable/cross-platform name becomes the **main** command for its action
- whatever it displaces becomes a **short** command on the same action, so it still works
- nothing that works before this PR stops working after it

Main commands are now the names shared with AutoCAD, BricsCAD, DraftSight and QCAD: `rotate`, `scale`, `mirror`, `trim`, `explode`, `fillet`, `bevel` (+ `chamfer`), `offset`, and `parallel`. The `mod*`-prefixed names from 2.2.1 (`modrotate`, `modscale`, `modmirror`, `modtrim`, `modexplode`, `modfillet`, `modbevel`, `modoffset`) are kept as short commands, alongside `linepar`/`lineoff` for the parallel-line tool.

### `offset` stays on the Modify > Offset action (not moved to Parallel)

The original patch moved `offset` to be an alias of the parallel-line tool, matching its pre-2.2.0 meaning. That would re-open issue #1833: `RS_Creation::createParallel` only handles lines, arcs, circles and parabolas, so the parallel tool can't offset a polyline, whereas `RS_Polyline::offset()` is implemented and reachable through the real Modify > Offset action. AutoCAD, BricsCAD and DraftSight all define OFFSET over polylines too, and QCAD - the project LibreCAD forked - binds `offset` to its own Modify > Offset tool and `lineoffset`/`parallel` to the parallel tool, i.e. the same split LibreCAD master already has. A command name can only map to one action, so this had to be a choice, not an alias, and issue #1833 is the reason it was chosen this way originally.

This also restores the `o` short command on Modify > Offset, which has been unbound since the fix for #1833 (the toolbar button has displayed "Offset : offset, o" as its available commands for about two years without `o` actually working).

### `bevel` changes meaning - the one deliberate breaking change

`bevel` currently runs the *fillet* action (asks for a radius) because it was added to fillet's alias list by mistake in c2e41bfb7. It now runs the actual bevel/chamfer action (asks for two lengths), matching LibreCAD 2.2.0, LibreCAD's own `CORNER_BEVEL` vs `CORNER_RADIUS` terminology, and the Bevel menu entry. Also added `chamfer` as a main command, since that's the name AutoCAD/BricsCAD/DraftSight actually use - "bevel" as a bound command is essentially QCAD/LibreCAD-only.

### Verification

No duplicate main commands, no duplicate short commands, and no command name appearing as both main and short, anywhere in the table - checked before and after merging current master. The table also compiles cleanly in both states.

---

<details>
<summary>Original PR description</summary>

I am submitting this PR to revert LibreCAD's CLI compatibility on key commands that I use regularly as a CAD user.

Note that `offset` is now an alias for `lineoff` and not for `modoffset`, in keeping with the reversion to a more platform agnostic CLI. `offset` was used for this up to v2.2.0.

Here are the changes I originally identified to revert this:

```
incompatible command --> compatible command

lineoff    --> offset
modrotate  --> rotate
modscale   --> scale
modmirror  --> mirror
modtrim    --> trim
modoffset  --> offset
modbevel   --> bevel
modexplode --> explode
```

Other critical commands maintain good compatibility like `l`, `pl`, `c`, `move`, `mtext`, `dimlinear`, and many other common commands. But the above list is the scope of the regression I identified.

</details>

### DIFF
--- a/librecad/src/cmd/lc_commandItems.h
+++ b/librecad/src/cmd/lc_commandItems.h
@@ -255,7 +255,8 @@ const LC_CommandItem g_commandList[] = {
         {
             {{"linepar", QObject::tr("linepar", "create parallel")},
              {"parallel", QObject::tr("parallel", "create parallel")},
-             {"lineoff", QObject::tr("lineoff", "create parallel")}},
+             {"lineoff", QObject::tr("lineoff", "create parallel")},
+             {"offset", QObject::tr("offset", "create parallel")}},
             {{"pa", QObject::tr("pa", "create parallel")},
              {"ll", QObject::tr("ll", "create parallel")}},
             RS2::ActionDrawLineParallel
@@ -702,19 +703,22 @@ const LC_CommandItem g_commandList[] = {
         },
         // rotate
         {
-            {{"modrotate", QObject::tr("modrotate", "modify - rotate")}},
+            {{"modrotate", QObject::tr("modrotate", "modify - rotate")},
+             {"rotate", QObject::tr("rotate", "modify - rotate")}},
             {{"ro", QObject::tr("ro", "modify - rotate")}},
             RS2::ActionModifyRotate
         },
         // scale
         {
-            {{"modscale", QObject::tr("modscale", "modify - scale")}},
+            {{"modscale", QObject::tr("modscale", "modify - scale")},
+             {"scale", QObject::tr("scale", "modify - scale")}},
             {{"sz", QObject::tr("sz", "modify - scale")}},
             RS2::ActionModifyScale
         },
         // mirror  (Removed extra space from translation sting.)
         {
-            {{"modmirror", QObject::tr("modmirror", "modify -  mirror")}},
+            {{"modmirror", QObject::tr("modmirror", "modify -  mirror")},
+             {"mirror", QObject::tr("mirror", "modify -  mirror")}},
             {{"mi", QObject::tr("mi", "modify -  mirror")}},
             RS2::ActionModifyMirror
         },
@@ -739,7 +743,8 @@ const LC_CommandItem g_commandList[] = {
         },
         // trim
         {
-            {{"modtrim", QObject::tr("modtrim", "modify - trim (extend)")}},
+            {{"modtrim", QObject::tr("modtrim", "modify - trim (extend)")},
+             {"trim", QObject::tr("trim", "modify - trim (extend)")}},
             {{"tm", QObject::tr("tm", "modify - trim (extend)")}},
             RS2::ActionModifyTrim
         },
@@ -758,26 +763,24 @@ const LC_CommandItem g_commandList[] = {
         },
         // offset
         {
-            {{"modoffset", QObject::tr("modoffset", "modify - offset")},
-             {"offset", QObject::tr("offset", "modify - offset")}
-         },
+            {{"modoffset", QObject::tr("modoffset", "modify - offset")}},
             {{"mo", QObject::tr("mo", "modify - offset")},   // - v2.2.0r2
              {"moff", QObject::tr("moff", "modify - offset")}},
             RS2::ActionModifyOffset
         },
         // bevel
         {
-            {{"modbevel", QObject::tr("modbevel", "modify - bevel")}},
+            {{"modbevel", QObject::tr("modbevel", "modify - bevel")},
+             {"bevel", QObject::tr("bevel", "modify - bevel")}},
             {{"bev", QObject::tr("bev", "modify - bevel")},
              {"ch", QObject::tr("ch", "modify - bevel")}},
             RS2::ActionModifyBevel
         },
         // fillet
         {
-            {{"modfillet", QObject::tr("modfillet", "modify - fillet")}},
-            {{"fi", QObject::tr("fi", "modify - fillet")},
-             {"fillet", QObject::tr("fillet", "modify - fillet")},
-             {"bevel", QObject::tr("bevel", "modify - fillet")}},
+            {{"modfillet", QObject::tr("modfillet", "modify - fillet")},
+             {"fillet", QObject::tr("fillet", "modify - fillet")}},
+            {{"fi", QObject::tr("fi", "modify - fillet")}},
             RS2::ActionModifyRound
         },
         // divide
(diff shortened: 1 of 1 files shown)


librecad/src/cmd/lc_commandItems.h
```diff
@@ -758,26 +763,24 @@ const LC_CommandItem g_commandList[] = {
...
             {{"bev", QObject::tr("bev", "modify - bevel")},
              {"ch", QObject::tr("ch", "modify - bevel")}},
             RS2::ActionModifyBevel
         },
         // fillet
         {
-            {{"modfillet", QObject::tr("modfillet", "modify - fillet")}},
-            {{"fi", QObject::tr("fi", "modify - fillet")},
-             {"fillet", QObject::tr("fillet", "modify - fillet")},
-             {"bevel", QObject::tr("bevel", "modify - fillet")}},
```
pgmorgan, line 780: Not strictly necessary for this PR, but I have put fillet as a long form command name (which accepts a list per command) rather than a shortform alias.

librecad/src/cmd/lc_commandItems.h
```diff
@@ -758,26 +763,24 @@ const LC_CommandItem g_commandList[] = {
         },
         // offset
         {
-            {{"modoffset", QObject::tr("modoffset", "modify - offset")},
-             {"offset", QObject::tr("offset", "modify - offset")}
-         },
```
pgmorgan, line 763: Notably `offset` is now listed under `lineoff` and `parallel` command, no longer here.

pgmorgan (commented):
Note key changes.
These commands represent major commands I use regularly in CAD and which changed after v2.2.0.
I hope this PR will help revert the course of LibreCAD development so subsequent versions can benefit from the long history of CLI compatibility.

dxli:
I have no idea about the reasonings of the current naming.

I agree that the situation is a little messy with sudden changes of commands.

the issue is: can we avoid another sudden change by designing a naming to be most compatible between the old and new names.

for example, offset can be an alias of modoffset. For collisions, we can add an app option, or even a command message to avoid users from getting more confused 

my point is that adding is better than removing, if possible.

dxli (dismissed):
I feel it's good to keep old commands, unless a real conflict encountered.

sand1024:
@dxli 
Omg, I feel that this area with commands becomes to be a total mess... 

As I'll finalize and commit my current scope, I'll add UI editor for commands (something similar to management of shortcut) - it will let everyone to use onw favorite commands.

pgmorgan:
@sand1024 
@dxli 

Thanks for your comments.  Whether with @sand1024  suggestion or not, I still think the **default CLI** commands should **revert** to at least include the cross platform compatible CLI commands.  

The change away from platform agnostic CLI compatibility is very recent.  It happened between 2.2.0 and 2.2.1.

It is important to look at 2D CAD drafting as a skill, and not only a platform specific tool.  AutoCAD has 43 years of decent CLI stability.  LibreCAD has benefitted from sharing that stable CLI for 15 years.  This has allowed people to learn and apply this skills across tools and computers with little friction.  The LibreCAD community of users can benefit from returning to that common CLI as it facilitates access and application of the skill in a broader set of computational and institutional requirements at home, at school, at work, and across computers with greater or lesser compute resources.

As a critical case in the Linux community (which has a wide diversity of types of users), it is very common to access software in the package manager repositories and there are benefits to this approach even if it delays access to releases.  All current LTS Ubuntu versions and all current LTS Debian versions, including variant distros based on these, are still on `librecad` 2.2.0 or earlier in the `apt` repositories.  When such users who access `librecad` through their package managers arrive to version 2.2.1 or greater in the future, the `librecad` community would have long bifurcated along this critical and unnecessary interface difference unless we revert this change.

As an advocate for both `librecad` and FOSS I think this should not be overlooked, and should instead (with charity and without blaming) be treated as a _regression_.  I don't think it should be left to each user to manually re-align this common compatibility individually, especially given the reasonable delay in the user base reaching current releases and the benefits of a cross tool CLI.

Here are the changes I have identified to revert this:

```
incompatible command --> compatible command

lineoff    --> offset
modrotate  --> rotate
modscale   --> scale
modmirror  --> mirror
modtrim    --> trim
modoffset  --> offset
modbevel   --> bevel
modexplode --> explode
```
Other critical commands maintain good compatibility like `l`, `pl`, `c`, `move`, `mtext`, `dimlinear`, and many other common commands.  But the above list is the scope of the regression I have identified.  I hope that we can clean this up ourselves rather than leave it to each future user to handle or not.

Thank you kindly

dxli:
@pgmorgan 

This is the right direction to clean up. WIth this PR, the main commands are replaced by the more intuitive ones, i.e. more compatible ones, while all previous commands are expected to work as aliases, with the exception of "bevel"(a synonym of chamfer now).

The impact to general end users is expected to be small, as all previous commands would still work as before, with the only correction of "bevel".


Action | Main | Short (all still work)
-- | -- | --
rotate | rotate | modrotate, ro
scale | scale | modscale, sz
mirror | mirror | modmirror, mi
trim | trim | modtrim, tm
explode | explode | modexplode, xp
fillet | fillet | modfillet, fi
bevel | bevel, chamfer | modbevel, bev, ch
offset | offset | modoffset, mo, moff
parallel | parallel | linepar, lineoff, pa, ll

dxli (approved):
Re-approving after the offset/o rework and description update.

PR #2516, opened 2026-03-31
UI: Symbol to insert

PushKK:
1. ~~Block common signs.~~
Many signs we can insert from keyboard. It's fast and easy.
2. Add Euro sign.
3. Correct Russian.ts for work with "Symbol to insert".

### DIFF
--- a/librecad/src/ui/dialogs/entity/qg_dlgtext.ui
+++ b/librecad/src/ui/dialogs/entity/qg_dlgtext.ui
@@ -962,21 +962,24 @@
             <string>Plus / Minus (±)</string>
            </property>
           </item>
-          <item>
-           <property name="text">
-            <string>At (@)</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
-            <string>Hash (#)</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
-            <string>Dollar ($)</string>
-           </property>
-          </item>
+// We can use keyboard "Shift + 2"
+//          <item>
+//           <property name="text">
+//            <string>At (@)</string>
+//           </property>
+//          </item>
+// We can use keyboard "Shift + 3"
+//          <item>
+//           <property name="text">
+//            <string>Hash (#)</string>
+//           </property>
+//          </item>
+// We can use keyboard "Shift + 4"
+//          <item>
+//           <property name="text">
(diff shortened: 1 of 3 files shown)


librecad/src/ui/dialogs/entity/qg_dlgtext.ui
```diff
@@ -962,21 +962,24 @@
...
-           <property name="text">
-            <string>Hash (#)</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
-            <string>Dollar ($)</string>
-           </property>
-          </item>
+// We can use keyboard "Shift + 2"
```
dxli, line 965: There's a good reason to keep this.

For example, when the input is in Chinese typing mode, Shift+2 input the Chinese version of '@', which is actually different from the ASCII one.

dxli (changes requested):

librecad/src/ui/dialogs/entity/qg_dlgtext.ui
PushKK: I have Chinese-English keyboard and can check it tomorrow.

PushKK: I check it.
I use Dell keyboard Chinese-English and Windows 11 Pro 25H2 Eng.
Chinese keyboard layout use two mode - English mode and Chinese mode.
In English mode user can use Shift+4 and receive "$", in Chinese mode - "¥":
Screenshot 2026-04-01 092135
Screenshot 2026-04-01 092157

PushKK: I not know about PC in Chinese family or home but in office of IT-companies all PC have Chinese and English language in OS Windows installed.

In English-Russian languages you can see different in keys too.

PushKK: I need use GD&T signs and try insert it in code:
изображение


But for it need special font "ASME_GDT.ttf"
https://www.fontspace.com/category/asme,gdt

Or use "lc_open_gost_a regular" font as default (not unicode):
изображение

dxli (approved):

dxli:
To avoid merging trouble, I will take the Russian TS from your branch first.

Issue #1700, opened 2023-10-12
MacOS Dark Mode - LibreCAD is not really usable!


U--d--o:
### Expected behavior
A user interface where you can see the characters and icons even when running MacOS in dark mode

There are 2 possibilities, one after the other.

1. the simplest and fast in implementation:
Disable/ignore Dark Mode on MacOS for LibreCAD.

2. the hard way of customization, which can take a little longer after point 1 from me.

But please, so that you can use LibreCAD also on a Mac in Dark Mode, please implement point 1 asap!

### Observed behavior
[https://forum.librecad.org/Backround-color-tp5721547p5723650.html](https://forum.librecad.org/Backround-color-tp5721547p5723650.html

### Operating System and LibreCAD version info
MacOS
Sonoma 14.0 (23A344), 

LibreCAD
Version: 2.2.0.2
Compiler: Clang 13.0.0
Compiled on: Jul 29 2023
Qt Version: 5.15.10
Boost Version: 1.82.0
System: macOS 14.0

Issue #1606, opened 2023-04-14
Accessibility: icons are invisible in dark themes


Gering-ding-ding-ding-din:
### Prologue
Because of health problems regarding my eyes I am forced to use dark/black color themes exclusively combined with a screen brightness of <= 3%. Therefore I'm using a black color theme in KDE Plasma.

### Expected behavior
All icons should be visible.

### Observed behavior
Most icons are invisible.

### Steps to reproduce or sample file
Open LibreCAD in KDE Plasma with a dark theme (e.g. "Breeze Dark" or "Breeze Black") selected.

Breeze Black:
![Screenshot_20230415_013844_breezeblack](https://user-images.githubusercontent.com/22856448/232171756-76590566-c633-4014-922f-6eebbabd3254.png)

Breeze Dark:
![Screenshot_20230415_013844_breezedark](https://user-images.githubusercontent.com/22856448/232171927-dcc15905-6d84-4f00-ab35-6ae07bcb12fb.png)

### Operating System and LibreCAD version info
Operating System: Arch Linux 
KDE Plasma Version: 5.27.4
KDE Frameworks Version: 5.105.0
Qt Version: 5.15.9
Kernel Version: 6.2.10-arch1-1 (64-bit)
Graphics Platform: X11

LibreCAD Version: 2.2.0-alpha
Compiler: GNU GCC 12.2.0
Compiled on: Dec 17 2022
Qt Version: 5.15.9
Boost Version: 1.80.0
System: Arch Linux

gerardbm:
The green of the icons is the only part that have contrast ratio enough to be seen with a dark and a light background. So we only need tot create a version of this icons changing the black for white.

Librecad needs an option to choose between two icons themes, and I will do the icons for the dark mode if someone implements this option.

U--d--o:
Same Problem using Dark Mode on MacOS!

There are 2 possibilities, one after the other.

1. the simplest and fast in implementation:
Disable/ignore Dark Mode on MacOS for LibreCAD.

2. the hard way of customization, which can take a little longer after point 1 from me.

**But please, so that you can use LibreCAD also on a Mac in Dark Mode, please implement point 1 asap!**

flywire:
Quick and dirty conversion needs testing: https://github.com/flywire/LibreCAD/tree/dark-mode/librecad/res/icons/dark-mode

lordofbikes:
Many thanks for this! 👍 
[...]

Issue #1144, opened 2019-09-22
Poor visibility of UI when using dark UI theme


Zmuogs:
I using KDE Plasma 5.16.5 in Linux. When UI theme is dark (f.e. "Breeze Dark" in this case) then  most of icons, some labels in LibreCAD is poorly visible (black on dark background). See attached pictures.
 UI should adapt to dark UI theme.

![example_01](https://user-images.githubusercontent.com/20404470/65386478-e1f5ae80-dd44-11e9-8b02-f6b300bf9e92.png)
![example_02](https://user-images.githubusercontent.com/20404470/65386479-e1f5ae80-dd44-11e9-86b4-2c361ecdf289.png)
![example_03](https://user-images.githubusercontent.com/20404470/65386480-e1f5ae80-dd44-11e9-8b45-4d01a0b7de51.png)
![example_04](https://user-images.githubusercontent.com/20404470/65386481-e28e4500-dd44-11e9-9f1a-efda720e7766.png)

LibreCAD:
Version: 2.1.3
Compiler: GNU GCC 8.2.0
Compiled on: Sep 17 2018
Qt Version: 5.12.2
Boost Version: 1.67.0
System: Ubuntu 19.04

Zmuogs:
Note on black label "Command:" in attached screenshot: when LibreCAD is started fresh then that label is white, but mouseclick in edit field of that "Command:" ("Commant:" turns blus) and then hit ESC -> "Command:" turns black.

natrius:
Jumping in, its hard to see anything. I'm using "Starcraft" and, well...

![image](https://user-images.githubusercontent.com/7243447/73089545-c22eec80-3ed6-11ea-9fed-70778b86e811.png)

I tried to search for a setting to change the theme just for this program back to breeze default so i can work with it, but i could not find anything. I think another program had it, but unfortunately i cannot remember :(

MikePryadko:
I have the same situation:

![librecad_dark](https://user-images.githubusercontent.com/31248040/73748253-14d79680-476a-11ea-8f9a-1880090dc4a0.png)
```
Version: 2.1.2
Compiler: GNU GCC 6.2.0
Compiled on: Sep 17 2016
Qt Version: 5.9.5
Boost Version: 1.62.0
System: Ubuntu 18.04.4 LTS
```
My DE is Xfce (theme `adwaita-dark`), so I use `qt5-style-plugins` for applying GTK theme to Qt

[...]

Issue #364, opened 2013-03-06
Mac OS Window Size on Rotated Display


Bowler-Hat:
When running LibreCAD on a Mac with monitor that is rotated 90º, so that the display is taller than it is wide, the program forces its width to match the height of the monitor.  I can resize the window vertically, but the width is always wider than the display.  It appears to be acting as if the monitor is not rotated.  If I maximise the window it fits to the display properly.

Mac OS 10.7.5 on a Mac Mini with a Dell Monitor, 1200 x 1920

r-a-v-a-s:
I can't test this so I won't work on it;
however, I did find some information:
http://stackoverflow.com/a/25546487/1217270
http://doc.qt.io/qt-5/qsensor.html

r-a-v-a-s:
http://doc.qt.io/qt-5/qscreen.html#primaryOrientation-prop

PR #2520, opened 2026-04-04
UI: Add option to force light theme in settings

greenjava:
When the dark theme is enabled, some toolbar icons become difficult to see.
To provide a simple solution without having to redesign all icons for dark theme compatibility, this pull request introduces an option to force the application to use a light theme.

Changes introduced
- Added a new checkbox in the Settings panel that allows users to enforce a light color scheme, regardless of the operating system's theme.
- When enabled, the application switches to a light palette, improving visibility of toolbar icons that were designed for light backgrounds.
- A restart of the application is required for the change to take effect.

- Dark theme issue (icons are hard to see)

image

- Option to force the light theme

image

- Result after restarting the application

image

This could be a quick fix for the issues:
#1144
#364
#1606
#1700

### DIFF
--- a/librecad/src/main/main.cpp
+++ b/librecad/src/main/main.cpp
@@ -40,6 +40,7 @@
 
 #include <QDir>
 #include <QPushButton>
+#include <QStyleFactory>
 #include <QTimer>
 #include <QToolBar>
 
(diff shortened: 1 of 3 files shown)


dxli (approved):

sand1024:
@greenjava 
>To provide a simple solution without having to redesign all icons for dark theme compatibility

Well, actually icons redesign is not needed - it's ok just to change colors of icons, in Icon Style section of Widget Options (as described there https://github.com/LibreCAD/LibreCAD/pull/2078). 

Yet yes, forcing light theme may be useful indeed. 

@dxli 
I suppose this anyway rather partial solution - I think one day support of themes should be improved in general. I planned to do this, yet had no time so far.

sand1024:
no need to merge this. It's partial and will be replaced by my next PR with deeper support of themes.

PR #2523, opened 2026-04-07
Offset implementation proposal for ellipses

qwilvove:
Hello community,

After working with ellipses, I have noticed that the offset tool did not work for them.

I have implemented the offset function for ellipses, and I want to share this implementation since it could be useful for users.

For this implementation, I use getNearestPointOnEntity() to determine  the correct sign of the distance to apply.
I have noticed that we can observe jitters when trying to offset the ellipse with "Snap on entity" enabled. Sometimes, the projected result of the offset is drawn inside, and sometimes it is drawn outside of the initial ellipse.
This "issue" can be observed when offsetting other existing shapes as well. (So I suppose that it's not due to this implementation, and I don't even know if this is an issue).

Maybe offsetting an ellipse requires a different approach, and I am convinced that there are hundreds of ways to perform and optimise this functionality.

Regards,

qwilvove

### DIFF
--- a/librecad/src/lib/engine/rs_ellipse.cpp
+++ b/librecad/src/lib/engine/rs_ellipse.cpp
@@ -1220,6 +1220,40 @@ RS_Vector RS_Ellipse::dualLineTangentPoint(const RS_Vector& line) const
     return lineEqu(vp0) < lineEqu(vp1) ? vp0 : vp1;
 }
 
+/**
+  * this function creates offset
+  *@coord, position indicates the direction of offset
+  *@distance, distance of offset
+  * return true, if success, otherwise, false
+  *
+  *Author: Dongxu Li & qwilvove
+  */
+bool RS_Ellipse::offset(const RS_Vector& coord, const double& distance) {
+    double r0(coord.distanceTo(getCenter()));
+    double r1(r0);
+    double nearestRadius = getNearestPointOnEntity(coord).distanceTo(getCenter());
+    double majorRadius = getMajorRadius();
+    double minorRadius = getMinorRadius();
+    if(r0 > nearestRadius){
+        //external
+        r0 = majorRadius + fabs(distance);
+        r1 = minorRadius + fabs(distance);
+    }else{
+        r0 = majorRadius - fabs(distance);
+        r1 = minorRadius - fabs(distance);
+        if(std::min(r0, r1)<RS_TOLERANCE) {
+            return false;
+        }
+    }
+
+    setRatio(r1 / r0);
+    RS_Vector majorP = getMajorP();
+    majorP.scale(r0 / majorRadius);
+    setMajorP(majorP);
+    calculateBorders();
+    return true;
+}
+
 void RS_Ellipse::move(const RS_Vector& offset) {
     data.center.move(offset);
     //calculateEndpoints();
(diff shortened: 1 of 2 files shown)


dxli:
For offsetting of ellipses (or any other conics beyond lines or circles), the result is not a conic section anymore.

|Eccentricity |$a$ |$b$ |$d$ | Max deviation| % of $d$| Notes|
| -| -|- |- |- |- |- |
| Low | 5 | 4 | 0.5 | 0.056 | 11% | Nearly circular — very good |
| Typical | 5 | 3 | 0.5 | 0.126 | 25% | Noticeable but often acceptable |
| Moderate | 5 | 2 | 0.5 | ~0.220 | 44% | Clearly visible difference |
| High | 5 | 1 | 0.3 | 0.214 | 71% | Approximation breaks down |
| Large d  d  d | 10 | 9 | 0.5 | 0.026 | 5% | Still close when round |

I suggest we continue to implement as splines: number of control points should be good enough to keep the maximum deviation smaller than 1 pixel at the time of creation. Of course, user can change zoom factor later on, and we probably don't want to over engineer here.

dxli (changes requested):
As discussed:

1, for offset: we prefer spline approximation over ellipse approximation, because of limitations of ellipses;
2, It's possible to provide an approximated version of "ellipse offsetting", and we need to discuss this further.

qwilvove:
Thank you for those clarifications.

I am interested in trying to implement ellipse offset feature.

(Edit : after having examined the code in greater detail, I found answers for most of the following questions, in italic)

I am wondering during which step of the process the resulting shape of the offset (LC_SplinePoints, maybe RS_Spline ?) should be created (directly inside RS_Ellipse::offset() function ?, RS_Modification::offset() ? or maybe elsewhere ?)

_Does the code already contain an example of such an operation (where the resulting entity type of a modification is different from the initial shape) ?_
(Edit : The tool to create a circle from a arc seems to be an example of this.)

_Regarding the possibility of providing an approximated version of ellipse offsetting keeping the ellipse type for the resulting shape, maybe this could be set in Ui_ModifyOffsetOptions widget. We could ask the user to chose between two values in a combo box, or maybe more simple and clear : using a check box (unchecked by default) "force resulting shape to be an ellipse (inaccurate offset)"._
(Edit : I am not sure of this anymore, since in this case the option will be visible for other type of entities as well)

Thank you in advance for your feedback and your help.

Best regards,

qwilvove

qwilvove:
For the implementation of the ellipse offset, I'm currently looking into the following approach :

- In RS_Modification::offset(), check if the entity is an ellipse. If so, create a new LC_SplinePoints.
- Call a new function on this spline ( defineFromEllipse(RS_Ellipse *ellipse, int numberOfPoints) ) to add points on the spline to approximate initial ellipse shape.
- Offset the spline.

dxli:
@qwilvove 

I am thinking about two ways:

1, as you suggested we either create the spline directly, or approximate the ellipse as a spline and delegate the offset creation to spline, as you suggested;
2, we create a new entity type of ellipse offset, handle it up to precision: length, intersection, bounding box, rendering, etc; (the renderer will use a spline internally, and ensure the max deviation is less than 1 pixel, similar to what we do with hyperbolas).

Clearly, route 2 will have to solve lots of math tasks, and significant amount of implementation.

sand1024:
@dxli 

> 2, we create a new entity type of ellipse offset,  

Hm... ok, but what about DXF format compatibility with this approach? 

P.S. I still not sure that approach with spline approximation is reasonable to force it. It slows down rendering significantly, with quite doubtful benefits comparing to linear approximation.

dxli:
PR #2545 implemented ellipse offset by LC_SplinePoints. Closing this one.

PR #2535, opened 2026-04-22
Fix: Replace setDevicePixelRatioF() function with setDevicePixelRatio() for Qt6 compatibility

abhinanda303:
setDevicePixelRatioF() is causing build errors in some Qt6 environments and should be reverted to setDevicePixelRatio() particularly on the file librecad/src/lib/gui/render/widget/lc_widgetviewportrenderer.cpp


dxli (approved):
This is a fixed needed urgently.

thanks!

PR #2567, opened 2026-05-16
Win32 registry and pen palette fixes

dxli:
1, Win32 dxf file association fixes in registry;
2, Pen Palette options should be saved;
3, Updated TS Files (may be separated into its own PR before merging).


librecad/src/ui/lc_penpalettewidget.cpp
```diff
@@ -401,17 +401,9 @@ void LC_PenPaletteWidget::invokeOptionsDialog(bool focusOnFile){
...
-                // yet this is also quite a messy way... So let's ask for restart.
-                QMessageBox::warning( this, tr("Pen palette"),
-                                      tr("Location of pens file is changed, please restart the application so new pens file will be used.\n\n"
-                                         "Please note that if you'll save pen via editor without restart, current pens from palette will be saved "
-                                         "in the new file and therefore existing content of it will be overridden."),
-                                      QMessageBox::Ok,
-                                      Qt::NoButton);
+                // Save current pens to the new file immediately so no restart is needed
+                // and the user's palette is preserved at the new location.
+                penPaletteData->saveItems();
```
sand1024: @dxli 
In general yes, but what if file specified as new location already contains data? 
With such approach, they will be just **overridden silently**, which is also not too good. 
Actually, I suppose that was the reason why I've asked the user to restart in the original code...

dxli: @sand1024 

The options should be in RS_Settings. Still need to be concistent between 2.2.1 and the master branch though.

The PR is not finalized yet.

sand1024:
@dxli 
Well, of course such things like pen's setting (and other ones) may be stored in settings. 

However, I'm really not sure this is right direction - moreover, I suppose as much shared settings as possible should live in external files (or, at least,  there should be possibility to export/import them). 

The reason for this is quite straightforward - at the moment, there is a minor exchange of various related assets within the users community. 

Letting the users to create and share some specific settings (like pens, or color schemes for icons or for UI), or menus configurations, or commands mapping and so on) - may be a really good step in usability and community support.  Plus - it will be possible to include various pre-defined configurations as part of installation (imagine that on first start the user may select, say, dark/light theme and later select color scheme for the drawing, plus some UI configuration options).

PR #2572, opened 2026-05-19
DWG support: round 3

dxli:
Improve support for DWG versions up to 2017 (AC1027).


sand1024:
@dxli 
Hi! 

I've noticed you do improvements in supporting formats - that's really great!

I think as you do development in that area, probably you may also add support for storing/reading arc dimensions data?

As for now, there is quite a strange situation - there is an action and entity (`LC_DimArc`) to draw arc dimensions. 

However, original author didn't create the code for storing and reading `LC_DimArcData `to DXF or DWG. 

Thus, `LC_DimArc` may be created during editing session, yet after reopening the document - it's just lost - and that's weird that there is no error for this in tracker.

dxli:
@sand1024 ,

Added support for LC_DimArc for DWG and DXF.

sand1024:
@dxli 
Oh, that's great - thank you!
As you'll merge this PR, I can try to test it using files created by other CADs, if that will be helpful.

librecad/src/lib/actions/rs_snapper.cpp
```diff
@@ -1092,8 +1092,12 @@ RS_Vector RS_Snapper::doSnapToAngle(const RS_Vector &currentCoord, const RS_Vect
 }
 
 RS_Vector RS_Snapper::toGraph(const QMouseEvent* e) const{
-     const QPointF &pointF = e->position();
-     RS_Vector result = m_viewport->toWorldFromUi(pointF.x(), pointF.y());
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```
sand1024, line 1095: @dxli 
I'm just curious - what for did you add this check? 
I suppose master is for QT6 anyway, so its not necessary to support Qt 5 there? Or do I miss something? Please advice.

librecad/src/ui/dock_widgets/views_list/lc_namedviewslistwidget.h
```diff
@@ -23,6 +23,7 @@
 #ifndef LC_NAMEDVIEWSLISTWIDGET_H
 #define LC_NAMEDVIEWSLISTWIDGET_H
 
+#include <QIcon>
```
sand1024, line 26: @dxli 
I noticed that in several files you've added includes that are redundant. Did you experience some issues with compiling or other tools?  I'm asking since previously these files were compiled fine both locally by MSVC and GCC and using continuous integration - so just curious what is the purpose of such includes (in general, I prefer to clean such things, but probably there is some specific reason I'm not aware of).

PR #2586, opened 2026-05-28
Add Direct Line action for typed-length line drawing

pcfixindude:
## Summary

This PR adds a new line drawing action named **Direct Line** (`LC_ActionDrawLineDirect`).

Direct Line supports a direction-based typed-length workflow — type a distance, aim with the mouse, press Enter. The tool continues automatically from each new endpoint, with support for architectural distance input, polar snap, and several multi-entity drawing macros (opening, window, door).

The new action does not replace or modify the existing Snake Line action.

## Workflow

1. Activate Direct Line (`dline` or `dl`).
2. Click a start point.
3. Type a distance.
4. Move the mouse to aim the direction.
5. Press Enter — the segment is placed and the tool continues from the new endpoint.
6. Right-click or Escape finishes.

### Architectural feet-inch input

When enabled in **Application Preferences → Defaults**, distance input accepts feet-inch strings:

- `10-3` → 10 ft 3 in
- `3'0` → 3 ft 0 in
- `10-3 7/8` → 10 ft 3⅞ in

Falls back to standard numeric/expression input (`RS_Math::eval`) when disabled or when the string is not a feet-inch format.

### Polar Snap

- A toolbar toggle (`snapAngle`) activates angle-increment snapping globally during Direct Line aiming.
- Holding **Shift** temporarily enables it at the configured angle.
- Object snaps always take priority over polar snap (guarded by a proximity check against the raw mouse position).
- Angle increment is configured in Preferences (default 15°).

Implemented by overriding `doGetMouseSnapPoint` in `LC_ActionDrawLineDirect`.

### Wall opening mode (`o` / `opening`)

Available after at least one wall segment has been drawn. The opening direction is locked to the previous wall segment angle.

1. Type `o` or `opening`.
2. Enter the opening width (all distance formats supported).
3. Click/aim to choose which side of the wall the markers appear on.

Draws **three lines** as one undoable operation:

- perpendicular marker at the opening start
- span along the wall direction
- perpendicular marker at the opening end

Active start point advances to the far side of the opening. Marker depth is configurable in Preferences (default 4).

### Window mode (`w` / `window`)

Available whenever a valid start point exists. The window direction comes from the mouse aim (not wall history).

1. Type `w` or `window`.
2. Enter the window width.
3. Click/aim to set the window direction.

Draws **five lines** as one undoable operation:

- center line
- two parallel offset lines
- closing cap at the near end
- closing cap at the far end

Active start point advances to the far side of the window. Offset width is configurable in Preferences (default 1.5).

### Door mode (`d` / `door`)

Available after at least one wall segment has been drawn. The door direction is locked to the previous wall segment angle.

1. Type `d` or `door`.
2. Enter the door width.
3. Click/aim to choose the swing side.
4. Click near the desired hinge end to choose hinge placement.

Draws **two lines and one arc** as one undoable operation:

- span line along the wall direction
- perpendicular door leaf line at the hinge point
- quarter-circle swing arc from the leaf endpoint to the opposite corner of the opening

Active start point advances to the far side of the door.

## Preferences added

Four new settings in **Application Preferences → Defaults**:

| Setting | Key | Default | Description |
|---|---|---|---|
| Architectural feet-inch input | `/Defaults/ArchitecturalInput` | off | Enables feet-inch distance parsing |
| Polar snap angle | `/Defaults/PolarSnapAngle` | 15 | Angle increment in degrees |
| Opening marker depth | `/Defaults/OpeningDepth` | 4 | Perpendicular marker length for opening mode |
| Window offset width | `/Defaults/WindowOffsetWidth` | 1.5 | Offset from center line for window mode |

## Implementation notes

- All drawing macros (opening, window, door) use the existing `performTrigger` / `doPrepareTriggerEntities` path, placing all entities in one undo cycle.
- Lines are held in `QList<RS_LineData> m_pendingOpening`; arcs in `QList<RS_ArcData> m_pendingArcs` (used by door mode).
- Door swing arc uses `RS_Arc` / `RS_ArcData`; the `reversed` flag is derived analytically to always produce a 90° quarter-circle for all four hinge/swing combinations.
- Direct Line's internal history (`pPoints->history`) is extended with `HA_Opening`, `HA_Window`, and `HA_Door` entries so the action's own undo/redo correctly restores the active start point.
- `doGetMouseSnapPoint` is overridden to apply polar snap only in free-snap states (no object snap hit), and is suppressed during aim states where only side/hinge selection is needed.

## Documentation

A **Direct Line** section has been added to `ACTIONS_README.md`, documenting all commands, sub-workflows, and Preferences settings.

## Commands added

| Command | Aliases | Notes |
|---|---|---|
| `dline` / `dl` | — | Invocation |
| `o` / `opening` | — | Wall opening mode |
| `w` / `window` | — | Window mode |
| `d` / `door` | — | Door mode |
| `close` | — | Close sequence to start point |
| `undo` | — | Undo last segment/macro |
| `redo` | — | Redo |
| `pl` | `polyline` | Convert segments to polyline |
| `start` | — | Set new start point |

## Testing performed

Built and launched successfully on macOS using Homebrew Qt 5:

```bash
cmake --build /Users/irds/Projects/librecad-custom/build --target librecad -j$(sysctl -n hw.logicalcpu)
open /Users/irds/Projects/librecad-custom/build/librecad.app
```


sand1024:
@pcfixindude

hm... based on the code, it seems that you've originally started your working not by pulling master branch but with some older one, right? If it is so, I'm not sure it's reasonable to push into the master, probably into one of 2.2.1.x branches - the difference in codebase is too significant there.  
Plus your PR includes lot's of commits that are already in master... 

And a small note regarding the implementation - I suspect that your version of the LC_ActionDrawLineDirect class action is just a modified version of LC_ActionDrawLineSnake class - and thus it contains duplicated code. 

If it is so, I think it's better to rely on classical inheritance rather than on copy/paste approach, so the result will be cleaner and easier to maintain.

dxli (changes requested):
as @sand1024 noticed, need to rebase on the master branch

sand1024:
@dxli
I'm afraid for master it will be hard for him to rebase to master... 
It should be possible to merge this action into 2.2.1 branch and use the action till 2.2.2 will be stable - but for me that functionality will be not needed for 2.2.2 - I suppose I cover it already there: 
Relative Point Input Assistant - fast coordinate entry - https://youtu.be/oyI51XT704k;
So if it will be merged in 2.2.1 - some users may use to use it and this action should be supported later as a legacy one ...  

Thus is is hard for me to say what is right direction there..

sand1024:
not applicable to master branch

Issue #2578, opened 2026-05-22
Block factor scale


beppe61:
When inserting a block from the library multiple times, changing the scale factor after the first one, subsequent insertions do not reflect the new scale factor.
Thanks

Version: v2.2.1.5
Compiler: Microsoft Visual C++
Compiled on: May  2 2026
Qt Version: 5.15.2
Boost Version: 1.87.0
System: Windows 10 Version 2009

PR #2606, opened 2026-06-08
[OLD] Fix proposal for #2578 - Library block factor scale

qwilvove:
Hello,

Here is a fix proposal for #2578.

The idea here is to drop the use of RS_Modification::paste(), and use another process instead : 

1. create a block which contains all entities from the library item.
2. add the block to the document if it does not already exist.
3. insert the block with properties based on the values entered by the user.

This code still contains some points that would benefit from clarification (which are present in current code) :

- If a library item contains blocks, they are skipped. (Is that the intended behaviour ? Should those blocks be decomposed into atomic entities before being inserted as parts of the main block ? Maybe trying to preserve the block/insert structure of the library item during the insert is another idea, but this would result in the creation of numerous blocks within the document, which might not suit the user needs.)
- If the operation is undone afterwards, the inserted block remains in the document block list (which could be complicated to manage since the block may exist in the document before the insert attempt).
- If a block with a same name as the library item already exist in the document, the operation results in inserting this block instead.

I point this PR to 2.2.1. Maybe pointing it to master is the way to go. What is the current dev policy between master and 2.2.1 ?

P.S. : I drop the use of RS_Units::convert() in this proposal, but maybe it is needed to keep it here.

I hope this proposal will help us find ways to resolve the initial issue.

Thanks in advance.

Best regards,

qwlvove

### DIFF
--- a/librecad/src/lib/creation/rs_creation.cpp
+++ b/librecad/src/lib/creation/rs_creation.cpp
@@ -929,29 +929,49 @@ RS_Insert* RS_Creation::createLibraryInsert(RS_LibraryInsertData& data) {
     RS_Graphic g;
     if (!g.open(data.file, RS2::FormatUnknown)) {
         RS_DEBUG->print(RS_Debug::D_WARNING,
-                        "RS_Creation::createLibraryInsert: Cannot open file: %s", data.file.toStdString().c_str());
+                        "RS_Creation::createLibraryInsert: Cannot open file: %s",
+                        data.file.toStdString().c_str());
         return nullptr;
     }
 
-    // unit conversion:
-    if (graphic) {
-        double uf = RS_Units::convert(1.0, g.getUnit(),
-                                      graphic->getUnit());
-        g.scale(RS_Vector(0.0, 0.0), RS_Vector(uf, uf));
+    QString blockName = QFileInfo(data.file).completeBaseName();
+    RS_BlockData blockData = RS_BlockData(blockName, RS_Vector(0.0, 0.0), false);
+
+    // Add all entities to a block, except insert entities
+    RS_Block* block = new RS_Block(container, blockData);
+    int numberOfEntities = 0;
+    for(auto e: g.getEntityList()) {
+        if (e) {
+            RS_Entity* c = e->clone();
+            if (c->rtti() != RS2::EntityInsert) {
+                block->addEntity(c);
+                numberOfEntities++;
+            }
+        }
+    }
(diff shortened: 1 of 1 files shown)


dxli:
@qwilvove 

This should go to the master branch, since it's more than just a direct bug fix. It's conceivable to backport it to 2.2.1 in the future.

librecad/src/lib/creation/rs_creation.cpp
```diff
@@ -929,29 +929,49 @@ RS_Insert* RS_Creation::createLibraryInsert(RS_LibraryInsertData& data) {
...
+    QString blockName = QFileInfo(data.file).completeBaseName();
+    RS_BlockData blockData = RS_BlockData(blockName, RS_Vector(0.0, 0.0), false);
 
-    //g.scale(RS_Vector(data.factor, data.factor));
-    //g.rotate(data.angle);
+    // Add all entities to a block, except insert entities
+    RS_Block* block = new RS_Block(container, blockData);
+    int numberOfEntities = 0;
+    for(auto e: g.getEntityList()) {
+        if (e) {
```
dxli, line 944: just some trivial coding style: we prefer (nullptr != e) over bool(e): just choose being explicit over implicit conversions.

librecad/src/lib/creation/rs_creation.cpp
```diff
@@ -929,29 +929,49 @@ RS_Insert* RS_Creation::createLibraryInsert(RS_LibraryInsertData& data) {
...
-                                      graphic->getUnit());
-        g.scale(RS_Vector(0.0, 0.0), RS_Vector(uf, uf));
-    }
+    QString blockName = QFileInfo(data.file).completeBaseName();
+    RS_BlockData blockData = RS_BlockData(blockName, RS_Vector(0.0, 0.0), false);
 
-    //g.scale(RS_Vector(data.factor, data.factor));
-    //g.rotate(data.angle);
+    // Add all entities to a block, except insert entities
+    RS_Block* block = new RS_Block(container, blockData);
```
dxli, line 941: coding style: try to avoid raw new/delete, and use smart pointers for new code.
Also, the RAII style

dxli (changes requested):
wondering whether it's better to create a new helper method, to avoid either:'

1, directly call paste();
2, duplicate the logic in paste().

Also, it helps, if we can clean up the logic in paste() as well.

qwilvove:
Thank you for those advices. I've created a new PR to master with a simpler fix -> #2624.

PR #2620, opened 2026-06-12
dxf2png: add missing RS_FONTLIST->init() call

raffa0001:
Text entities were not rendered in dxf2png CLI export because RS_FONTLIST->init() was never called, leaving the font list empty when openDocAndSetGraphic() processes the DXF file.

Also fixes absolute output path handling with -o flag.


dxli (approved):
Thanks!

PR #2624, opened 2026-06-15
Fix proposal for #2578 - Library insert factor scale

qwilvove:
Hello,

Here is a fix proposal for #2578.

This PR is a new version of my previous PR #2606.
This new PR targets master and is a much simpler fix.

In fact, in RS_Modification::paste, RS_PasteData properties (factor and rotation) were mixed in both block creation and insert definition.
With this fix, the block is created with its own scale and rotation, the insert holds the intended scale and rotation properties set by the user.

I've also added a commit to fix indent in both RS_Creation::createLibraryInsert & RS_Modification::paste functions.

Best regard,

qwilvove


dxli:
@qwilvove 

Thanks a lot for the troubleshooting, and the PR is a solid fix.

I added a little bit more on top of the PR and pushed it directly to 2.2.1

dxli:
manually merged. Closing now.

PR #2646, opened 2026-06-21
DWG support: round 5

dxli:
1, more pre-R13 DWG support;
2, fixed new drawing dwg: overwritting and open printing issues;
3, avoid "Cannot check for updates" message boxes.

### DIFF
--- a/librecad/src/lib/engine/document/container/lc_pathbuilder.cpp
+++ b/librecad/src/lib/engine/document/container/lc_pathbuilder.cpp
@@ -36,6 +36,38 @@
 #include "rs_math.h"
 #include "rs_debug.h"
 
+#include <cmath>
+
+namespace {
+
+bool hasFiniteValue(double value) {
+  return std::isfinite(value);
+}
+
+bool hasUsableVector(const RS_Vector& vector) {
+  return vector.valid
+      && hasFiniteValue(vector.x)
+      && hasFiniteValue(vector.y)
+      && hasFiniteValue(vector.z);
+}
+
+bool hasUsablePoint(const QPointF& point) {
+  return hasFiniteValue(point.x()) && hasFiniteValue(point.y());
+}
+
+bool hasUsableArcParameters(double x, double y, double width, double height, double startAngle, double sweepAngle) {
+  return hasFiniteValue(x)
+      && hasFiniteValue(y)
+      && hasFiniteValue(width)
+      && hasFiniteValue(height)
+      && width > 0.
+      && height > 0.
+      && hasFiniteValue(startAngle)
+      && hasFiniteValue(sweepAngle);
+}
+
+}
+
 namespace LC_LoopUtils {
 
 PathBuilder::PathBuilder(RS_Painter* painter)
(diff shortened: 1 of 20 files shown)


librecad/src/lib/engine/document/entities/rs_arc.cpp
```diff
@@ -985,14 +994,32 @@ void RS_Arc::createPainterPath(RS_Painter* painter, QPainterPath& path) const {
 }
 
 void RS_Arc::draw(RS_Painter* painter) {
+  if (painter == nullptr
```
sand1024, line 997: @dxli 
Hm.. are you sure that this is right approach? 
This check will be performed on each drawing - if it's really necessary, probably it's better to add separate flag and calculate it once after updating entity?  
As the process of loading and editing entities is fully localizable - it's possible to add something like "validate()" method to RS_Entity, call it from appropriate places (like initial loading and somewhere in undo cycle or so) - and later on just check for the flag only. 

Actually, that's looks like a huge overhead and not efficient approach - the method will be called lots of times and will always return the same result of validation (until the entity will be changed - yet actually that will be a clone, even not the same entity). 

Please consider this - the rendering pipeline is the critical performance path.

librecad/src/lib/engine/document/container/lc_pathbuilder.cpp
```diff
@@ -124,37 +162,112 @@ QPointF PathBuilder::toGuiPoint(const RS_Vector& vp) const {
...
   double startAngleDeg = m_painter->toUCSAngleDegrees(arc->getData().startAngleDegrees);
   double angularLength = arc->getData().angularLength;
+  if (!hasUsableArcParameters(minCorner.x, minCorner.y, uiSize.x, uiSize.y, startAngleDeg, angularLength))
+    return false;
+
+  if (!appendStartPoint(arc->getStartpoint()))
+    return false;
+
   // arcTo without arcMoveTo connects from current position without starting a new subpath
   m_path.arcTo(minCorner.x, minCorner.y, uiSize.x, uiSize.y, startAngleDeg, angularLength);
```
sand1024, line 232: @dxli 
I'm not 100% sure, yet arcTo() may bring us the same issue with Qt-drawing arcs that we've faced some time ago - un-exact rendering on edges with some resolutions. 

Please check this. 

P.S Actually, I think there is a faster and simple approach for the rendering and paths processing, something like using linear approximation of entities (it is already present for most of them) and build paths and clipping based on lines rather than on curves. 
That should be definitely fast, it will later let port rendering to OpenGL (which is good with lines and triangles) and also may deliver quite interesting features - say, using 
Clipper - https://github.com/AngusJohnson/Clipper2 or (https://www.angusj.com/clipper2/Docs/Overview.htm) - it will be possible no only do efficient clip parts to avoid redundant renderings, but also do things like finding contours or so. 

So the pipeline may be something line standard pipeline: 
defining entities to paint -> approximation -> clipping lines -> rendering. The entire render will just set pen properties and render lines..

PR #2664, opened 2026-07-02
Avoid stale layer tree item use while editing layers

alpine-algo:
## Summary
- Keep stable original layer data in LC_LayerDialogEx instead of relying on a raw LC_LayerTreeItem pointer remaining valid while the dialog is open.
- Re-resolve the edited tree item from the stable RS_Layer pointer after the dialog is accepted before applying rename/type changes.
- Fail gracefully if the layer disappears while the dialog is open.

## Why
The layer edit dialog stores an LC_LayerTreeItem pointer. That item can become stale if the layer tree is rebuilt while the dialog is open, for example while changing layer type or construction/informational attributes. Validation then reads through the stale item and can crash before changes are applied.

## Tests
- `git diff --check origin/master..HEAD`
- `qmake6 -r <checkout>/librecad.pro`
- `make -j24`
- `<build>/unix/librecad -h`
- Manual local reproduction flow on a patched LibreCAD build: edit a reference layer, change construction/informational state and line type, accept the dialog. The pre-patch build crashed in LC_LayerDialogEx::validate; the patched build no longer crashed.

### DIFF
--- a/librecad/src/ui/dock_widgets/layers_tree/lc_layerdialog_ex.cpp
+++ b/librecad/src/ui/dock_widgets/layers_tree/lc_layerdialog_ex.cpp
@@ -41,6 +41,11 @@ LC_LayerDialogEx::LC_LayerDialogEx(QWidget* parent, const QString& name, LC_Laye
     m_layerTreeModel = model;
     m_layerList = ll;
     m_editedTreeItem = treeItem;
+    if (treeItem != nullptr) {
+        m_editedLayer = treeItem->getLayer();
+        m_originalLayerName = treeItem->getName();
+        m_originalLayerType = treeItem->getLayerType();
+    }
 }
 
 void LC_LayerDialogEx::languageChange(){
@@ -250,14 +255,20 @@ void LC_LayerDialogEx::validate() {
                 break;
             }
             case MODE_EDIT_LAYER: {
-                QString originalName = m_editedTreeItem->getName();
-                int originalLayerType = m_editedTreeItem->getLayerType();
-                bool typeChanged = newLayerType != originalLayerType;
-                bool nameChanged = originalName != layerName;
+                LC_LayerTreeItem *editedTreeItem = m_layerTreeModel->getItemForLayer(m_editedLayer);
+                if (editedTreeItem == nullptr) {
+                    QMessageBox::information(parentWidget(),
+                                             QMessageBox::tr("Layer Properties"),
+                                             QMessageBox::tr("The edited layer is no longer available."),
+                                             QMessageBox::Ok);
+                    return;
+                }
+                bool typeChanged = newLayerType != m_originalLayerType;
+                bool nameChanged = m_originalLayerName != layerName;
                 if (nameChanged || typeChanged) {
                     // inner name should be changed - so we have to check for possible duplicates.
                     newLayerNamesList = m_layerTreeModel->getLayersListForRenamedPrimary(
-                        m_editedTreeItem, layerName, newLayerType);
+                        editedTreeItem, layerName, newLayerType);
                 }
                 break;
             }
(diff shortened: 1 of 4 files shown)


librecad/src/ui/dock_widgets/layers_tree/lc_layerdialog_ex.h
```diff
@@ -74,6 +74,9 @@ protected slots:
     int m_mode;
     LC_LayerTreeModel* m_layerTreeModel;
     LC_LayerTreeItem *m_editedTreeItem {nullptr};
+    RS_Layer *m_editedLayer {nullptr};
+    QString m_originalLayerName;
+    int m_originalLayerType {-1};
```
dxli: can we use the enum type  LC_LayerTreeItem::NOT_DEFINED_LAYER_TYPE here?

librecad/src/ui/dock_widgets/layers_tree/lc_layerdialog_ex.cpp
```diff
@@ -41,6 +41,11 @@ LC_LayerDialogEx::LC_LayerDialogEx(QWidget* parent, const QString& name, LC_Laye
     m_layerTreeModel = model;
     m_layerList = ll;
     m_editedTreeItem = treeItem;
+    if (treeItem != nullptr) {
+        m_editedLayer = treeItem->getLayer();
```
dxli, line 45: better to check for null layer here

dxli (dismissed):

librecad/src/ui/dock_widgets/layers_tree/lc_layerdialog_ex.h
alpine-algo: Addressed in 6b66d15 by using LC_LayerTreeItem::NOT_DEFINED_LAYER_TYPE for the member initializer.

librecad/src/ui/dock_widgets/layers_tree/lc_layerdialog_ex.cpp
alpine-algo: Addressed in 6b66d15 by only copying original layer state after treeItem->getLayer() is non-null.

alpine-algo:
Thanks @dxli, that makes sense. Updated in 394213c so the stored original layer type uses LC_LayerTreeItem::LayerType, with explicit casts at existing int API boundaries.

dxli (approved):

sand1024:
@alpine-algo 
@dxli 

Introducing LC_LayerTreeItem::LayerType is not a good idea there and may leads to bugs. 

I suppose it is actually duplicates RS_Layer::LayerType. So, while new enum is introduced there - 
some code is still relied to the old one, like code in QG_LayerTreeItemAcceptorSecondary below. 

The correct fix if new enum is introduced it to correct references to it too - or, reference/extend existing one.

Otherwise, we have just have duplicate code and all that resulting mess.. I'll clean that up, yet next time please try to be more precise with the existing codebase and avoid not needed duplications. 

```
bool acceptLayerTreeItem(LC_LayerTreeItem* item) const override{
        const int ltype = item -> getLayerType();
        bool result = true;

        if (m_includeMode){
            result = (ltype == RS_Layer::LayerType::DIMENSIONAL) || (ltype == RS_Layer::LayerType::ALTERNATE_POSITION) || (ltype == RS_Layer::LayerType::INFORMATIONAL);
        }
        else{
            if (item -> getPrimaryItem() != nullptr){
                result = (ltype != RS_Layer::LayerType::DIMENSIONAL) && (ltype != RS_Layer::LayerType::ALTERNATE_POSITION) && (ltype != RS_Layer::LayerType::INFORMATIONAL);
            }
        }
        return result;
    }
```

alpine-algo:
@sand1024 Thanks for the follow-up.

The intent of the PR was only to stop holding a raw tree-item pointer across rebuilds (the crash path). The later change to a named `LC_LayerTreeItem::LayerType` member was type polish on top of that, not required for the fix. Fair to revisit for consistency with the rest of the layer-type usage.

Issue #2013, opened 2025-01-12
"Cancel"-button closes drawing without saving instead of aborting the close operation


ElDonRicko:
When attempting to close a drawing without previously saving and selecting "Cancel" or "X" (Top-right corner) the drawing gets closed without being saved.

### Expected behavior
The Close operation should be stopped and the Drawing should remain open when selecting "Cancel" or "X"

### Observed behavior
The Drawing gets closed without being saved, as if selecting "Discard".

### Steps to reproduce or sample file
Close any Drawing without saving and choose "Cancel" or "X"

### Video
https://github.com/user-attachments/assets/2ad9340f-7c89-4cab-857b-8ba70a4f78d9

### Operating System and LibreCAD version info
Version: 2.2.1_rc4
Compiler: GNU GCC 14.2.1
Erstellt am: Jan  7 2025
Qt Version: 5.15.16
Boost Version: 1.86.0
System: Version: 2.2.1_rc4
Compiler: GNU GCC 14.2.1
Erstellt am: Jan  7 2025
Qt Version: 5.15.16
Boost Version: 1.86.0
System: Arch Linux x86_64

sand1024:
@ElDonRicko 
Could you please check this in 2.2.2 alpha branch? The logic for closing is changed there.

ElDonRicko:
@sand1024 
Everything now works as expected, except when I try to close only the tab of the current drawing.
"Yes" and "No" work correctly, but selecting "Cancel" or clicking the "X" (to close the pop-up) creates a new temporary save file named "#Name.dxf".
If I then attempt to close and cancel again, the application saves "#Name.dxf" as "Name.dxf" and deletes the temporary "#Name.dxf" file.

When using the top-right "X" button to close the entire application, "Cancel" and "X" on the pop-up close the pop-up correctly every time, which is the expected behavior.

I hope my description is understandable, as I am not a programmer.

sand1024:
@ElDonRicko 
Oh, thank you for the detailed description. Yes, it seems that the is some glitch with "cancel" logic. 

As for #name.dxf file - it's a backup file, it's created from time to time as well as auto-save version of the file. Yet anyway, I'll check again.

PR #2666, opened 2026-07-03
Fix proposal for #2013 (fix cancel action on MDIWindow close event)

qwilvove:
Hello,

This is a fix proposal for #2013.

I hope that this is a good fix for this very dangerous issue, which currently affects stable release.

Best regards,

qwilvove

### DIFF
--- a/librecad/src/main/qc_mdiwindow.cpp
+++ b/librecad/src/main/qc_mdiwindow.cpp
@@ -253,6 +253,7 @@ void QC_MDIWindow::closeEvent(QCloseEvent* ce) {
             break;
         case QG_ExitDialog::Cancel:
             cancel = true;
+            break;
         default:
             break;
         }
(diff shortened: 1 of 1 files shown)


librecad/src/main/qc_mdiwindow.cpp
```diff
@@ -252,6 +252,7 @@ void QC_MDIWindow::closeEvent(QCloseEvent* ce) {
             cancel = false;
             break;
         case QG_ExitDialog::Cancel:
+            cancel = true;
```
dxli, line 255: better to explicitly add "break;", or [[fallthrough]]

dxli (approved):

qwilvove:
Thank you for your feedback, I've added a commit to the PR to add the missing "break;" line.

qwilvove:
I wonder if the fix would be better with:

```
        case QG_ExitDialog::Cancel:
        default:
            cancel = true;
            break;
```

than my current proposal:

```
        case QG_ExitDialog::Cancel:
            cancel = true;
            break;
        default:
            break;
```

If default is reached, the window is closed without saving (like the current cancel action behaviour), cancelling the operation instead may be safer.

PR #2675, opened 2026-07-09
Fix spline close command in current action

ideocentric:
## Summary
- Port the spline `close` command to the current `LC_ActionDrawSpline` implementation.
- Create a valid wrapped closed spline without changing the next spline closed preference.
- Keep an under-populated spline active on a premature right-click.
- Add a focused action-level regression test.

## Verification
- `QT_QPA_PLATFORM=offscreen ./build-pr2675/librecad_spline_command_tests "[spline][command]"`
- qmake6 configure followed by `make -j6`
- `QT_QPA_PLATFORM=offscreen ./LibreCAD.app/Contents/MacOS/LibreCAD --help`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,9 +91,30 @@ scripts/PWD
 
 # Mac store files
 .DS_Store
+.DS_Store?
+.AppleDouble
+.LSOverride
+Icon
+
+# macOS thumbnail/resource forks and metadata
+._*
+.DocumentRevisions-V100
+.fseventsd
+.Spotlight-V100
+.TemporaryItems
+.Trashes
+.VolumeIcon.icns
+.com.apple.timemachine.donotpresent
+
+# macOS directories that might appear on network/AFP volumes
+.AppleDB
+.AppleDesktop
+Network Trash Folder
+Temporary Items
+.apdisk
 
 #pkg files
-*.pkg
+
 #vim files
 *.swp
 *.orig
@@ -120,3 +141,14 @@ core.*
 .pixi
 *.egg-info
 node_modules
+
+# CLion / JetBrains IDEs
+.idea/
+*.iml
+*.iws
+.idea_modules/
+cmake-build-*/
+# CLion / JetBrains user-specific and generated files
+.clwb/
+.run/
+out/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 ### Changed
 
 ### Fixed
+- spline drawing: restored the `close`/`c` command (it was disabled), which closes the loop by returning to the first control point and finalizes
+- spline drawing: finalizing with too few control points now reports a message and keeps the in-progress spline instead of silently discarding it
 - various issues in LFF font files
 - segfault with malformed LFF font files
 - CVE-2023-30259, OOB reads in importshp plugin (#1481)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2485,6 +2485,7 @@ if (BUILD_TESTS)
         librecad/src/lib/filters/tests/dwg_buffer_round_trip_tests.cpp
         librecad/src/lib/creation/tests/lc_action_draw_line_parallel_tests.cpp
         librecad/src/lib/creation/tests/lc_action_draw_polyline_tests.cpp
+        librecad/src/lib/creation/tests/lc_action_draw_spline_tests.cpp
         librecad/src/lib/engine/undo/tests/rs_undo_tests.cpp
         librecad/src/lib/filters/tests/dwg_entity_encode_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_header_encode_round_trip_tests.cpp
@@ -2812,6 +2813,9 @@ if (BUILD_TESTS)
     add_librecad_filter_micro_test(
         librecad_polyline_command_tests
         librecad/src/lib/creation/tests/lc_action_draw_polyline_tests.cpp)
+    add_librecad_filter_micro_test(
+        librecad_spline_command_tests
+        librecad/src/lib/creation/tests/lc_action_draw_spline_tests.cpp)
     add_executable(librecad_console_command_utils_tests
         EXCLUDE_FROM_ALL
         librecad/src/main/lc_release_label.cpp
--- a/librecad/src/actions/drawing/draw/spline/lc_action_draw_spline.cpp
+++ b/librecad/src/actions/drawing/draw/spline/lc_action_draw_spline.cpp
@@ -144,13 +144,14 @@ void LC_ActionDrawSpline::onMouseLeftButtonRelease([[maybe_unused]] int status, 
 
 void LC_ActionDrawSpline::onMouseRightButtonRelease(const int status, [[maybe_unused]] const LC_MouseEvent* e) {
     const auto spline = m_actionData->spline;
-    if (status == SetNextPoint && spline != nullptr) {
-        const size_t nPoints = spline->getNumberOfControlPoints();
-        const bool isClosed = spline->isClosed();
-        // Issue #1689: allow closed splines by 3 control points
-        if (nPoints > static_cast<size_t>(spline->getDegree()) || (isClosed && nPoints == 3)) {
-            trigger();
+    if (status == SetNextPoint) {
+        if (spline == nullptr ||
+            spline->getNumberOfControlPoints() <= static_cast<size_t>(getDegree())) {
+            commandMessage(tr("Cannot finalize spline: at least %1 control points required.")
+                           .arg(getDegree() + 1));
+            return;
         }
+        trigger();
     }
     deletePreview();
     initPrevious(status);
@@ -195,12 +196,11 @@ bool LC_ActionDrawSpline::doProcessCommand(const int status, const QString& comm
             break;
         }
         case SetNextPoint: {
-            /*if (checkCommand("close", c)) {
+            if (checkCommand("close", command)) {
                 close();
-                updateMouseButtonHints();
-                return;
-            }*/
-            if (checkCommand("undo", command)) {
+                accept = true;
+            }
+            else if (checkCommand("undo", command)) {
                 undo();
                 updateActionPrompt();
                 accept = true;
@@ -223,7 +223,7 @@ QStringList LC_ActionDrawSpline::getAvailableCommands() {
             if (m_actionData->history.size() >= 2) {
                 cmd += command("undo");
             }
-            else if (m_actionData->history.size() >= 3) {
+            if (m_actionData->history.size() > getDegree()) {
                 cmd += command("close");
             }
             break;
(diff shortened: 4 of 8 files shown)


dxli:
@ideocentric 

Did you test with the command line feature? Say, the end user should be able specify control points by coordinates from the command line, so command line activation should be supported (shortcut: spacebar).

Finalizing the spline should only be enabled, if there's a valid preview (meaning total number of validate control points is  large enough.

ideocentric:
I am reviewing the code now.. the current application does not seem to understand the close command prior to the change I made.  In this case I will review the code and propose fixing both the gui and command for this.

dxli:
> I am reviewing the code now.. the current application does not seem to understand the close command prior to the change I made. In this case I will review the code and propose fixing both the gui and command for this.

We are using some shared shortcuts for all actions, like the colon(:) spacebar to activate the command widget. Also, typing in coordinates, including '@' to indicate relative coordinates, should not be effected.

A simple fix is probably to onlly accept the event for Enter.

ideocentric:
**Reworked to address the review — summary of the evaluation and approach**

Thanks @dxli. The latest commit reworks this to cover all three points.

**Enter only.** `keyPressEvent` now handles `Qt::Key_Enter` and ignores everything else. The canvas already routes a physical Return through `LC_EventHandler::enter()` as a synthetic `Key_Enter`, so the earlier `Key_Enter`/`Key_Return` pair was redundant and risked touching shared shortcuts. Command-widget activation (`:`/spacebar), coordinate entry, and `@`-relative coordinates are untouched — those keys are ignored and propagate normally.

**Command-line control points verified.** `spline` → `0,0` → `0,10` → `10,10` → `10,0` → Enter builds and commits; absolute `x,y`, relative `@x,y`, and polar `d<a` all reach `onCoordinateEvent` independently of the finalize handler.

**Finalize gated on a valid preview.** Enter and the right mouse button now share one `finalizeSpline()` that requires enough control points to form a drawable curve — `> degree` (i.e. `degree + 1`) for the control-point spline, `≥ 3` for the piecewise-quadratic through-points spline. If there aren't enough, it shows a message and keeps drawing instead of silently discarding the spline. This drops the old `(closed && nPoints == 3)` case, which committed a closed cubic that fails `RS_Spline::validate()` (needs `unwrapped ≥ degree + 1`) and renders empty.

**`close` / `c` restored.** It was commented out, and `getAvailableCommands()` had inverted logic that made it unreachable. Modeled on the polyline/line `close`: it re-issues the first control point as the closing point and finalizes, so the curve returns to its origin — the appended point supplies the `(degree + 1)`-th control point, letting a spline with as few as `degree` points close. (The original commented-out `close()` used `RS_Spline::getStartpoint()`, an unimplemented stub returning an invalid vector — likely why it was disabled; it now uses the stored first control point.)

**Modeling note.** We compared against the polyline and line tools. The structural difference is that `li`/polyline commit incrementally (each segment written as you go) while the spline is deferred — one entity held in the action, committed only on finalize. That's why it uniquely needs a dependable finalize and must never discard on an under-specified finish; verification therefore lives inside the write path so every gesture is guarded identically.

Tested interactively: finalize with enough/too-few points (Enter + RMB), `close`/`c` at 3 points and too-few, command-line coordinates, `undo`, and the through-points variant.

librecad/src/actions/drawing/draw/spline/lc_actiondrawsplinepoints.cpp
```diff
@@ -112,10 +114,38 @@ void LC_ActionDrawSplinePoints::onMouseLeftButtonRelease([[maybe_unused]]int sta
...
+        e->ignore();
+    }
+}
+
+void LC_ActionDrawSplinePoints::finalizeSpline() {
+    // A drawable points-spline is a piecewise quadratic, so a real curve needs at
+    // least three points (fewer only yield a point or a straight line). Verifying
+    // here keeps every finalize gesture guarded and never discards work silently.
+    if (m_actionData->spline.get() == nullptr
+        || m_actionData->spline->getData().splinePoints.size() < 3){
```
dxli: hard-coded number 3 is only good for quadratic splines. For example, a linear spline requires 2 control points(line segments), while a cubic spline would require 4 control points(if not closed).

It's probably better to use RS_Spline::validate().

Please note our RS_Spline implementation is ancient. Instead drawing true splines, it draws line segments to simulate splines.

We probably want to redo the implementation.

dxli (changes requested):

librecad/src/actions/drawing/draw/spline/lc_actiondrawsplinepoints.cpp
sand1024: @dxli 
@ideocentric 

I don't think this is a right direction. 

I'm concerned by completely different thing - this is a new way for completing command - which is supported only for splines. 

So such approach does not look consistent with other commands, as for me. 

Actually there are lots of commands that my be finished by right button - so I think a more correct way is 
1) review all commands;
2) check which ones may benefit from completing by Enter 
3)support this as a common command completion mechanism (probably even on higher level than the command)

Rather than adding this feature only for these two ones (as here it looks just like a shortcut or patch, that is not consistent with other places). 

At the end, commands classes may be even untouched and handling closing of command by pressing 
Return will be handled earlier.

ideocentric:
Thanks @sand1024 and @dxli — both points are well taken, and I'd rather align with the project's conventions than push a one-off.

**On the direction (@sand1024).** You're right that Enter-to-finish isn't the established pattern here. I checked the tree to be sure I understood it: the base `RS_ActionInterface::keyPressEvent()` just `ignore()`s, polyline/line don't override it, and finishing a multi-point object is done with the right mouse button (plus the `close` command). The only draw action that currently completes on Enter is `RS_ActionDimLeader`. So adding it to splines alone would be inconsistent — point taken.

My motivation was simply that Enter-to-complete (alongside right-click) is the norm in most CAD/drawing programs I've used, and with the behavior undocumented I was deriving intent from the source. But that's exactly why a shared mechanism is the better answer: the plumbing already exists — a physical Return is routed `QC_ApplicationWindow → onEnterKey → processEnterKey → LC_EventHandler::enter()`, which synthesizes a `Key_Enter` to the current action. So Enter could be handled once at that layer (or a base-class hook) for any action with a committable preview, rather than per-action. If there's appetite for that, I'm happy to prototype it separately and keep this PR narrow.

**On the validity gate (@dxli).** Fully agree on not hard-coding a magic number. One clarification from reviewing the two tools, in case it's useful — there are two different entities in play:
- `RS_ActionDrawSpline` → `RS_Spline`, a variable-degree control-point B-spline (1–3). Its gate already derives from `getDegree()` (needs `degree + 1`), not a literal.
- `LC_ActionDrawSplinePoints` → `LC_SplinePoints`, which is fixed piecewise-**quadratic** (no degree field) and is validated by `LC_SplineHelper::validate()` rather than `RS_Spline::validate()`. The `< 3` is the quadratic minimum for *that* entity, but you're right it shouldn't be a bare literal.

Proposed fix: have each finalize path defer to the entity's own validator (`RS_Spline::validate()` for the control-point spline, `LC_SplineHelper::validate()` for the points spline) instead of reimplementing the rule in the action — single source of truth, correct across degrees. And I agree the underlying RS_Spline implementation (line-segment approximation) is due for a rework; glad to help there too if it's on the roadmap.

**Suggested path forward** — splitting the uncontroversial from the design decision:
1. Keep the real bug fixes: restoring the `close`/`c` command (it was commented out + unreachable) and gating finalize on the entity validator.
2. Drop the spline-only Enter handling from this PR, and open a separate proposal for a shared Enter-completion at the event-handler level — only if the team wants it.

Does that split sound right? And on the spline rework, is there an existing issue/direction you'd want me to build on?

sand1024:
@ideocentric 
>The only draw action that currently completes on Enter is RS_ActionDimLeader.

Well, most probably no - as far as I remember, there could be more of such actions (and several ones do not complete automatically to simplify repetitive command operation). I'll check that with more details.

You're right - most probably, that's related to EventHandler. The other thing, that one also should be revised as well as cmd widget for better generic support of keyboards and command line. I plan to rewrite (or at least modify significantly) cmd, yet slightly stuck with other features. 

As for spline implementation - well, yes, right, it should be rewritten. I don't think that it's current line-based approximation is really bad in principle - for me, it's rather the question of it's implementation and performance. 

Actually, I'm not happy at all with the current painting pipeline due to performance reasons (arcs, circles, elipses and splines) - for me, the current rendering is slow, yet simply had no time to take a closer look... 

In general there is long set of features that may be improved/added - yet unfortunately, there is no common backlog for them (and probably one should be created)

ideocentric:
Thanks @sand1024 — that all makes sense, and I'm glad the EventHandler direction lines up with what you're planning.

I don't want to get ahead of the cmd-widget / EventHandler rework you're describing, so let me propose narrowing this PR sharply:

**Scope it down to just the `close`/`c` command for the control-point spline.** Today that command doesn't actually work — the dispatch in `doProcessCommand` is commented out (so typing `close`/`c` does nothing), and the existing `close()` relies on `RS_Spline::getStartpoint()`, which is an unimplemented stub returning an invalid vector. I'd fix it following the existing `RS_ActionDrawPolyline::close()` model (append the first point, validate, commit), with a guard so it only commits a drawable spline. Self-contained bug fix, no new UX convention.

**Drop the Enter-key handling entirely** and defer generic Enter/keyboard completion to your EventHandler + cmd-widget rework — that's clearly the right layer, and I'd rather not introduce a spline-only pattern ahead of it. Dropping it also reverts the points-spline file, which makes @dxli's `validate()` note moot for now; happy to revisit finalize gating when the spline itself is reworked.

The bigger items — a shared Enter-completion mechanism, the spline rewrite, and the painting-pipeline performance — feel like they belong in that rework and a proper backlog. +1 on creating one; glad to help populate/triage if that's useful.

Would that scoped version (close-command fix only) be welcome as this PR, or would you rather close it and fold even that into the larger rework? Either's fine by me.

dxli (dismissed):

ideocentric:
Reworked per the discussion above — this PR is now scoped down to just the `close`/`c` command fix, as proposed.

**What changed since the approval:**
- **Dropped the Enter-key handling entirely.** Deferring generic Enter-to-complete to the `LC_EventHandler` / command-widget rework @sand1024 described — glad to open that as a separate proposal when it's timely.
- **Reverted the points-spline file** (`lc_actiondrawsplinepoints.*`) to master, which also removes the hard-coded `< 3` gate @dxli flagged.
- **Kept the real fix:** restored the `close`/`c` command for the control-point spline. It was commented out in `doProcessCommand`, made unreachable by inverted `getAvailableCommands` logic, and the old `close()` relied on `RS_Spline::getStartpoint()` — an unimplemented stub returning an invalid vector. It now follows the `RS_ActionDrawPolyline::close()` model and is guarded to only commit a drawable spline (≥ degree control points).
- **Right-click finalize** now reports a message and keeps the in-progress spline when there aren't enough control points, instead of silently discarding it.

Net diff is now 4 files (the two control-point spline files, `CHANGELOG`, and a `.gitignore` addition).

@dxli @sand1024 — sorry for the churn after the approval; flagging since the code changed materially. Happy to adjust further.

dxli (dismissed):

Issue #2162, opened 2025-05-27
ARM MacOS builds?


silasb:
I'd like to use ttf2lff but after installing freetype and trying relink freetype to the correct location things didn't work.  That is when I noticed that the binaries are built for X86.  Trying to get a freetype X86 build on my M1 is probably not worth the effort.

If we have ARM MacOS builds this should eliminate some of these issues.

Unless someone knows how to link arm64 libraries to a x86 correctly...

Here are the steps that I went through and got stuck at.

```
install_name_tool -change /usr/local/opt/freetype/lib/libfreetype.6.dylib /opt/homebrew/opt/freetype/lib/libfreetype.6.dylib ttf2lff
```

```
❯ ./ttf2lff /Users/silas/Downloads/franklin-gothic/FranklinGothic.ttf
dyld[32360]: Library not loaded: '/opt/homebrew/opt/freetype/lib/libfreetype.6.dylib'
  Referenced from: '/Applications/LibreCAD.app/Contents/MacOS/ttf2lff'
  Reason: tried: '/opt/homebrew/opt/freetype/lib/libfreetype.6.dylib' (mach-o file, but is an incompatible architecture (have (arm64), need (x86_64))), '/usr/local/lib/libfreetype.6.dylib' (no such file), '/usr/lib/libfreetype.6.dylib' (no such file), '/opt/homebrew/Cellar/freetype/2.13.3/lib/libfreetype.6.dylib' (mach-o file, but is an incompatible architecture (have (arm64), need (x86_64))), '/usr/local/lib/libfreetype.6.dylib' (no such file), '/usr/lib/libfreetype.6.dylib' (no such file)
zsh: abort      ./ttf2lff /Users/silas/Downloads/franklin-gothic/FranklinGothic.ttf
```

dxli:
@silasb 

probably, we need to build ibfreetype from source for arm support.

please have a look at our building script, and we need arm support for mac anyway

[building workflow
](https://github.com/LibreCAD/LibreCAD/blob/master/.github/workflows/build-all.yml)

silasb:
I can help.  When I get back from vacation I'll setup a couple different VMs for testing.

cbertelli:
I tried LibreCAD-v2.2.1.2-17-g7f9a4dff5-arm64.dmg on MacOS 15.7.3 without luck.
If I double click it says: “LibreCAD” is corrupted and cannot be opened.
I tried 
```
xattr -rc LibreCAD.app
```
[...]

Issue #2574, opened 2026-05-20
Damaged Image - LibreCAD-v2.2.1.5-arm64.dmg


GeoffatMM:
Hi,

Hope this is the right place to raise this?

Just downloaded [LibreCAD-v2.2.1.5-arm64.dmg] from (https://github.com/LibreCAD/LibreCAD/releases/download/v2.2.1.5/LibreCAD-v2.2.1.5-arm64.dmg) but it will not open.

It tells me the image is damaged.  I have tried several times with the same result.

I am installing to an iMac 24 with the M1 chip.

dxli:
@GeoffatMM 

For macOS arm64 builds, the app is NOT signed. To workaround the "damaged" error (https://github.com/LibreCAD/LibreCAD/issues/2162):

```bash
xattr -rc LibreCAD.app
sudo codesign --force --deep --sign - LibreCAD.app
```

lokmanexe-commits:
**Environment**

* Mac Studio (M2 Max)
* macOS Sequoia 15.7.7 (24G720)
* LibreCAD v2.2.1.5 arm64 (GitHub Release)

**Problem**

The Intel build works under Rosetta, but the arm64 build cannot be launched.

When launching the application, macOS reports that the application is damaged. The crash report shows:

```
Exception Type: EXC_BAD_ACCESS (SIGKILL (Code Signature Invalid))
Termination Reason: Namespace CODESIGNING, Code 2 Invalid Page
```

I tried the following:

```bash
xattr -cr /Applications/LibreCAD.app
sudo codesign --force --deep --sign - /Applications/LibreCAD.app
spctl -a -vv /Applications/LibreCAD.app
```

The app was re-signed successfully:

```
replacing existing signature
```

However, `spctl` still reports:

```
/Applications/LibreCAD.app: rejected
```

Before re-signing, it reported:

```
code has no resources but signature indicates they must be present
```

I also confirmed that this is the native arm64 build:

```bash
mdls /Applications/LibreCAD.app
```

```
kMDItemExecutableArchitectures = (
    arm64
)
```

The Intel (x86_64) build launches correctly under Rosetta, but the arm64 build consistently fails with the code-signing error.

If needed, I can also provide the full crash report.

rschwiebert:
I might be less sophisticated than the above users, but when I try to use the installer simply nothing happens.

cinpol:
I have the same issue, I cannot open/install the downloaded file on M1. It says the file is damaged.

[...]

PR #2676, opened 2026-07-09
macOS: build a signed universal (arm64 + x86_64) binary

ideocentric:
## Summary

Makes the macOS build produce a single **signed, universal (arm64 + x86_64)** app so it runs natively on both Apple Silicon and Intel, and fixes the arm64 "app is damaged" launch failure (#2162).

## Background

- The arm64 DMG would not launch because `macdeployqt` rewrites the binary's load commands, invalidating the signature, and the script never re-signed the bundle. On Apple Silicon macOS refuses to launch an arm64 binary whose signature is invalid.
- The two macOS CI jobs both ran on the arm64 `macos-14` runner and effectively produced the same architecture: the arm64 job's `QMAKE_APPLE_DEVICE_ARCHS` was silently dropped because `build-osx.sh` reset `QMAKE_OPTS` to an empty string.

## Changes

- **build-osx: honor `QMAKE_OPTS` from the environment** so CI-provided qmake options actually reach qmake.
- **Guard the `arm_acle.h` force-include** behind a small `__aarch64__`-guarded prefix header (`mac_arm_acle_prefix.h`). Qt6's `qyieldcpu.h` uses `__yield()` without including `<arm_acle.h>`; the previous unconditional `-include arm_acle.h` breaks the x86_64 slice of a universal build ("ACLE intrinsics support not enabled"). The guarded header is a no-op on x86_64, so arm64-only, x86_64-only and universal builds all compile.
- **Sign the app bundle after `macdeployqt`** (#2162). With a real Developer ID (`-cert=`) the bundle is signed with the hardened runtime and, when notarytool credentials are provided via the environment, the DMG is notarized and stapled; otherwise an ad-hoc signature is applied so the app still launches. Notarization is skipped by default, so no Apple account is needed for an ordinary build.
- **CI: one universal macOS job** replacing the two near-identical jobs, using `QMAKE_APPLE_DEVICE_ARCHS=arm64 QMAKE_APPLE_DEVICE_ARCHS+=x86_64`.
- **Declare `NSHighResolutionCapable`** in the bundle plist.

## Testing

- Verified locally that the guarded prefix header compiles a universal binary (`x86_64 arm64`), while the previous unconditional include fails on the x86_64 slice.
- Verified both signing paths locally: ad-hoc and a real Developer ID. `codesign --verify --deep --strict` passes, with the full Apple authority chain, hardened runtime and Gatekeeper `accepted (source=Developer ID)`.
- The full universal qmake build and DMG packaging are exercised by CI.

## Fork CI — green

The full **Continuous build** workflow ran green on my fork's `master` (commit `dfcf22f0e`), exercising the universal macOS job end-to-end: universal Freetype → universal (arm64 + x86_64) build → `macdeployqt` → ad-hoc codesign → DMG. All platform jobs (Linux x64/arm64, macOS, Windows x64/arm64, Snap) passed.

Run: https://github.com/ideocentric/LibreCAD/actions/runs/29131972709

(`UpdateAssets` is intentionally skipped on the fork — its `if:` requires `repository_owner == 'LibreCAD'` — so this run only produces build artifacts, not releases.)

## Notes

`LSMinimumSystemVersion` was intentionally left out; it should be set together with a pinned `MACOSX_DEPLOYMENT_TARGET` and a verified minimum macOS version, as a follow-up.


dxli (approved):

dxli (dismissed):
looks we should do the similar to the 2.2.1 branch. After that, the workaround for dmg loading errors in README.md can be removed

ideocentric:
@dxli agreed on getting this onto `2.2.1` so the stable-channel DMGs launch and the README workaround can be removed.

One wrinkle I hit while looking at the port: **`2.2.1` is a Qt5 branch.** Its mac jobs use Qt 5.15 as two separate builds — `clang_64` for Intel (`macos-15-intel`) and `brew qt@5` for arm64 (`macos-15`) — producing `LibreCAD-….dmg` and `LibreCAD-…-arm64.dmg`. This PR's single-universal-binary approach relies on Qt6's `clang_64` package being universal; Qt 5.15's macOS libraries aren't, which is presumably why `2.2.1` has the two separate jobs. So the universal-single-binary part doesn't port cleanly.

What *does* port — and is the part that actually fixes the arm64 "app is damaged" launch — is the **sign-after-`macdeployqt`** change in `build-osx.sh` (deploy → `codesign` the bundle → build the DMG). The `2.2.1` jobs already pass `-cert=-`, so I'd adapt it to treat `-cert=-`/empty identity as an **ad-hoc** signature (dropping `--timestamp`, which fails on ad-hoc) while keeping the real-Developer-ID path for when credentials are present. That would let the existing per-arch DMGs launch and let the README workaround go, without touching the two-job structure.

Two questions before I open anything:
1. Does a targeted **signing-only** fix for the `2.2.1` Qt5 jobs match what you had in mind, or did you want a universal binary there too (which would need a different Qt5 mechanism)?
2. Prefer it as a separate PR against `2.2.1` now, or after this one lands so the approach is settled?

Happy to put it up whenever you say.

dxli:
@ideocentric I did a deep review of this and have pushed an updated branch (your commits and authorship preserved, rebased onto current `master`, now **MERGEABLE**). The approach is sound — I reproduced both of your core premises locally before touching anything:

- bare `-include arm_acle.h` really does fail the x86_64 slice with *"ACLE intrinsics support not enabled"*, and without it `__yield()` fails on arm64; your guarded prefix header compiles clean on x86_64-only, arm64-only **and** fat (`-arch arm64 -arch x86_64`) invocations;
- `build-universal-freetype-macos.sh` runs green and produces a genuine fat static archive, against which a `FT_Init_FreeType` program links *and runs* on both slices.

## What changed

**1. Conflict resolution (was the main blocker).** `#2742` landed on `master` after you wrote this and rewrote the same region of `build-osx.sh`. The textual conflict is a trap: git hoisted your `macdeployqt` call *above* the markers and factored out the shared `if [[ $CODESIGN_IDENTITY ]]`, so the two halves are bodies of **different statements** — and both naive resolutions are wrong. Taking your side reverts `#2742` (virtual-keyboard strip, `-fs HFS+`, `/Applications` symlink); taking master's side means **no `codesign` ever runs on CI** (master only re-signs when the strip fires, which never happens on official Qt), reintroducing exactly the `#2162` failure this PR fixes. Resolved by hand: deploy → strip → **always sign** → verify → staged HFS+ DMG. Master's now-redundant `macdeployqt -codesign=` and its conditional re-sign are dropped, since the unconditional sign supersedes both.

**2. `codesign --deep` never reaches `Contents/Resources/plugins`.** LibreCAD's ten plugin dylibs live there, and `Resources/` is not a nested-code location — verified empirically that `--deep` leaves them untouched. Invisible today (CI is ad-hoc only), but the day a real Developer ID is wired up, **notarization rejects the bundle and library validation refuses to load every plugin**. Now signed explicitly before the outer `--deep` sign.

**3. Freetype supply chain.** The tarball ships inside the released DMG via `ttf2lff`, and the savannah URL redirects to rotating volunteer mirrors (I probed it: 4 distinct hosts in 5 requests, one a personal domain). Now pinned by SHA256 — cross-checked byte-identical from savannah **and** SourceForge — plus `curl -f` (previously an HTTP error page was saved as the tarball and failed later as a confusing `tar` error), HTTPS-only redirects, retries, and a `trap` to clean the temp dir. Verified both ways: correct hash builds, wrong hash **fails closed with nothing built**.

**4. Verification step.** Nothing asserted the two properties this PR exists to deliver. Added `lipo -info` + `codesign --verify --deep --strict` after signing, so a thin or unsigned bundle fails the build instead of shipping. Relevant because universality rides on unquoted word-splitting of `QMAKE_OPTS` — one innocent future quoting "fix" would silently ship thin DMGs.

**5. `bash -x` leaked `NOTARIZE_PASSWORD`** into the log (masked on GitHub, not elsewhere) — wrapped in `( set +x; … )`.

**6. README** (lines 50–53) still said arm64 builds are unsigned and prescribed a manual `codesign`, which this PR makes obsolete. Updated to describe the universal, ad-hoc-signed reality; the quarantine note stays until real notarization is wired up.

Also: trailing newlines on both new files, and master's newer `UpdateAssets` `if:` guard preserved through the merge.

## Verification

Ran the **entire resolved pipeline** locally (Qt 6.11, Apple Silicon): full build → `macdeployqt` → strip → sign → verify → DMG.

- `codesign --verify --deep --strict` → **valid on disk, satisfies its Designated Requirement**
- the built app **launches** (the `#2162` symptom is gone)
- DMG mounts with `LibreCAD.app` + the `/Applications` symlink, partition type `Apple_HFS` — `#2742`'s layout intact
- virtual-keyboard strip fired: `platforminputcontexts/` empty, zero QML/Quick/OpenGL frameworks
- `ttf2lff` has **no freetype dylib dependency** — statically linked against the universal build via `FREETYPE_DIR`

One thing I could **not** test locally: the actual universal link, since Homebrew's Qt is thin arm64 (`lipo` confirms) — that path needs CI's official universal `clang_64` Qt, which your fork run already exercised green. The new `lipo`/`codesign` assertions will now catch any regression there automatically.

## Your two open questions

The `2.2.1` port is largely **moot now**: `#2574` already landed there and passes `-cert=-`, so those jobs ad-hoc sign today. A signing-only port isn't urgent, and I agree the universal-binary half doesn't port to a Qt5 branch. Left as **draft** — flip it when you're happy with the changes.

dxli:
Follow-up: went back over my own additions with a minimum-change eye and trimmed them. Net **−4 lines** versus the previous push, and one of the removals was fixing a mistake of mine.

**Dropped the `lipo -info` loop from `build-osx.sh` (−5 lines).** I had commented it "fail the build rather than shipping a thin app" — that was simply wrong: `lipo -info` exits 0 on a thin binary, so it printed information and asserted nothing. Worse, a *real* universality assertion does not belong in `build-osx.sh` at all, because that script is also used for legitimate single-arch local builds (I ran one). Moved it to the CI job instead, where universal genuinely is the contract, and made it actually assert:

```yaml
lipo -archs LibreCAD.app/Contents/MacOS/LibreCAD | tee /dev/stderr | grep -qw arm64
lipo -archs LibreCAD.app/Contents/MacOS/LibreCAD | grep -qw x86_64
```

Verified it discriminates: passes on fat, **fails on thin**. So a silently-thin "universal" DMG now breaks CI instead of shipping.

**Simplified the plugin signing (−3 lines).** Signing `Contents/Resources` instead of `Contents/Resources/plugins` drops the `if [ -d ... ]` guard entirely — `Contents/Resources` always exists in an app bundle, whereas the `plugins` subdirectory needed guarding against `find` returning 1 under `set -e`. Confirmed it still catches all 10 plugin dylibs, that `find` with no matches exits 0, and it is marginally more robust (any dylib under `Resources`, not just that one subdirectory).

**Kept `codesign --verify --deep --strict`** (dropped its `--verbose=2`): one line, and it directly guards the property this whole reordering exists to produce.

---

While doing this I also questioned the largest single piece — the `arm_acle` prefix header — and **it has to stay**, though not for the reason I assumed. Qt **6.11** reordered `qyieldcpu.h` to try `__builtin_arm_yield` (a real clang builtin, no header needed) *before* `__yield()`, so on 6.11 the workaround is dead code and compiles clean without it. But CI pins **Qt 6.9**, where the branches are in the *opposite* order — `__yield()` is tested first, `__has_builtin(__yield)` is true on arm64 clang, and the build fails with `implicitly declaring library function '__yield'`. Verified against the 6.8/6.9/6.9.0 sources and reproduced the failure locally. Deleting it would have broken CI.

I also tried collapsing the header into a one-line `-Xarch_arm64` flag in `src.pro`; clang rejects it outright — *"invalid Xarch argument … options requiring arguments are unsupported"* — so the prefix header is the right mechanism, not incidental complexity. Worth noting for whenever the Qt floor moves past 6.9: at that point the header and its `src.pro` hunk can both be deleted.

**Re-verified the trimmed pipeline end to end** (full build → deploy → strip → sign → verify → DMG): signature valid, app launches, virtual-keyboard strip intact (0 entries, 0 QML/Quick/OpenGL frameworks), DMG mounts `Apple_HFS` with the `/Applications` symlink, freetype rebuilt fat from the pinned checksum.

PR #2677, opened 2026-07-09
CMake: fix the macOS bundle (name, Info.plist, icon)

ideocentric:
## Summary

Makes a local CMake (e.g. CLion) macOS build produce a proper application bundle that matches the qmake release build: the capitalized name `LibreCAD.app`, a filled-in `Info.plist`, and the application icon.

## Background

The CMake target is named `librecad`, so the build produced `librecad.app` / `Contents/MacOS/librecad`, whereas the qmake release build (`src.pro` sets `TARGET = LibreCAD`) ships `LibreCAD.app`. It also pointed `MACOSX_BUNDLE_INFO_PLIST` at `Info.plist.app`, which uses qmake-style `@TOKEN@` placeholders (`@ICON@`, `@EXECUTABLE@`, `@BUNDLEIDENTIFIER@`, `@FULL_VERSION@`, `@TYPEINFO@`). CMake doesn't define those, so they were substituted to empty — the bundle ended up with no identifier, executable name, version or icon, and the `.icns` was never copied. (That empty plist also made the bundle fail `codesign --verify` with "invalid Info.plist".)

## Changes (macOS only; the Linux `librecad` binary is unchanged)

- Set `OUTPUT_NAME LibreCAD` so the bundle is `LibreCAD.app` with `Contents/MacOS/LibreCAD`.
- Fill the shared `Info.plist.app` template via `configure_file(@ONLY)` with the icon, executable name, bundle identifier (`org.librecad.librecad`) and version.
- Copy `librecad/res/images/librecad.icns` into `Contents/Resources` and reference it as `CFBundleIconFile`.

## Testing

A local CMake build now produces `LibreCAD.app` with `Contents/Resources/librecad.icns`, a valid `Info.plist` (`plutil -lint` OK) carrying the correct `CFBundleIdentifier` / `CFBundleExecutable` / `CFBundleIconFile` / version, and the application icon shows in Finder.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2238,11 +2238,38 @@ qt_add_executable (librecad
                    ${MAIN_SOURCES}
 )
 
-# On macOS, make it a bundle and set info.plist
+# On macOS, produce a proper bundle: the capitalized name (LibreCAD.app with
+# Contents/MacOS/LibreCAD) used by the qmake release build, a filled-in
+# Info.plist and the app icon. The Linux binary stays lowercase "librecad".
 if(APPLE)
+    # Info.plist.app is shared with the qmake build and uses @TOKEN@
+    # placeholders. Fill them here (configure_file @ONLY) so the bundle gets a
+    # valid identifier, name, version and icon instead of empty values.
+    set(ICON "librecad.icns")
+    set(EXECUTABLE "LibreCAD")
+    set(BUNDLEIDENTIFIER "org.librecad.librecad")
+    set(FULL_VERSION "${PROJECT_VERSION}")
+    set(TYPEINFO "????")
+    configure_file(
+        "${CMAKE_CURRENT_SOURCE_DIR}/librecad/src/Info.plist.app"
+        "${CMAKE_CURRENT_BINARY_DIR}/Info.plist"
+        @ONLY)
+    unset(ICON)
+    unset(EXECUTABLE)
+    unset(BUNDLEIDENTIFIER)
+    unset(FULL_VERSION)
+    unset(TYPEINFO)
+
+    # Copy the application icon into the bundle's Resources.
+    set(LIBRECAD_ICNS "${CMAKE_CURRENT_SOURCE_DIR}/librecad/res/images/librecad.icns")
+    set_source_files_properties("${LIBRECAD_ICNS}" PROPERTIES
+        MACOSX_PACKAGE_LOCATION "Resources")
+    target_sources(librecad PRIVATE "${LIBRECAD_ICNS}")
+
     set_target_properties(librecad PROPERTIES
         MACOSX_BUNDLE TRUE
-        MACOSX_BUNDLE_INFO_PLIST "${CMAKE_CURRENT_SOURCE_DIR}/librecad/src/Info.plist.app"
+        MACOSX_BUNDLE_INFO_PLIST "${CMAKE_CURRENT_BINARY_DIR}/Info.plist"
+        OUTPUT_NAME LibreCAD
     )
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2244,10 +2244,10 @@ qt_add_executable (librecad
 if(APPLE)
     # Info.plist.app is shared with the qmake build and uses @TOKEN@
     # placeholders. Fill them here (configure_file @ONLY) so the bundle gets a
-    # valid identifier, name, version and icon instead of empty values.
+    # valid name, version and icon instead of empty values. The bundle
+    # identifier is spelled out in the template itself.
     set(ICON "librecad.icns")
     set(EXECUTABLE "LibreCAD")
-    set(BUNDLEIDENTIFIER "org.librecad.librecad")
     set(FULL_VERSION "${PROJECT_VERSION}")
     set(TYPEINFO "????")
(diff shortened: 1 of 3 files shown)


dxli (approved):

dxli (dismissed):

dxli:
Reviewed this in depth, then pushed a second commit extending it to the qmake side so both build systems agree. Rebased onto current `master` (which now includes #2676); **MERGEABLE**.

## Your original commit: verified by building it

I built it and checked every claim — all hold:

| Claim | Result |
| --- | --- |
| `LibreCAD.app` with `Contents/MacOS/LibreCAD` | ✅ `Linking CXX executable LibreCAD.app/Contents/MacOS/LibreCAD` |
| Icon copied into Resources | ✅ `Copying OS X content LibreCAD.app/Contents/Resources/librecad.icns` |
| Valid, filled `Info.plist` | ✅ `plutil -lint` OK, all keys populated |
| Fixes the `codesign` failure | ✅ sign + `codesign --verify --deep --strict` now pass |
| Linux unaffected | ✅ `OUTPUT_NAME` is inside `if(APPLE)` |

I also hit the bug independently before reading your PR: a `LibreCAD.app` in my tree had `CFBundleIdentifier => ""`, `CFBundleExecutable => ""` and failed `codesign` with *"invalid Info.plist"* — with a lowercase `librecad` inside, i.e. a CMake build. A wrinkle worth knowing: APFS is case-insensitive, so `librecad.app` and `LibreCAD.app` are the *same directory* — that bundle was a mixture of CMake's `librecad` and qmake's `ttf2lff`. Your change ends that.

Checks that came back clean: no variable shadowing (`ICON`/`EXECUTABLE`/`FULL_VERSION`/`TYPEINFO` appear nowhere else in the 2,500-line file); `Info.plist.app` is correctly registered as a configure dependency; Qt's own generated `.qt/info_plist/librecad/Info.plist` is byte-identical, so no conflict; `install(TARGETS librecad BUNDLE ...)` follows `OUTPUT_NAME` automatically; nothing else depends on the old output name; and `configure_file` is genuinely the right mechanism, since the template is shared with qmake and uses qmake's `@TOKEN@` names, which CMake's native `MACOSX_BUNDLE_*` variables can't drive.

## Second commit: making qmake match

Your PR exposed that the **qmake** build — the one that ships — had three wrong values. I confirmed all three in the DMG CI published from `master` minutes ago (`619-ge51b4d079`):

```
CFBundleIdentifier          com.yourcompany.LibreCAD
CFBundleShortVersionString  2.2.2_alpha1
CFBundleGetInfoString       string LibreCAD 2.2.2_alpha1-619-ge51b4d079
```

- **Identifier.** qmake builds this from `QMAKE_TARGET_BUNDLE_PREFIX` (default `com.yourcompany`) plus `TARGET`, so it can't produce lowercase `org.librecad.librecad`. I verified with a throwaway qmake project that setting the prefix alone gives `org.librecad.LibreCAD`, and that `QMAKE_BUNDLE` has no effect — the substitution happens in qmake's C++, not a `.prf`. So the identifier is now spelled out in the shared `Info.plist.app` rather than left as a `@BUNDLEIDENTIFIER@` token, which makes one value serve both builds. The CMake side no longer needs to set it, so that `set`/`unset` pair is removed.
- **Version.** `VERSION` only stripped from the first `-`, leaving `2.2.2_alpha1`, which isn't the numeric `x.y.z` Apple wants. Now reduced to `2.2.2`, matching `PROJECT_VERSION`. The full descriptive version is still preserved in `CFBundleGetInfoString`.
- **`CFBundleGetInfoString`.** `PlistBuddy`'s `Set` takes no type argument (only `Add` does), so the literal word `string` was landing in the value.

**One correction worth flagging, because a build caught it and reading would not have.** My first attempt used a `sed -E` regex with `\.` and `\1`; qmake eats the backslashes, `VERSION` came out empty, and qmake silently fell back to **`1.0.0`**. (The pre-existing `s/\-.*//g` escaping its dash was the clue.) Rewritten with `cut`, which has nothing for qmake to mangle, and verified against every real `LC_VERSION` shape including the `v2.2.1.5` tag form on `2.2.1`.

## Verification

Full qmake **and** CMake builds, then compared the two plists key by key:

```
CFBundleIdentifier          MATCH  org.librecad.librecad
CFBundleExecutable          MATCH  LibreCAD
CFBundleIconFile            MATCH  librecad.icns
CFBundleShortVersionString  MATCH  2.2.2
CFBundleVersion             MATCH  2.2.2
```

Both `plutil -lint` clean, both bundles launch, `CFBundleGetInfoString` is now `LibreCAD 2.2.2.6-alpha` with no stray `string`.

## Unrelated note

While tracing this I checked whether #2676's `arm_acle` fix was complete: `common.pri:80` still has the unguarded `-include arm_acle.h` under `equals(QT_ARCH, arm64)`, and it's inherited by every subproject. It does **not** affect CI — I confirmed `master`'s `BuildMacOS` job succeeded and its DMG is genuinely universal (`LibreCAD` and `ttf2lff` both `x86_64 arm64`, signature verifying, `/Applications` symlink present), so `QT_ARCH` evidently isn't plain `arm64` for the official universal Qt. It would only bite a *local* universal build against Homebrew's Qt, where I did see the flag applied. Not touched here — mentioning it in case you want a follow-up.

PR #2680, opened 2026-07-10
DWG support: round 8

dxli:
## Summary

QA fixes:

1. **Null pointer crash prevention** - Comprehensive null pointer guards throughout entity iteration and property access paths. Prevents SIGSEGV crashes when entities are in transient states (e.g. mid-import in dummy containers) with null parents, null layers, or null graphic references. Affects 22+ iteration loops in `rs_entitycontainer.cpp` plus sites in `lc_dimstyle.cpp`, `rs_dimension.cpp`, `rs_graphic.cpp`, `lc_undosection.cpp`, `lc_graphicviewport.cpp`, `lc_copyutils.cpp`, `doc_plugin_interface.cpp`, and `lc_simpletests.cpp`.

2. **Remove spurious DWG support warning** - The unconditional "DWG support is not complete!" message box was shown before every DWG import regardless of whether the file would open successfully. Removed this pre-emptive nag while preserving the actual error handling that shows appropriate messages on real import failures.

3. **DWG/DXF version console output on import** - Prints the file version (e.g. `DWG file version: AC1032/R2018`) to stdout on every successful DWG/DXF import for diagnostics. Also adds a diagnostic test case for the visualization_condominium_with_skylight.dwg file.

4. Insert transform fixes: fixed misplaced inserts.

## Test Plan

- [ ] Build succeeds on all platforms (Linux, macOS, Windows)
- [ ] Open a DWG file - no warning dialog shown, version printed to console
- [ ] Open makeall-plus.dwg - entities render correctly, no crashes
- [ ] Hover over entities in drawing - no SIGSEGV from null parent/layer
- [ ] Copy/paste entities - no crash
- [ ] Undo/redo operations - no crash


sand1024:
@dxli 
I'm just curious, which specific issue is addressed by this merge? Are there any specific issues with closing exists? The only one I remember were related to saving cycle, other than that closing was quite smooth. Is it for file closing or for app closing?  
I'm quickly checked the code and some fragments looks strange a bit for me (similar to AI-generated one), so I'm not sure what for these fixes are. Could you please expand on? 
Thank you.

sand1024:
@dxli 
I suspect that your fixes is applied not to the source place, but also to the last part in call chain. 

I think that modification of code within LC_QuickInfoEntityData and LC_QuickInfoWidget is not needed, and they just overload the interface and responsibilities of the class. 

The checking shutdown state should be performed earlier, in caller code. 

The reason for this is quite straightforward - that widget is just one of possible places that may suffer from incorrect state during application closing. 

The same situation with closing may occur with other functionality, so it should be fixed on highest level possible. 

Please consider this.

librecad/src/lib/engine/undo/lc_undosection.cpp
```diff
@@ -77,8 +77,12 @@ bool LC_UndoSection::undoableExecute(const RS_Document::FunUndoable& doUndoable,
     if (success) {
         if (!ctx.entitiesToDelete.isEmpty()) {
             for (const auto e: std::as_const(ctx.entitiesToDelete)) {
+                if (e == nullptr) {
```
sand1024, line 80: @dxli 
how it could be null there??? It's severe developer bug it is so, it is better to use assert if it so. 
Otherwise, it's just a redundant check and performance overhead... 
It looks like over-defensive programming all over the code - while it could be just enough to check incoming data only once where they occur and just to early check or fail-fast

librecad/src/lib/engine/undo/lc_undosection.cpp
```diff
@@ -77,8 +77,12 @@ bool LC_UndoSection::undoableExecute(const RS_Document::FunUndoable& doUndoable,
...
+                    continue;
+                }
                 const auto layer = e->getLayer(true);
-                if (!layer->isLocked()) {
+                // Null layer: treat as unlocked (e.g. mid-import / no graphic).
+                if (layer == nullptr || !layer->isLocked()) {
```
sand1024, line 85: @dxli 
again, I'm not sure I understand when layer could be null for valid entity...

librecad/src/lib/engine/document/rs_graphic.cpp
```diff
@@ -885,36 +885,34 @@ void RS_Graphic::setDefaultDimStyleName(const QString& name) {
...
 }
 
 LC_DimStyle* RS_Graphic::getEffectiveDimStyleForEdit(const QString& styleName, const RS2::EntityType dimType,
                                                      const LC_DimStyle* styleOverride) const {
-    const auto globalDimStyle = getResolvedDimStyle(styleName, dimType);
-    LC_DimStyle* resolvedDimStyle = nullptr;
+    LC_DimStyle* globalDimStyle = getResolvedDimStyle(styleName, dimType);
     if (styleOverride == nullptr) {
-        resolvedDimStyle = globalDimStyle->getCopy();
+        return globalDimStyle != nullptr ? globalDimStyle->getCopy() : nullptr;
```
sand1024, line 968: @dxli 
if global dimStyle is null - it's just a severe developer bug, why check it each time?

librecad/src/lib/engine/document/entities/rs_entity.cpp
```diff
@@ -922,10 +951,13 @@ RS_Pen RS_Entity::getPenResolved() const {
     const bool widthByLayer = p.isWidthByLayer();
     const bool lineByLayer = p.isLineTypeByLayer();
     if (colorByLayer || widthByLayer || lineByLayer) {
-        const RS_Layer* l = getLayerResolved();
+        // Drop dangling layer pointers left on block members after import
```
sand1024, line 954: @dxli 
Hm.. I not sure I understand where form incorrect layers occurs - probably either from blocks or during some import. 

However, even if it is so - why not consider updating the import procedure and performing all needed data cleanup only once right after import? So later all entities will be without dangling layer pointer. 

Why try to detect all that garbage data that my occur once each time during further runtime operation and so just do not needed repetitive checks? 

I'm not sure this is correct.

librecad/src/lib/engine/document/entities/rs_entity.cpp
```diff
@@ -788,13 +789,37 @@ RS2::Unit RS_Entity::getGraphicUnit() const {
...
+    RS_Layer *l = nullptr;
     if (m_layer == nullptr /*|| layer->getName()=="ByBlock"*/) {
         if (m_parent != nullptr) {
             return m_parent->getLayerResolved();
         }
         return nullptr;
     }
-    return m_layer;
+    l = m_layer;
+    // When the entity is in a document, never hand out dangling layer
```
sand1024, line 800: @dxli 
well, this is quite a "popular" function... and I don't understand why it is necessary to check for the invalid layer each time on runtime there, instead of correct mess in data (if any) earlier and once.

librecad/src/lib/engine/document/container/rs_entitycontainer.cpp
```diff
@@ -100,6 +100,9 @@ RS_EntityContainer::RS_EntityContainer(const RS_EntityContainer& other) : RS_Ent
     if (m_autoDelete) {
         // fixme - sand - check this logic, looks suspicious!
         for (auto& it : *this) {
+            if (it == nullptr) {
```
sand1024, line 114: @dxli 
I've also note that sometimes containers contains nullptr instead of valid reference to entity - yet didn't investigated how they occurs there so far.

I suppose we'd better prevent holding nulls in entities list in container (or probably add custom iterator, which will skip nulls - which is less preferable approach, of course) - rather than overload the codebase with all that numerous check for nulls.

dxli:
@sand1024

I was doing DWG/DXF support QA. The crashes became a blocker for frequent file opening/closing. That was the reason why I diverged a little to troubleshoot and add those quick fixes.

Yes, it's much better if we could bisect to find the root cause for proper fixes. The null pointers are just symptoms, and the real issue could be in design. If confirmed, the fix should be of the highest priorities, as we didn't have such a serious data model issue in years.

Issue #2681, opened 2026-07-11
[BUG] The main line inside a Dimension element does not change of layer


vicdoval:
The main line inside a Dimension element does not change of layer/color

**To Reproduce**
1 Create a Dimension element inside one layer 
2 Change the object layer to another layer with a different layer color with Modify Attributes command or with the properties dialog
The main line will continue to have the color of the first layer. The number and other secondary lines change as expected

**Expected behavior**
Everything should change accordingly

<img width="1330" height="920" alt="Image" src="https://github.com/user-attachments/assets/fb6fe531-1581-4a22-bcbe-740f5071ed7c" />

I'm using Version: v2.2.1.5
Compiler: GNU GCC 11.4.0
Compiled on: May  2 2026
Qt Version: 5.15.2
Boost Version: 1.74.0
System: Linux Mint 22.2

qwilvove:
Hello,

After some tests, I only see this issue if the text in dimensions is horizontal. Dimension lines are splitted to create clear space to see text. Those splitted lines are using the wrong colour (colour of the current layer in my tests).

PR #2682, opened 2026-07-11
Fix proposal for #2681 - Wrong dimension line colour when text is horizontal

qwilvove:
Hello,

Here is a fix proposal for #2681.

Applying pen is missing for splitted lines on dimensions if text is horizontal.

Best regards,

qwilvove

### DIFF
--- a/librecad/src/lib/engine/rs_dimension.cpp
+++ b/librecad/src/lib/engine/rs_dimension.cpp
@@ -316,8 +316,14 @@ void RS_Dimension::updateCreateHorizontalTextDimensionLine(const RS_Vector& p1,
         // extend dimension line outside arrows
         if (showArrows) {
             auto dir = RS_Vector::polar(arrowSize*2, dimensionLine->getAngle1());
+            // same pen/layer as the main dimension line, so the pieces
+            // follow the dimension's layer and pen (issue #2681)
             dimensionLineOutside1 = new RS_Line{this, p1 - dir, p1};
+            dimensionLineOutside1->setPen(pen);
+            dimensionLineOutside1->setLayer(nullptr);
             dimensionLineOutside2 = new RS_Line{this, p2 + dir, p2};
+            dimensionLineOutside2->setPen(pen);
+            dimensionLineOutside2->setLayer(nullptr);
         }
 
         // move text to the side if it won't fit either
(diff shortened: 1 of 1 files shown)


dxli:
@qwilvove 

sorry, didn't realize you already fixed this one.

Since I have pushed PR #2684, need to really review whether your fix is the right way to go.

qwilvove:
@dxli

No problem.

I've tested with the changes made in PR #2684, but unfortunately, the issue still seems to be there.

Here is the sample DXF file that I use to test the issue #2681:
[dim-test.zip](https://github.com/user-attachments/files/29969814/dim-test.zip)

I really think that the issue is related to the use of horizontal text in dimensions.

Best regards,

dxli:
@qwilvove Thank you for the fix — it is correct, and it is the missing half of #2681 that #2684 could not reach. I have updated the branch (with your commit and authorship intact, `98a928a89`):

- **Rebased onto current `2.2.1`**, so it now sits on top of the merged #2684. The two fixes are complementary: #2684 makes the *dimension's own* pen resolve ByLayer after a layer change, but that only helps children whose pens are ByBlock. These four lines had a default-constructed `RS_Pen` (valid, concrete — never inherits), which is exactly why they ignored the layer change. Your `setPen(pen)` + `setLayer(nullptr)` puts them on the same resolution path as the main line, arrows and ticks.
- **Added three short comments** at the two creation sites; no code changes beyond your original eight lines.

Verification done on the updated branch:

- **Audit for completeness**: these four lines are the only children in the whole dimension family on `2.2.1` (`rs_dimension` both text branches, aligned, linear, radial, diametric, angular, dimarc) added without `setPen`/`setLayer` — nothing else needs the same treatment.
- **Empirical A/B with your `dim-test.dxf`**, rendered via `librecad dxf2pdf` and counting stroke-colour operators in the PDF:
  - before the fix: 55 cyan + 58 green + **12 black** strokes (the split segments of the 6 dimensions);
  - after the fix: 61 cyan + 64 green + **0 black** — the 12 wrong strokes move exactly to the two layer colours.
- **Full build** of the branch passes (Qt 5.15, macOS arm64), and the export paths were checked: R2000+ DXF never serializes dimension children (they are regenerated via `update()` on load, so loaded files get the fix), and the R12 `*D` block path now writes proper ByBlock attributes for these segments instead of the bogus default pen.

Ready to merge from my side.

dxli (approved):

PR #2683, opened 2026-07-12
Fix #2679: SIGSEGV on quit from stale mouse-move during teardown

### LINKED ISSUE #2679: [BUG] SIGSEGV on quit: Quick Info hover processed during doCloseAllFiles() teardown
**Description**

LibreCAD crashes with `SIGSEGV` when the main window is closed / the application is quit.

During shutdown, `QC_ApplicationWindow::doCloseAllFiles()` calls `processEvents()`, which dispatches a still-queued mouse-move event. That event drives the default action's hover highlighting, which feeds the hovered entity to the Quick Info widget and dereferences an entity/viewport that is already being torn down.

The crash occurs on the **close path only** — never while drawing. It is timing-dependent on whether a mouse-move is queued at close time; hovering the cursor over an entity in the drawing area as you close the window makes it reliable.

**Steps to reproduce**

1. Launch LibreCAD.
2. Create/open a drawing containing at least one entity, and move the mouse over the drawing area.
3. Close the main window (window close button, or Cmd-Q / Ctrl-Q).
4. LibreCAD crashes (`SIGSEGV`) during shutdown.

**Expected behavior**

The application closes cleanly.

**Environment**

- OS: macOS 15.7.7 (24G720), arm64 (Apple Silicon)
- Version: `master` @ commit `65743eb0a` (debug build)
- Compiler: Apple clang 17.0.0
- Qt: 6.11.1

**Analysis**

Signal 11 (SIGSEGV), fault address `0x118` (consistent with a near-null pointer + member offset).

The relevant part of the call stack (top → down to the close handler):

```
2   LC_QuickInfoEntityData::addVectorProperty(QString const&, RS_Vector const&, PropertyType)
3   LC_QuickInfoEntityData::collectGenericProperties(RS_Entity*)
4   LC_QuickInfoEntityData::processEntity(RS_Entity*)
5   LC_QuickInfoWidget::processEntity(RS_Entity*)
6   RS_ActionDefault::updateQuickInfoWidget(RS_Entity*)
7   RS_ActionDefault::highlightHoveredEntities(LC_MouseEvent*)
8   RS_ActionDefault::onMouseMoveEvent(int, LC_MouseEvent*)
9   RS_PreviewActionInterface::mouseMoveEvent(QMouseEvent*)
10  LC_EventHandler::mouseMoveEvent(QMouseEvent*)
11  QG_GraphicView::mouseMoveEvent(QMouseEvent*)
... (Qt event delivery) ...
27  QCoreApplication::processEvents(...)
28  QC_ApplicationWindow::doCloseAllFiles()
29  QC_ApplicationWindow::tryCloseAllBeforeExist()
30  QC_ApplicationWindow::closeEvent(QCloseEvent*)
```

(Frames 31–63 are the AppKit/Qt window-close-button dispatch into `closeEvent`.)

It is a teardown race: `doCloseAllFiles()` pumps `processEvents()` (frames 27–28) while the document/viewport is being destroyed, and a queued mouse-move (frames 8–24) reaches `RS_ActionDefault::highlightHoveredEntities` → the Quick Info widget → `LC_QuickInfoEntityData::collectGenericProperties` → `addVectorProperty` (frame 2), which dereferences a dangling entity/viewport (most likely `getRelativeZero()` on an already-destroyed view).

This is the same fragile hover / `getNearestEntity` subsystem as #2670 (fixed in #2672).

Possible fix directions:
- Avoid dispatching hover / Quick Info updates during shutdown — e.g. guard the `processEvents()` in `doCloseAllFiles()`, or short-circuit `RS_ActionDefault::highlightHoveredEntities()` / the Quick Info update when the view or document is closing.
- Null-guard the entity and view/viewport in the Quick Info path (`LC_QuickInfoEntityData::collectGenericProperties` / `addVectorProperty`).

ideocentric:
## Problem

LibreCAD can SIGSEGV on quit (exit 139). Closes #2679.

When closing, `QC_ApplicationWindow::doCloseAllFiles()` calls `qApp->processEvents(QEventLoop::AllEvents, 1000)` *after* `doClose()` has already detached the active view via `setupWidgetsByWindow(nullptr)` — which nulls the Quick Info widget's viewport. A stale queued **mouse-move** is then dispatched during teardown and flows through the still-active hover path:

```
processEvents → RS_ActionDefault::onMouseMoveEvent → highlightHoveredEntities
  → LC_QuickInfoWidget::processEntity → collectGenericProperties → addVectorProperty
  → LC_QuickInfoBaseData::getRelativeZero() → m_viewport->getRelativeZero()   // m_viewport == nullptr
```

`getRelativeZero()` dereferences the null viewport (fault at offset `0x118`), crashing on quit.

## Fix

Defense in depth, from root cause to leaf:

- **Root** — `doCloseAllFiles()` now processes events with `QEventLoop::ExcludeUserInputEvents`, so queued mouse/keyboard events are not dispatched during teardown. Repaints and deferred deletions still run.
- **Guards** — `LC_QuickInfoWidget::processEntity()` returns early when there is no active view, and `LC_QuickInfoBaseData::getRelativeZero()` null-guards the viewport, so the hover/Quick Info path can never dereference a detached viewport.

## Testing

macOS (Apple Silicon), Qt 6, debug build: reproduced the pre-fix crash (open a drawing, hover an entity, quit → SIGSEGV / exit 139), then confirmed it exits cleanly after the fix. Normal Quick Info hover behavior during editing is unchanged.

### DIFF
--- a/librecad/src/ui/dock_widgets/entity_info/lc_quickinfobasedata.cpp
+++ b/librecad/src/ui/dock_widgets/entity_info/lc_quickinfobasedata.cpp
@@ -250,5 +250,11 @@ QString LC_QuickInfoBaseData::formatInt(const int &x) const{
 }
 
 const RS_Vector &LC_QuickInfoBaseData::getRelativeZero() const {
+    if (m_viewport == nullptr){
+        // No active viewport (e.g. during teardown, see issue #2679): return a
+        // safe default instead of dereferencing a null pointer.
+        static const RS_Vector zero{0.0, 0.0};
+        return zero;
+    }
     return m_viewport->getRelativeZero();
 }
--- a/librecad/src/ui/dock_widgets/entity_info/lc_quickinfowidget.cpp
+++ b/librecad/src/ui/dock_widgets/entity_info/lc_quickinfowidget.cpp
@@ -149,6 +149,14 @@ LC_QuickInfoWidget::~LC_QuickInfoWidget(){
  * @param en entity
  */
 void LC_QuickInfoWidget::processEntity(RS_Entity *en){
+    // During application/document teardown the active view is detached via
+    // setGraphicView(nullptr), leaving the entity data with a null viewport.
+    // Guard against a stale mouse-move being processed then, which would
+    // dereference the null viewport in getRelativeZero() (issue #2679). With no
+    // active view there is nothing to display anyway.
+    if (m_graphicView == nullptr){
+        return;
+    }
     setWidgetMode(MODE_ENTITY_INFO);
     if (en == nullptr){
         clearEntityInfo();
--- a/librecad/src/ui/main/qc_applicationwindow.cpp
+++ b/librecad/src/ui/main/qc_applicationwindow.cpp
@@ -1324,7 +1324,11 @@ bool QC_ApplicationWindow::doCloseAllFiles(){
             }
             doArrangeWindows(RS2::CurrentMode);
         }
-        qApp->processEvents(QEventLoop::AllEvents, 1000);
+        // Exclude user-input events while closing: doClose() has already detached
+        // the active view, so dispatching a stale queued mouse-move here would run
+        // hover/Quick Info against a torn-down viewport and crash (issue #2679).
+        // Repaints and deferred deletions still need to be processed.
+        qApp->processEvents(QEventLoop::ExcludeUserInputEvents, 1000);
     }
     return false;
 }


ideocentric:
Closing in favor of #2680 — I hadn't seen you'd already opened a shutdown fix for issue #2679 when I put this up. #2680 is the more complete solution (terminal-state event handler, stale-view rejection, and the regression test), so this is redundant. Apologies for the duplicate.

One detail from our diagnosis, in case it's useful for #2680: the specific fault we hit is `doCloseAllFiles()` calling `qApp->processEvents(QEventLoop::AllEvents, 1000)` after `doClose()` has detached the view, which dispatches a queued mouse-move mid-teardown → `LC_QuickInfoBaseData::getRelativeZero()` dereferences the now-null viewport. Passing `QEventLoop::ExcludeUserInputEvents` there is a cheap root-level guard that stops queued user input being delivered during close while still allowing repaints/deferred-deletions — complementary to the terminal-state handling in #2680. Feel free to fold in or ignore if already covered.

Thanks for getting to this so quickly!

Issue #2522, opened 2026-04-06
The color of the block does not depend on the color of the layer


kberkut:
Thanks for making this editor! But it would be great if the color of the block lines changed depending on the color of the layer in which the block is located. Currently, the color of the block lines is always black and white and does not depend on the color of the layer.

Версия: v2.2.1.4
Компилятор: GNU GCC 15.2.1
Собран: Mar  3 2026
Версия Qt: 5.15.18
Версия Boost: 1.89.0
Система: Arch Linux

sand1024:
@kberkut 
well, just to confirm - you mean the color of block lines when the INSERT with that that block is rendered within the document? 

Or the color of entities within the block editor? 

P.S - in most cases, lines in block should have "By Block" color attribute. If one is set, when block is inserted into the drawing, the attributes of the block insert are propagated to inner entities of the block. 

@dxli - it seems in my local version for 2.2.2 colors of blocks works fine, yet as far as I remember I've made some adjustments for similar case. Let me commit my changes and review this issue again then.

kberkut:
Thank you for your response. I seem to have misunderstood how the block properties are set. In my opinion, it would be more logical to name the "By Layer" option, as it would apply the layer's settings.
Currently, I have set the block attributes to "By Block" when editing the block, and the color of the block's lines and fill has changed to the layer's color, which seems illogical from a terminology perspective.

kberkut:
It's still unclear about the display of blocks. I created a block on layer 1. I edited the block by changing its attributes for all three parameters to the value ByBlock. Then, I inserted several instances of the created block into layer 1 from the block list. Next, I used the properties parameter (prop) to specify the location of some blocks on layer 2. When layer 2 is disabled, only blocks with yellow color are displayed. When layer 1 is disabled, all blocks are displayed. Is this behavior incorrect, or am I mistaken?

[test.zip](https://github.com/user-attachments/files/26533572/test.zip)

dxli:
fixed in both 2.2.1 and master branches: set insert pen to ByLayer for correct ByBlock color resolution

PR #2684, opened 2026-07-13
Fix #2681: Dimension line doesn't change layer

### LINKED ISSUE #2681: [BUG] The main line inside a Dimension element does not change of layer

dxli:
## Summary
Fix for issue #2681: The main line inside a Dimension element does not change of layer.

## Root Cause
When changing the layer of a Dimension entity, the dimension line and arrow heads did not change color to match the new layer's color.

The root cause is that dimension child entities (dimension lines, arrows, text) have ByBlock pens and nullptr layers. When resolving ByBlock colors at draw time, the code uses parent->getPen().getColor() (unresolved). If the dimension entity had an explicit color from setPenToActive(), ByBlock children would resolve to that explicit color instead of the new layer's color.

## Fix
Set dimension pen to ByLayer when changing layer, similar to the fix for issue #2522 for Insert entities. This ensures ByBlock children resolve to the parent's layer color at draw time.

The fix is located in RS_Modification::changeAttributes() in rs_modification.cpp and only sets ByLayer for attributes the user isn't explicitly changing.

### DIFF
--- a/librecad/src/lib/modification/rs_modification.cpp
+++ b/librecad/src/lib/modification/rs_modification.cpp
@@ -351,6 +351,17 @@ bool RS_Modification::changeAttributes(
 
         if (data.changeLayer==true) {
             cl->setLayer(data.layer);
+            if (RS_Information::isDimension(cl->rtti())) {
+                if (!data.changeColor) {
+                    pen.setColor(RS_Color(RS2::FlagByLayer));
+                }
+                if (!data.changeWidth) {
+                    pen.setWidth(RS2::WidthByLayer);
+                }
+                if (!data.changeLineType) {
+                    pen.setLineType(RS2::LineByLayer);
+                }
+            }
         }
 
         if (data.changeColor==true) {


vicdoval:
image
I have downloaded the version with this commit  and it still shows the same issue 

[Version: v2.2.1.5-23-g13ee4fad8
Compiler: GNU GCC 11.4.0
Compiled on: Jul 13 2026
Qt Version: 5.15.2
Boost Version: 1.74.0
System: Linux Mint 22.2]

PR #2685, opened 2026-07-13
Fix proposal for library block insert issues on master

qwilvove:
Hello,

Here is a fix proposal for library block insert issues on master.

This is based on commit 7046ed6 by @dxli.

It fixes 2 issues:
- Issue 2578, which is fixed on 2.2.1 but not on master.
- Scale disparities between preview and inserted block (due to unit conversion being performed twice in RS_ActionLibraryInsert).

Best regards,

qwilvove

### DIFF
--- a/librecad/src/actions/dock_widgets/library/rs_actionlibraryinsert.cpp
+++ b/librecad/src/actions/dock_widgets/library/rs_actionlibraryinsert.cpp
@@ -62,12 +62,6 @@ void RS_ActionLibraryInsert::setFile(const QString& file) {
     m_actionData->prev = std::make_unique<RS_Graphic>();
     if (!storage.loadDocument(m_actionData->prev.get(), file, RS2::FormatUnknown)) {
         commandMessage(tr("Cannot open file '%1'").arg(file));
-    } else {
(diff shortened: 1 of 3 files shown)


dxli (approved):

sand1024:
@dxli 
I think it's better to rollback this PR, apply my pending PR and then try to apply this fix (if it will be still relevant, as as far as I remember I've updated copy/paste functionality) over merged version. 
It will be hard to merge functionality from this PR with LibreCAD#2540

qwilvove:
@sand1024
I’m sorry, I hadn’t checked whether your PR was ready to be merged. I fully agree with your approach.

sand1024:
@qwilvove 
I've merged my PR to the master, please grab the last copy, take a look to it and re-apply your changes (if this is still necessary, as as far as I remember I've also made changes in that area). 
In order to merge, I had to override yours fixes due to refactoring, sorry for that. 

Anyway, it is most probably that relevant code could be either in RS_Modification or in LC_CopyUtils. Also, please note that original RS_ActionLibraryInsert class was renamed to LC_ActionBlockLibraryInsert (yet in general it remains the same as before)

PR #2705, opened 2026-07-18
Fix proposal for polyline trim action

qwilvove:
Hello,

I've noticed that polyline trim action is not taken into account after being performed.
The missing of ```return true;``` in ```LC_ActionPolylineTrim::doTriggerModifications``` seems to be the root of the problem.

Here is a PR to fix it.

Best regards,

qwilvove


sand1024:
@qwilvove 
Yes, it looks like you're right with your fix, that was a nasty mistyping bug and it's weird how it was missing.

dxli (approved):

PR #2709, opened 2026-07-25
Fix proposal for build fails with Qt 6.4

qwilvove:
Hello,

When trying to build master with Qt 6.4 (Qt version available on Debian 12), I've noticed that the build fails on some points.

Here is the system config I've tested:

Version: 2.2.2_alpha1-594-g430d423af
Compiler: GNU GCC 12.2.0
Compiled on: Jul 25 2026
Qt Version: 6.4.2
Boost Version: 1.74.0
System: Debian GNU/Linux 12 (bookworm)

I needed to perform some changes on the code in order to build:
- add ```#include <QStringBuilder>``` in files where the syntax ```QString % QString``` is used.
- replace ```QCheckBox::checkStateChanged``` (added in Qt 6.7) by ```QCheckBox::toggled```
- use ```#include <QtGlobal>``` instead of ```#include <qassert.h>``` where ```Q_ASSERT``` is used, since ```<QtAssert>``` seems to have been added in Qt 6.5.
- replace ```class LC_DimStyleItem;``` by ```#include "lc_dimstyleitem.h"``` (not sure if this one is related to Qt, or maybe to GCC)

I've tested those changes with Qt 6.11 and it builds with no problem:
Version: 2.2.2_alpha1-595-g8ee7f32e4
Compiler: GNU GCC 16.1.1
Compiled on: Jul 25 2026
Qt Version: 6.11.1
Boost Version: 1.91.0
System: Arch Linux

Perhaps it is no longer necessary to continue supporting Qt 6.4, but since the project README says that the branch 2.2.2 should build with Qt 6.4, and the changes made here are very minimal, I suppose that it wouldn’t be good to lose that compatibility.

Best regards,

qwilvove

### DIFF
--- a/librecad/src/lib/engine/document/variables/rs_variable.h
+++ b/librecad/src/lib/engine/document/variables/rs_variable.h
@@ -28,6 +28,7 @@
 #define RS_VARIABLE_H
 
 #include <QString>
+#include <QStringBuilder>
 
 #include "rs.h"
 #include "rs_vector.h"
--- a/librecad/src/lib/engine/undo/rs_undo.cpp
+++ b/librecad/src/lib/engine/undo/rs_undo.cpp
@@ -27,9 +27,10 @@
 
 #include "rs_undo.h"
 
-#include<iostream>
-#include <qassert.h>
+#include <iostream>
 #include <unordered_set>
+
+#include <QtGlobal>
 
 #include "rs_debug.h"
 #include "rs_undocycle.h"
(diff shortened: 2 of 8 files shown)


sand1024:
@qwilvove 
Oh, yes, that might be. Actually, I suppose we'd better move to fresher stable Qt version rather than support older ones. 

Another similar direction - adding support of C++ 20 standard instead of 17

Issue #2708, opened 2026-07-25
[BUG] Segmentation fault on Parallel Circle


hapees:
**LibreCAD Crash Report**

Crash Time (epoch s.us): 1784991508.327376
PID: 373139

Signal: 11 (SIGSEGV)
Signal Code: 1
Fault Address: 0x10

```
Call Stack:

./unix/librecad() [0xa75edd]
/lib64/libc.so.6(+0x19fb0) [0x7ff1c7020fb0]
./unix/librecad() [0x6ad405]
./unix/librecad() [0xaece7b]
./unix/librecad() [0x5613fa]
./unix/librecad() [0xa726a4]
./unix/librecad() [0x6f22d8]
./unix/librecad() [0x561242]
./unix/librecad() [0x6959f5]
./unix/librecad() [0x693d93]
./unix/librecad() [0x5a8251]
./unix/librecad() [0xe73d6c]
/lib64/libQt6Widgets.so.6(_ZN7QWidget5eventEP6QEvent+0x218) [0x7ff1c8eb0778]
/lib64/libQt6Widgets.so.6(_ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent+0x8e) [0x7ff1c8e3f92e]
/lib64/libQt6Widgets.so.6(_ZN12QApplication6notifyEP7QObjectP6QEvent+0xf06) [0x7ff1c8e4aa26]
/lib64/libQt6Core.so.6(_ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent+0x168) [0x7ff1c770f7b8]
/lib64/libQt6Widgets.so.6(_ZN19QApplicationPrivate14sendMouseEventEP7QWidgetP11QMouseEventS1_S1_PS1_R8QPointerIS0_Ebb+0x539) [0x7ff1c8e49139]
/lib64/libQt6Widgets.so.6(+0xc6d84) [0x7ff1c8ec6d84]
/lib64/libQt6Widgets.so.6(+0xc983b) [0x7ff1c8ec983b]
/lib64/libQt6Widgets.so.6(_ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent+0x8e) [0x7ff1c8e3f92e]
/lib64/libQt6Core.so.6(_ZN16QCoreApplication15notifyInternal2EP7QObjectP6QEvent+0x168) [0x7ff1c770f7b8]
[...]

PR #2710, opened 2026-07-25
Fix proposal for #2708 - LC_ActionDrawLineParallel crash

qwilvove:
Hello,

Here is a fix proposal for #2708.
The fix is based on what is done in ```LC_ActionDrawLineParallelThrough::doTriggerModifications```.
Maybe the fix is not optimal and a further work is needed on the file.

Best regards,

qwilvove

### DIFF
--- a/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
+++ b/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
@@ -106,14 +106,11 @@ bool LC_ActionDrawLineParallel::doUpdateDistanceByInteractiveInput(const QString
 }
 
 bool LC_ActionDrawLineParallel::doTriggerModifications(LC_DocumentModificationBatch& ctx) {
-    if (m_entity != nullptr) {
-        RS_Creation::createParallel(*m_coord, m_distance, m_numberToCreate, m_entity, false, ctx.entitiesToAdd);
-        return true;
-    }
-    return false;
+    RS_Creation::createParallel(*m_coord, m_distance, m_numberToCreate, m_entity, false, ctx.entitiesToAdd);
+    return true;
 }
 
-void LC_ActionDrawLineParallel::onMouseMoveEvent([[maybe_unused]] const int status, const LC_MouseEvent* e) {
+void LC_ActionDrawLineParallel::onMouseMoveEvent(const int status, const LC_MouseEvent* e) {
     *m_coord = {e->graphPoint}; // copy is needed there!
 
     m_entity = catchAndDescribe(e, RS2::ResolveAll);
@@ -145,8 +142,17 @@ void LC_ActionDrawLineParallel::onMouseMoveEvent([[maybe_unused]] const int stat
     }
 }
 
-void LC_ActionDrawLineParallel::onMouseLeftButtonRelease([[maybe_unused]] int status, [[maybe_unused]] const LC_MouseEvent* e) {
-    trigger();
+void LC_ActionDrawLineParallel::onMouseLeftButtonRelease(const int status, [[maybe_unused]] const LC_MouseEvent* e) {
+    switch (status) {
+        case SetEntity: {
+            if (m_entity != nullptr) {
+                trigger();
+            }
+            break;
+        }
+        default:
+            break;
+    }
 }
 
 void LC_ActionDrawLineParallel::onMouseRightButtonRelease(const int status, [[maybe_unused]] const LC_MouseEvent* e) {
--- a/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
+++ b/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
@@ -106,8 +106,11 @@ bool LC_ActionDrawLineParallel::doUpdateDistanceByInteractiveInput(const QString
 }
 
 bool LC_ActionDrawLineParallel::doTriggerModifications(LC_DocumentModificationBatch& ctx) {
-    RS_Creation::createParallel(*m_coord, m_distance, m_numberToCreate, m_entity, false, ctx.entitiesToAdd);
-    return true;
+    if (m_entity != nullptr) {
+        RS_Creation::createParallel(*m_coord, m_distance, m_numberToCreate, m_entity, false, ctx.entitiesToAdd);
+        return true;
+    }
+    return false;
 }
 
 void LC_ActionDrawLineParallel::onMouseMoveEvent([[maybe_unused]] const int status, const LC_MouseEvent* e) {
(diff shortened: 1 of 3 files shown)


sand1024:
@qwilvove 
well, I'm not sure this is right fix. The actual question is different - why and when m_entity is null and why trigger() is still called for this invalid situation? 

Simply checking for null there is just a fighting the consequent, not the reason. 

Plus - the overall template workflow in base classes that implements action's triggering logic, includes a validation method: 

```
bool RS_PreviewActionInterface::doCheckMayTrigger() {
    return true;
}
```

It is called from the trigger action: 

```
void RS_PreviewActionInterface::trigger() {
    RS_ActionInterface::trigger();
    deletePreview();
    deleteInfoCursor();
    deleteHighlights();
    deleteSnapper();
    if (doCheckMayTrigger()) {
        doTrigger();
    }
    drawSnapper();
    m_graphicView->redraw();
}
```

So, we should rather put that guard check for null into overridden doCheckMayTrigger(), I suppose. 

Another options that leads to call the trigger() is here 

```
void LC_ActionDrawLineParallel::onMouseLeftButtonRelease([[maybe_unused]] int status, [[maybe_unused]] const LC_MouseEvent* e) {
    trigger();
}

```
And here is also a good place to check whether we may do trigger (if m_entity is not null). 

Moreover, the lack of the check in this method is the actual reason of the crash - the user my simply do perform click without selecting entity (and this is a bug). 

So I suppose the right way to fix - is just to add check into onMouseLeftButtonRelease() and, potentially, the guard check into doCheckMayTrigger(). The last one is not needed as the only way to trigger that action is via mouse click, yet if (quite unlikely) in the future some other way of invocation will occur (like command line or so) - that guard will be helpful.

qwilvove:
@sand1024
Thank you very much for this very detailed reply. If it’s not too much trouble, I’d like to try and find a more appropriate fix myself, using your comments as a guide, so that I can learn how to resolve this sort of bug. If you’d prefer to make the fix yourself, that’s fine with me too.

qwilvove:
I've noticed that the action can be triggered by clicking on the centre of the circle. Is this intended ?

librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
```diff
@@ -145,8 +142,17 @@ void LC_ActionDrawLineParallel::onMouseMoveEvent([[maybe_unused]] const int stat
     }
 }
 
-void LC_ActionDrawLineParallel::onMouseLeftButtonRelease([[maybe_unused]] int status, [[maybe_unused]] const LC_MouseEvent* e) {
-    trigger();
+void LC_ActionDrawLineParallel::onMouseLeftButtonRelease(const int status, [[maybe_unused]] const LC_MouseEvent* e) {
+    switch (status) {
```
sand1024, line 146: @qwilvove 
well, I'm not really sure that check for the status is needed there (while the check fully valid!). It's nice if we'll add some new status of the action later on, yet most probably adding new status will require update of all action's code and from this point of view it is a little over-defense as for me - but it's ok.

sand1024:
> I've noticed that the action can be triggered by clicking on the centre of the circle. Is this intended ?

well, it might that catchEntity() will find the entity - be depending on your snap settings, radius of entity and so on. For example - you may try to set, say, snap entity (or snap to endpoints etc. - with disabled "snap to grid" and disabled "snap free" - so the nearest entity will be caught on mouse move (and it is not necessary that mouse cursor should be over that entity).

dxli:
@qwilvove Thanks for this, and sorry it sat so long. Short version: the crash it was written for is
gone, but the PR is still worth merging for two other reasons, so I've pushed a commit with tests
to your branch and I plan to merge it.

**Why it is no longer a crash fix.** #2715 landed on master and fixed #2708 one layer down —
`RS_Creation::createParallel` now returns early on a null entity (`rs_creation.cpp:144`), with
regression coverage in `rs_creation_null_entity_tests.cpp`. The segfault no longer reproduces
without your change.

**Why I still want it.** Your guard prevents two things #2715 cannot reach, because both are
consequences of `trigger()` running at all:

1. **A no-op click destroys redo history.** `trigger()` → `doTrigger()` →
   `RS_Document::undoableModify` constructs `LC_UndoSection` (`rs_document.cpp:103`) *before* the
   modification lambda runs, so the undo cycle opens before anything knows the click produced
   nothing. `RS_Undo::startUndoCycle` then prunes everything past `m_redoPointer` and frees the
   undone entities at cycle start (`rs_undo.cpp:81-99`) — the `hasUndoable()` gate at the end only
   stops the empty *step* from being recorded, far too late. On master today: draw something,
   Ctrl+Z, click empty space with the parallel tool, and your redo is silently gone. With your
   guard it survives.

2. **Clicking while entering the count creates un-previewed geometry.** `m_entity` is refreshed on
   every mouse move regardless of status (`lc_action_draw_line_parallel.cpp:116`, above the status
   switch), so before your change a click while the prompt read *"Enter number:"* created parallels
   the user never saw — the `number` command calls `deletePreview()` on the way in. Your
   `status == SetEntity` check makes that a proper no-op.

**@sand1024** — on your suggestion to also guard in `doCheckMayTrigger()`: the hook does exist in
this chain (`rs_previewactioninterface.h:250`, consulted inside `trigger()` before `doTrigger()`),
so it would work. I left it out for two reasons. It is redundant: tracing every route to
`trigger()` for this action, `onMouseLeftButtonRelease` is the only one — real clicks
(`rs_actioninterface.cpp:189`) and the relative-input widget's synthetic release
(`rs_previewactioninterface.cpp:943`) both funnel through it, `doProcessCommand` never triggers and
always returns `SetNumber` to `SetEntity`, there is no `onCoordinateEvent` override, no Enter-key
trigger, and `doInitWithContextEntity` is an empty stub. And on its own it would be worse: by the
time the hook runs, `trigger()` has already called `deletePreview()`/`deleteHighlights()`, so a
hook-only veto still throws away the hover preview on a rejected click. The call-site guard you
wrote never enters `trigger()` at all, which is the better place for it.

**What I pushed** (`8bd9a4a`, on top of your two commits, which I left untouched):
`librecad/src/lib/creation/tests/lc_action_draw_line_parallel_tests.cpp`, three cases driving the
real mouse handlers on a headless action — empty-space click preserves redo, `SetNumber` click
creates nothing, and a positive control that a caught entity still produces an undoable parallel so
the guard cannot over-suppress. Measured both ways: 11/11 assertions with your guard, 5 failures
across the first two cases without it. Full suite on the merged result: 12955 assertions in 736
test cases, exit 0. The test registration is anchored after `dwg_buffer_round_trip_tests.cpp`
rather than the block's natural position, because this branch predates #2715/#2728 and the natural
anchor conflicts with entries master has gained since; `git merge-tree` confirms the chosen spot
merges clean.

I'll merge with a message describing it as the action-level guard it now is rather than a crash
fix. Thanks for filing it and for reworking it to sand1024's review — the second version is the
right shape, and it turned out to be fixing something none of us had noticed.

Follow-up: the redo wipe is not specific to this action — any action whose `trigger()` reaches
`doTrigger()` with nothing to do destroys pending redo, because the pruning happens at
`startUndoCycle` rather than at commit. The general engine-level fix is now up as #2739; it moves
the prune to `endUndoCycle`'s non-empty branch, with regression tests for the empty-cycle, nested,
and shared-undoable cases. Your guard here remains the right call regardless — it also keeps the
hover preview alive on a rejected click, which no engine-level fix can do.

Issue #2636, opened 2026-06-18
[FEATURE] Font : wqy-unicode


city-1921:
Is LibreCAD 2.2.1.X 2.2.2.X considering bundling wqy-unicode.lff by default?

This is very helpful for fixing garbled text issues.

Thanks.

dxli:
WenQuanYi fonts are available at [wqy-unicode.lff](https://dokuwiki.librecad.org/lib/exe/fetch.php/community:fonts:wqy-unicode.zip)

The font files were removed from standard packages due to their large sizes.

Recently @swif proposed a way to reduce file sizes for lff fonts:

https://github.com/LibreCAD/LibreCAD/issues/2579

We will see whether we can regenerate the fonts with a reasonable size and include them in standard releases.

dxli:
@city-1921 ,

Using @siwf's python-lff, I updated LibreCAD's ttf2lff program, and the latest WenQuanYi lff font file from the official WenQuanYi ttf is much smaller in size and supports more characters:

[wqy-unicode.zip](https://github.com/user-attachments/files/29193154/wqy-unicode.zip)

city-1921:
@dxli 

The [wqy-unicode.lff] downloaded from the first link has a file size of 41.4 MB.

The file downloaded from the second link has a size of 46.5 MB.

The file seems to have gotten bigger, not smaller.

But that's fine.

Thank you.

siwf:
@dxli new *wqy-unicode.lff* is malformed,
char code line has additional *#*, i.e. for space char: `[#0020]`

As *wqy-unicode.lff* from wiki has 2 digit precision,
so tuning new *wqy-unicode.lff* with 2 digit precision too,
they have similar size after tuning.

compare of tuning result for two *wqy-unicode.lff*:

| File Name | Glyph | No Comment | Zero Vector | Round Vertex | Merge Path | Nested Char | Size | Tuned | Ratio |
|:-----|------:|----------:|-----------:|------------:|----------:|-----------:|-----:|------:|------:|
| wqy-unicode.lff (wiki) | 34599 |  | 943674 |  | 61 | 1355 | 43436980 | 33760295 | 77.7% |
| wqy-unicode.lff (new) | 34600 |  | 3 | 121347 | 62 |  | 48803860 | 33857917 | 69.4% |

PR #2711, opened 2026-07-26
LFF: accept optional '#' and leading zeros in glyph codes

dxli:
Part of #2636. Addresses the malformed-glyph-code problem @siwf reported on the
regenerated `wqy-unicode.lff`.

## The problem

`ttf2lff` emitted glyph headers as `[#0020]`, which the LFF reader rejected — the
leading `#` was not accepted, so those glyphs were skipped.

## Reader

Glyph headers are now parsed with an anchored `^\[#?([0-9A-Fa-f]{1,6})\]`, so all
four spellings load to the same character:

```
[0020]   [20]   [#0020]   [#20]
```

Widening the code to 6 hex digits also admits the full Unicode range up to
U+10FFFF. The previous 5-digit cap silently dropped any glyph in planes 15–16.

The regex is anchored and only applied to lines already known to start with `[`
(metadata lines starting with `#` are handled on a separate branch), so `#`
comments are unaffected.

## Writers

Both writers now emit the minimal form, since leading zeros are optional:

* `ttf2lff`'s `formatCharCode` drops the `setw(4)` padding — `[#41]`, not `[#0041]`.
* `RS_FilterLFF::fileExport` strips leading zeros from the `[..]` glyph header and
  from `C..` references.

The in-memory `RS_Block` name keeps its zero-padded `[hhhh] c` form, because the
font viewer (`lc_fontfileviewer.cpp`, `getName().mid(1,4)`) and the block list
depend on that fixed width. Only the on-disk file is compacted. `rs_filtercxf.cpp`
is deliberately untouched — the CXF parser requires exactly 4 hex digits.

## Compatibility

Existing fonts are unaffected: all 47 shipped `.lff` files use the padded `[0020]`
form and still parse identically. The change is additive — it only widens what the
reader accepts.

## Testing

* `librecad_lib` and `ttf2lff` build clean against current master.
* All 47 shipped `.lff` fonts parse unchanged.
* Verified `[0020]` / `[20]` / `[#0020]` / `[#20]` all load to U+0020.
* Converted a 3,381-glyph font: every header emitted compact, zero padded headers
  remaining.

### DIFF
--- a/librecad/src/lib/engine/document/fonts/rs_font.cpp
+++ b/librecad/src/lib/engine/document/fonts/rs_font.cpp
@@ -61,16 +61,18 @@ namespace {
         return (okay) ? QString::fromUcs4(&ucsCode, 1) : QString::fromUcs4(&invalidCode, 1);
     }
 
-    // Extract the unicode char from LFF font line
+    // Extract the unicode char from LFF font line, e.g. "[0041]" or "[#0041]"
+    // (the '#' is optional, so files written by either convention load the same way).
+    // Up to 6 hex digits so the full Unicode range (to U+10FFFF) is accepted.
     std::pair<QString, bool> extractFontChar(const QString& line) {
         // read unicode:
-        static QRegularExpression regexp("[0-9A-Fa-f]{1,5}");
+        static QRegularExpression regexp("^\\[#?([0-9A-Fa-f]{1,6})\\]");
         const QRegularExpressionMatch match = regexp.match(line);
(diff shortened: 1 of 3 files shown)


dxli:
Split out of the original combined PR. Companion PRs, both independent of this one at the file level:

- #2712 — `ttf2lff`: `-t` implies `-d 2` (the size half of #2636)
- #2713 — the CJK font files themselves; **requires this PR** to merge first, since those fonts use compact glyph codes.

PR #2715, opened 2026-07-27
Fix four null-entity crashes, including the parallel-tool segfault (#2708)

### LINKED ISSUE #2708: [BUG] Segmentation fault on Parallel Circle

dxli:
Fixes #2708

Four null-entity crashes, each with a demonstrated path to it. Every hunk here was
checked against two rules: it must fix a reachable crash, and it must be the
smallest change that does. Several earlier changes failed one or the other and
have been removed — see *Removed after validation* at the end.

## 1. The reported crash — `RS_Creation::createParallel`

`LC_ActionDrawLineParallel::onMouseMoveEvent` stores
`m_entity = catchAndDescribe(...)` on every move, which is `nullptr` whenever the
cursor is not within catch distance of an entity. The *preview* use is guarded;
`onMouseLeftButtonRelease` calls `trigger()` unconditionally, and the action does
not override `doCheckMayTrigger()` — whose base returns `true`. So the null reaches
`createParallel` and the virtual `e->rtti()` in its dispatch switch runs on a null
pointer. **Left-click on empty canvas with the Parallel tool active.**

Both triggers described in the issue reduce to this path. A circle that is
genuinely caught but admits no parallel does *not* crash — it produces an empty
batch, and `RS_Undo::endUndoCycle` discards the empty cycle.

## 2. `RS_Creation::createParallelThrough`

Same file, one function up, reachable from the same tool family with no entity ever
picked: in Parallel-through, type `number`, right-click, move the mouse.
`onMouseRightButtonRelease` clears `m_entity`, then `initPrevious()` — which is
`init(status - 1)` — lands on `SetPos`, whose *preview* call is unguarded. (The
*trigger* call at line 89 is guarded; the preview one at line 122 was not.)

## 3. Three dimension-tool SIGSEGVs — six lines

`RS_ActionDimAngular::setData()` reads `m_line1`/`m_line2` unconditionally, and
`LC_ActionCircleDimBase`'s `SetPos` *release* handler reads `m_entity->getCenter()`
where the matching *move* handler was already guarded. Both members default to
`nullptr`; both accessors are virtual, so this is a vtable load from address 0.

Neither action gates its `text` command on a status — and `getAvailableCommands()`
actively **advertises** `text` at those statuses, so this is a documented path.
From a freshly started tool: type `text`, right-click (`initPrevious()` lands on
`SetPos`), then simply move the mouse. Escape reaches it too, since
`RS_EventHandler::back()` synthesises a right-button release. A second variant
exists from `SetLine2`, with the first line picked and the second not.

`LC_ActionCircleDimBase` is the base of both `RS_ActionDimRadial` and
`RS_ActionDimDiametric`, so one missing check crashes three tools.

## 4. `RS_Modification::trim()` casting a container limit entity

`trim()` warned that a non-atomic limit entity was unsupported and then used it
anyway, casting it to `RS_AtomicEntity` at three sites. `trimStartpoint()`,
`trimEndpoint()`, `getTrimPoint()` and `prepareTrim()` are declared **only** on
`RS_AtomicEntity`, so the calls went through foreign vtable slots.

Reachable by hovering: `LC_ActionModifyTrim` resolves its limit entity with
`ResolveAllButTextImage` and filters only `EntityPolyline` — its `isAtomic()` check
is present but **commented out** in the source. `RS_Hatch::doGetDistanceToPoint`
ignores the resolve level and returns `this` whenever the cursor is inside a
boundary loop, and Trim-Two calls `previewTrim()` on every mouse move.

The casts are gated on `isAtomic()` rather than the entity being refused: a
container is a legitimate boundary for a one-sided trim, and `findIntersection()`
walks container children specifically to support that. Replacing the three gates
with a single early return would be smaller but would **regress** — on `master` the
atomic side is trimmed correctly and only the limit side is undefined, so an early
return turns a working one-sided trim into a no-op.

## 5. Five context-entity dereferences — one condition each

Twelve `doInitWithContextEntity` overrides hand-copy a block that replaces the
context entity — which `init()` guarantees non-null — with
`polyline->getNearestEntity()`, which returns null when the container has no
eligible child. Five then dereference the result; the other seven route it through
`isLine()`/`isArc()`, which null-test internally, and are genuinely safe.

**Being precise about reachability, because it is weaker than the others.** Such
polylines do exist — `rs_filterdxfrw.cpp` explicitly adds a vertex-less `POLYLINE`
to the document, and a one-vertex polyline also yields zero children. All five
actions are on the polyline context menu. But a zero-segment polyline reports a
distance of `RS_MAXDOUBLE`, so `catchContextEntity`'s 32-pixel test only admits it
in a document where nothing else is nearer *and* at a zoom factor around 3e-9.
Nothing clamps the factor that low, so it is not impossible — but it is not the
casual path the other four are. Included because the fix is one condition per site
and the dereference is real.

## Testing

Six cases across two files. `RS_Creation` and `RS_Modification` are free functions
and statics, so both are testable with no document or GUI fixture; neither had any
coverage.

The positive cases carry real weight: without them, a guard mutated into an
unconditional early return would pass every null test while silently disabling the
tool, and the trim gate could be made over-broad without anything noticing.

The calls are made plainly rather than wrapped in a no-throw assertion — the
failure is a segmentation fault, not an exception, and Catch2's fatal-signal
handler is what reports it. Verified in both directions: reverting `rs_creation.cpp`
makes the suite abort with `SIGSEGV - Segmentation violation signal`, exit 139;
reverting the `trim()` gates makes the container test do the same.

Full suite: no new failures. (The shapefile tests fail on `master` for an unrelated
reason — #2716.)

## Removed after validation

An earlier revision of this PR was 38 files. Auditing every hunk against
"necessary for a reachable crash, and minimal" removed a good deal of it:

* **Guards in `createParallelLine`/`Arc`/`Circle`/`SplinePoints`.** Each has exactly
  one caller — the switch inside `createParallel`, which dereferences the entity in
  the switch expression before dispatching, and is now guarded itself. Beyond being
  unreachable, replacing `Q_ASSERT` there would trade the only mechanism that flags
  a future contract violation for a silent no-op. They keep their asserts.
* **`doTriggerModifications` returning `!ctx.entitiesToAdd.isEmpty()`.** Tracing
  every consumer, an empty batch behaves identically whether success is true or
  false, and `ctx.success` is never read on that path. It reads better and fixes
  nothing.
* **Deduplicating the twelve copied resolution blocks** behind a shared resolver —
  a refactor that adds a virtual to the base of ~180 action subclasses and fixes
  nothing the five conditions above do not.
* **The action status machines** that permit these states in the first place.
  Correcting them changes right-click behaviour and deserves separate review.
* Assorted `Q_ASSERT` conversions with no reachable null caller, and a
  `getCurrentAction()` guard with no demonstrated repro.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2395,6 +2395,7 @@ if (BUILD_TESTS)
         librecad/src/lib/math/tests/rs_math_tests.cpp
         librecad/src/lib/math/tests/lc_quadratic_tests.cpp
         librecad/src/lib/creation/tests/rs_creation_null_entity_tests.cpp
+        librecad/src/lib/modification/tests/rs_modification_trim_tests.cpp
         librecad/src/lib/filters/tests/dwg_buffer_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_entity_encode_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_header_encode_round_trip_tests.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2394,6 +2394,7 @@ if (BUILD_TESTS)
         librecad/src/lib/engine/document/entities/tests/rs_text_bidi_tests.cpp
         librecad/src/lib/math/tests/rs_math_tests.cpp
         librecad/src/lib/math/tests/lc_quadratic_tests.cpp
+        librecad/src/lib/creation/tests/rs_creation_null_entity_tests.cpp
         librecad/src/lib/filters/tests/dwg_buffer_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_entity_encode_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_header_encode_round_trip_tests.cpp
--- a/librecad/src/actions/drawing/draw/dimensions/lc_actioncircledimbase.cpp
+++ b/librecad/src/actions/drawing/draw/dimensions/lc_actioncircledimbase.cpp
@@ -128,6 +128,9 @@ void LC_ActionCircleDimBase::onMouseLeftButtonRelease(const int status, const LC
             break;
         }
         case SetPos: {
+            if (m_entity == nullptr) {
+                break;
+            }
             RS_Vector snap = e->snapPoint;
             snap = getSnapAngleAwarePoint(e, m_entity->getCenter(), snap);
             fireCoordinateEvent(snap);
--- a/librecad/src/actions/drawing/draw/dimensions/rs_actiondimangular.cpp
+++ b/librecad/src/actions/drawing/draw/dimensions/rs_actiondimangular.cpp
@@ -337,6 +337,9 @@ int RS_ActionDimAngular::determineQuadrant(const double angle) const {
  */
 bool RS_ActionDimAngular::setData(const RS_Vector &dimPos, const bool calcCenter /*= false*/){
     bool result = false;
+    if (m_line1 == nullptr || m_line2 == nullptr) {
+        return result;
+    }
     if (m_line1->getStartpoint().valid && m_line2->getStartpoint().valid){
 
         if (!m_center.valid || calcCenter){
--- a/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
+++ b/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
@@ -107,9 +107,7 @@ bool LC_ActionDrawLineParallel::doUpdateDistanceByInteractiveInput(const QString
 
 bool LC_ActionDrawLineParallel::doTriggerModifications(LC_DocumentModificationBatch& ctx) {
     RS_Creation::createParallel(*m_coord, m_distance, m_numberToCreate, m_entity, false, ctx.entitiesToAdd);
-    // Nothing to commit when no parallel is possible: no entity under the cursor,
-    // or an inner circle parallel whose distance exceeds the radius.
-    return !ctx.entitiesToAdd.isEmpty();
+    return true;
 }
 
 void LC_ActionDrawLineParallel::onMouseMoveEvent([[maybe_unused]] const int status, const LC_MouseEvent* e) {
--- a/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
+++ b/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel.cpp
@@ -107,7 +107,9 @@ bool LC_ActionDrawLineParallel::doUpdateDistanceByInteractiveInput(const QString
 
 bool LC_ActionDrawLineParallel::doTriggerModifications(LC_DocumentModificationBatch& ctx) {
     RS_Creation::createParallel(*m_coord, m_distance, m_numberToCreate, m_entity, false, ctx.entitiesToAdd);
-    return true;
+    // Nothing to commit when no parallel is possible: no entity under the cursor,
+    // or an inner circle parallel whose distance exceeds the radius.
+    return !ctx.entitiesToAdd.isEmpty();
 }
 
 void LC_ActionDrawLineParallel::onMouseMoveEvent([[maybe_unused]] const int status, const LC_MouseEvent* e) {
--- a/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel_through.cpp
+++ b/librecad/src/actions/drawing/draw/line/lc_action_draw_line_parallel_through.cpp
@@ -77,6 +77,11 @@ void LC_ActionDrawLineParallelThrough::doInitWithContextEntity(RS_Entity* contex
         const auto polyline = static_cast<RS_Polyline*>(contextEntity);
         entity = polyline->getNearestEntity(clickPos);
     }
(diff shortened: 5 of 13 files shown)


sand1024:
@dxli 
it's more than verbose fix for that small issue. The code looks confident - yet most of reasoning there is useless, it's was just a small logical issue that does not requires so large code changes. 

Please refer to the discussion of the fix there: 
https://github.com/LibreCAD/LibreCAD/pull/2710


Please do not believe AI code without close verification of it, it generates lot's of garbage.

dxli:
This PR previously carried 38 files. Investigating #2708 surfaced several adjacent problems and folding them in made a 4-line fix unreviewable, so they are now proposed separately on their own merits:

- #2720 — `RS_Modification::trim()` casting a container limit entity to `RS_AtomicEntity`. Silent document corruption rather than a crash, reachable by hovering in Trim-Two. The most serious of them.
- #2719 — three release SIGSEGVs in the dimension actions (`RS_ActionDimAngular`, and `LC_ActionCircleDimBase` which Radial and Diametric share). Six lines.
- #2721 — five unchecked dereferences of a polyline segment resolved from the context menu.

All three are independent of this PR and of each other.

PR #2720, opened 2026-07-28
RS_Modification::trim: stop casting a container limit entity to RS_AtomicEntity

dxli:
`RS_Modification::trim()` warns that a non-atomic limit entity is unsupported and
then uses it anyway.

Found while investigating #2708. Split out because it is a different kind of bug
and, unlike the others, its failure mode is silent.

## The bug

```cpp
if (both && !limitEntity->isAtomic()) {
    RS_DEBUG->print(RS_Debug::D_WARNING, "RS_Modification::trim: limitEntity is not atomic");
}
// ...proceeds anyway:
return trim(limitCoord, static_cast<RS_AtomicEntity*>(limitEntity), ...);
```

Three sites cast the limit entity to `RS_AtomicEntity`. But `trimStartpoint()`,
`trimEndpoint()` and `getTrimPoint()` are declared **only** on `RS_AtomicEntity` —
they do not exist on `RS_Entity`. Calling them through a container pointer
dispatches to whatever occupies the same vtable slot on that container. On
`RS_Hatch`, one of those slots holds `removeEntity`.

So this does not fault on a null pointer. It calls a *wrong virtual on a real
object*, which corrupts the document rather than crashing — which is why it has
gone unnoticed.

## Reachable by hovering

* `LC_ActionModifyTrim` catches its limit entity with `RS2::ResolveAllButTextImage`,
  which deliberately returns `RS_Text` and `RS_MText` as whole containers.
* `RS_Hatch::doGetDistanceToPoint` returns `this` for a pattern hatch, so
  `catchEntity` hands the container back.
* The action rejects only `rtti() == RS2::EntityPolyline`. Its `isAtomic()` check
  is present but **commented out** in the source.
* Trim-Two calls `previewTrim()` on **every mouse move**, so no click is needed.

## The fix

A container *is* a legitimate boundary for a one-sided trim — `findIntersection()`
has an explicit branch for it — so refusing the entity outright would regress real
behaviour. Instead the three casts are gated on `isAtomic()`:

* trimming *the limit entity itself* is declined, and the existing warning says so;
* trimming *against* it still works exactly as before.

## Testing

New `rs_modification_trim_tests.cpp`, covering both directions:

* a container limit entity must not produce a trimmed second entity;
* an ordinary line limit must still trim both, so the fix is not over-broad.

The second case matters — without it, gating the casts too aggressively would pass
the first test while quietly breaking Trim-Two.

Builds clean, tests pass. The warning is visible in the test output, confirming the
container path is genuinely exercised.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2394,6 +2394,7 @@ if (BUILD_TESTS)
         librecad/src/lib/engine/document/entities/tests/rs_text_bidi_tests.cpp
         librecad/src/lib/math/tests/rs_math_tests.cpp
         librecad/src/lib/math/tests/lc_quadratic_tests.cpp
+        librecad/src/lib/modification/tests/rs_modification_trim_tests.cpp
         librecad/src/lib/filters/tests/dwg_buffer_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_entity_encode_round_trip_tests.cpp
         librecad/src/lib/filters/tests/dwg_header_encode_round_trip_tests.cpp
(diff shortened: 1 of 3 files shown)


dxli:

PR #2728, opened 2026-08-02
Fix Break/Divide leaving degenerate entities behind (#2700)

### LINKED ISSUE #2700: [BUG] Break-Divide - Shift Delete modifier leaves points behind
### Expected Behavior
For items to just delete without leaving behind points.

### Observed Behavior
I noticed an issue when recently testing the Break/Divide tool again.  When using the SHIFT modifier which is linked to a delete action, when this is performed on an item, points are left behind where that item used to be.

### Steps To Reproduce
-Draw a 2 point line
-Activate Break/Divide tool
-Hold down SHIFT key and LMB click on that line to delete it
-After it deletes you can see the two points left behind at the beginning and end of that line
-If cannot see them then try doing a window drag selection over the area to select where those 2 line points were.

### Operating System and LibreCAD Version
Version: LibreCAD-2.2.2_alpha1-587-g08fe3d360.AppImage
Compiler: GNU GCC 13.3.0
Compiled on: Jul 14 2026
Qt Version: 6.4.2
Boost Version: 1.83.0
System: Debian GNU/Linux 13 (trixie) (x11)

dxli:
Fixes #2700 — SHIFT-clicking a standalone line in Break/Divide deletes it but leaves two points behind.

## Cause

`LC_Division::SegmentDisposition` can say the selected segment runs to the entity's start, to its end, or lies between two intersections. It cannot say **"the segment is the entire entity"** — which is exactly what SHIFT asks for. So the five branches that SHIFT enables report `SEGMENT_INSIDE` while handing back the entity's own endpoints:

```
lc_division.cpp:57    arc,    no intersections
lc_division.cpp:84    circle, no intersections
lc_division.cpp:112   line,   no intersections
lc_division.cpp:273   line,   intersections only on its own endpoints
lc_division.cpp:395   arc,    intersections only on its own endpoints
```

The action then computes the *remaining* segments as the complement of the selected one. For a whole-entity selection that complement is empty — but the code emits it anyway:

| | emitted | what the user sees |
| --- | --- | --- |
| line | `(start, start)` and `(end, end)` | two zero-length lines: the reported points, selectable by window drag |
| arc | two zero-sweep arcs | invisible leftovers that take a **full-circle** bounding box, since `RS_Math::isAngleBetween` is true when `angle1 == angle2` |
| circle | complement of `0..2π` is another full turn | delete appears to do nothing; the circle becomes a full-sweep arc |
| preview | same guards drive the markers and info cursor | SHIFT-hover paints and announces division points on the entity's own endpoints |

The last two rows are the same bug, not extras: one missing enum value, four symptoms.

## Fix

Add the missing value. The five branches report `SEGMENT_ENTIRE`, and each `createEntitiesForX` reads it once into a local that gates all five consumers — snap segment, remaining segments, preview markers, info cursor, trigger.

```diff
-        SEGMENT_TO_END // snap is between end point of entity and intersection point
+        SEGMENT_TO_END, // snap is between end point of entity and intersection point
+        SEGMENT_ENTIRE // the entire entity is the segment, it has no intersections outside its own endpoints
```

Every other disposition keeps its exact previous behaviour, so break/divide by intersections is unaffected by construction rather than by a tolerance test.

`CircleSegmentData` had no disposition field and gains one; the circle paths that don't set it are the ones bounded by intersections, hence the `SEGMENT_INSIDE` default. `ArcSegmentData::segmentDisposition` was an `int` while its two sibling structs used the enum — with a fourth value in play that inconsistency stops being cosmetic, so it becomes `SegmentDisposition` here.

**Why not detect it from geometry.** My first attempt guarded the emission sites with `isMeaningfulDistance` / `RS_ArcData::isValid()`. That proxy fails both ways: a **reversed** arc touched only at its endpoints yields two remainders that are exact copies of the arc — valid geometry no degeneracy test can reject — and the same test can fire on a segment that is *not* the whole entity, where treating it as a full removal drops real geometry. Only `LC_Division` knows which case it is.

**The trigger has to be told too.** `doCheckMayTrigger()` refuses an empty `entitiesToCreate`, and the entity is removed only inside `doTriggerEntitiesPrepare()`. Suppressing the segments alone turns the bug into *"SHIFT-click refuses and the line stays"*. So the three helpers **return** whether the entity was fully consumed, and the check accepts an empty list in that case only — a locked entity or a click on an endpoint still refuses. Returning it keeps those helpers `const` instead of having them reach back into the action to set a flag.

With **Remove selected** cleared, the answer is false, so SHIFT on a whole entity is a no-op rather than an identity delete-and-recreate.

**Blast radius.** `segmentDisposition` is written only in `lc_division.cpp` and read only in `lc_action_modify_break_divide.cpp`; no `switch` anywhere. The only other `LC_Division` consumer, `LC_ActionInteractivePickDistance` (`:123`, `:223`, which does pass `allowEntireLine=true`), contains zero occurrences of it — it reads only `snapSegmentStart`/`snapSegmentEnd`, unchanged here.

## Verification

Built with `-DBUILD_TESTS=ON` and driven headlessly through a Catch2 harness modelled on `librecad/src/lib/gui/tests/rs_graphicview_close_tests.cpp`, constructing a real `LC_ActionModifyBreakDivide`:

| case | before | after |
| --- | --- | --- |
| SHIFT-delete isolated line | 2 zero-length entities | 0 |
| SHIFT-click end-to-end (`doOnLeftMouseButtonRelease`) | line deleted, 2 points remain | line deleted, nothing remains |
| SHIFT-delete isolated arc | 2 zero-sweep arcs | 0 |
| SHIFT-delete isolated circle | full-turn arc | 0 |
| SHIFT-delete reversed arc touched only at endpoints | 2 duplicate copies | 0 |
| SHIFT-hover preview on whole entity | 2 spurious markers | 0 |
| ordinary divide, crossed line | 2 real remainders | unchanged |
| ordinary divide, crossed circle | 2 real arcs | unchanged |

8/8 pass; the last two exist to catch an over-broad fix.

### Committed coverage

`librecad/src/lib/modification/tests/lc_division_tests.cpp` pins both halves of the fix.

*Classification* — that `LC_Division` reports `SEGMENT_ENTIRE` from all five branches.

*Consumer* — that the action acts on it. Three `createEntitiesForX()` are driven through a
test-local subclass; a subclass is needed either way, because `m_alternativeActionMode` is the flag
SHIFT sets, it is `protected`, and it has no setter, so without it `allowEntire*` stays false and
`LC_Division` never returns a whole-entity segment. **No production code and no CMakeLists change** —
the cases are appended to the file this branch already adds and registers.

Teeth, measured rather than assumed:

| mutation | result |
| --- | --- |
| delete all three `SEGMENT_ENTIRE` guards | 7 assertions fail; bounded-break cases stay green |
| delete the line / circle / arc guard individually | 3 / 2 / 2 fail |
| revert `doCheckMayTrigger`'s acceptance of whole-entity removal | exactly 1 fails — the trigger-acceptance case |
| report removal despite **Remove selected** off, per entity kind | exactly 1 fails each — the line, circle, and arc keep-selected sections |
| accept the keep-selected no-op as a removal in `doCheckMayTrigger` | exactly 1 fails — the trigger keep-selected case |
| revert the tangent-circle gate to `empty()` | 2 fail — the tangency classification and its action-level case |

The removal-despite-keep rows are the data-loss class: they would delete the entity the user asked
to keep. Every row was rebuilt and run, not reasoned.

The trigger-path test reaches the protected `m_triggerData` through the test subclass via
`decltype` — the nested `TriggerData` type stays private and no production visibility changes.
`rs_graphicview_close_tests.cpp` now uses the same reuse-or-create `QApplication` form as the
other two test files: its old by-value static constructed a second instance whenever another file
created the application first (silently tolerated on release Qt, an abort on assert-enabled Qt).

Full suite: **13006 assertions in 743 test cases, exit status 0**, in default order and under
randomised orderings. Note `BUILD_TESTS` is off in CI, so this is developer-local coverage, not a
gate.

## Addressed after review

The later commits close what the deep review found: the trigger acceptance and all three
keep-selection guards are mutation-pinned; the keep-selected no-op reports through its own string
instead of `"Invalid entity selected"` (`createEntitiesForX` returns a three-value
`SegmentCreation`, since the bool could not distinguish the no-op from a genuine refusal); a circle
touched by a single tangency now takes the whole-entity branch — one boundary point divides a
closed curve no more than none does, and the gate counts *distinct* points because the solver may
report a tangency once or as two coincident points; the pre-existing `TriggerData` leak on a
refused click is fixed; bounded-break remainders are asserted to have nonzero extent (bounded above
too for arcs, since a zero-sweep arc reports a full-turn length); and the endpoint-only arc case
runs both orientations.

The final commit closes the hygiene items: `LC_Division::find*` return `std::unique_ptr` (the two
`lc_actioninteractivepickdistance.cpp` call sites drop their manual deletes; that action never reads
`segmentDisposition`, so it is unaffected by `SEGMENT_ENTIRE` by construction), the three
`createEntitiesForX` are `[[nodiscard]]` with explicit discards at the preview sites, the dead
`LineSegmentData::snap` field is gone, and the preview path is mutation-pinned - restricting the
guard to the trigger fails exactly the new whole-entity preview assertion.

## Merge note

Please do not squash: the commit history is what keeps the ~27 semantic lines findable in the
455-line anchor diff — fix, behavior-preserving restyle, and tests are separate commits.

## Deliberately not in this PR

Found while working here, all pre-existing and none needed for the fix — happy to send any as separate PRs:

- The whole-entity SHIFT preview shows only the hover highlight. That is accurate - the old markers
  announced division points that did not exist - but whether deletion deserves a stronger visual
  affordance is a design choice, left open.
- `RS_Settings::init` is not idempotent (a second call leaks and swaps the singleton). Benign today;
  production infrastructure beyond this change's subject.

Note on the `QApplication` obstacle mentioned in earlier discussion: the trigger is not a second
`QApplication`-using file, and not a second `QApplication` being *constructed* — it is a second
`~QApplication` *running* at exit. `rs_graphicview_close_tests.cpp:43` holds a by-value static and
therefore owns the one destructor this process gets. A new file that reuses-or-heap-allocates and
never deletes is safe, which is what the tests here do; three such files now coexist at exit 0.

### DIFF
--- a/librecad/src/lib/modification/lc_division.cpp
+++ b/librecad/src/lib/modification/lc_division.cpp
@@ -48,7 +48,7 @@ LC_Division::LC_Division(RS_EntityContainer* entityContainer) : m_container{enti
  * @param allowEntireArcAsSegment
  * @return segment information
  */
-LC_Division::ArcSegmentData* LC_Division::findArcSegmentBetweenIntersections(const RS_Arc* arc, const RS_Vector& snap,
+std::unique_ptr<LC_Division::ArcSegmentData> LC_Division::findArcSegmentBetweenIntersections(const RS_Arc* arc, const RS_Vector& snap,
                                                                              const bool allowEntireArcAsSegment) {
     ArcSegmentData* result = nullptr;
     // detect all intersections
@@ -66,7 +66,7 @@ LC_Division::ArcSegmentData* LC_Division::findArcSegmentBetweenIntersections(con
         // determine selected segment edges
         result = findArcSegmentEdges(arc, snap, allIntersections, allowEntireArcAsSegment);
     }
-    return result;
+    return std::unique_ptr<ArcSegmentData>{result};
 }
 
 namespace {
@@ -93,7 +93,7 @@ int countDistinctPoints(const QVector<RS_Vector>& points) {
  * @param allowEntireCircleAsSegment
  * @return segment information
  */
-LC_Division::CircleSegmentData* LC_Division::findCircleSegmentBetweenIntersections(const RS_Circle* circle, const RS_Vector& snap,
+std::unique_ptr<LC_Division::CircleSegmentData> LC_Division::findCircleSegmentBetweenIntersections(const RS_Circle* circle, const RS_Vector& snap,
                                                                                    const bool allowEntireCircleAsSegment) {
     CircleSegmentData* result = nullptr;
     // detect all intersections
@@ -114,7 +114,7 @@ LC_Division::CircleSegmentData* LC_Division::findCircleSegmentBetweenIntersectio
         // determine selected segment edges
         result = findCircleSegmentEdges(circle, snap, allIntersections);
     }
-    return result;
+    return std::unique_ptr<CircleSegmentData>{result};
 }
 
 /**
@@ -124,7 +124,7 @@ LC_Division::CircleSegmentData* LC_Division::findCircleSegmentBetweenIntersectio
  * @param allowEntireLine
  * @return segment information
  */
-LC_Division::LineSegmentData* LC_Division::findLineSegmentBetweenIntersections(const RS_Line* line, const RS_Vector& snap,
+std::unique_ptr<LC_Division::LineSegmentData> LC_Division::findLineSegmentBetweenIntersections(const RS_Line* line, const RS_Vector& snap,
                                                                                const bool allowEntireLine) {
     LineSegmentData* result = nullptr;
     // find all intersection points for line
@@ -136,14 +136,13 @@ LC_Division::LineSegmentData* LC_Division::findLineSegmentBetweenIntersections(c
             result->segmentDisposition = SEGMENT_ENTIRE;
             result->snapSegmentStart = line->getStartpoint();
             result->snapSegmentEnd = line->getEndpoint();
-            result->snap = snap;
         }
     }
     else {
         // determine segments of line that was selected by the user
         result = findLineSegmentEdges(line, snap, allIntersections, allowEntireLine);
     }
-    return result;
+    return std::unique_ptr<LineSegmentData>{result};
 }
 
 /**
--- a/librecad/src/lib/modification/lc_division.cpp
+++ b/librecad/src/lib/modification/lc_division.cpp
@@ -57,7 +57,7 @@ LC_Division::ArcSegmentData* LC_Division::findArcSegmentBetweenIntersections(con
         if (allowEntireArcAsSegment) {
             // allowing deletion of complete arcs
             result = new ArcSegmentData();
-            result->segmentDisposition = SEGMENT_INSIDE;
+            result->segmentDisposition = SEGMENT_ENTIRE;
             result->snapSegmentStartAngle = arc->getAngle1();
             result->snapSegmentEndAngle = arc->getAngle2();
         }
@@ -84,6 +84,7 @@ LC_Division::CircleSegmentData* LC_Division::findCircleSegmentBetweenIntersectio
     if (allIntersections.empty()) {
         if (allowEntireCircleAsSegment) {
             result = new CircleSegmentData();
+            result->segmentDisposition = SEGMENT_ENTIRE;
             result->snapSegmentStartAngle = 0;
             result->snapSegmentEndAngle = M_PI * 2;
         }
@@ -111,7 +112,7 @@ LC_Division::LineSegmentData* LC_Division::findLineSegmentBetweenIntersections(c
         if (allowEntireLine) {
             // allow to delete entire line if SHIFT is pressed
             result = new LineSegmentData();
-            result->segmentDisposition = SEGMENT_INSIDE;
+            result->segmentDisposition = SEGMENT_ENTIRE;
             result->snapSegmentStart = line->getStartpoint();
             result->snapSegmentEnd = line->getEndpoint();
             result->snap = snap;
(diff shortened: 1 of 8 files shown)


librecad/src/lib/modification/lc_division.cpp
```diff
@@ -394,9 +395,11 @@ LC_Division::ArcSegmentData* LC_Division::findArcSegmentEdges(const RS_Arc* arc,
         // there are intersections only on edges. We may return the entire line as segment if SHIFT is pressed and the user would like to delete the entire entity
         if (allowEntireArcAsSegment) {
             result = new ArcSegmentData();
-            result->segmentDisposition = SEGMENT_INSIDE;
-            result->snapSegmentStartAngle = arcStartAngle;
-            result->snapSegmentEndAngle = arcEndAngle;
+            result->segmentDisposition = SEGMENT_ENTIRE;
+            // report the arc's own angles, as the no-intersections branch above does:
+            // arcStartAngle/arcEndAngle are swapped for a reversed arc
+            result->snapSegmentStartAngle = arc->getAngle1();
```
sand1024: @dxli 
It looks like AI junk code - why not reuse arcStartAngle and arcEndAngle variables as they are defined above?

sand1024:
@dxli 

Well, potentially that fix will work - however, as for me, this fix is a brilliant and excellent example of neuro slope code. 

AI had generated description that looks quite realistic as well as lots of the code which could even work... Yet that code is is redundant and dumb enough.  

The fix is actually much simpler - it's necessary just skip attempts to create zero length lines and arcs - and that's guard code is useful anyways. 

So to fix this issue, changes are:

1) just 3 added lines the code in `LC_ActionModifyBreakDivide::createLineEntity()` 

```
   ...
   }
   else{
   ----
      if (LC_LineMath::isNotMeaningfulDistance(start, end)) {
        return;
      }
   ----   
```
2) adding corresponding check in LC_ActionModifyBreakDivide::createArcEntity()
```
   ...
   }
   else{
   ----
    double angleDif = RS_Math::getAngleDifference(arcData.angle1, arcData.angle2);
        if (!LC_LineMath::isMeaningfulAngle(angleDif)) {
            return;
        }
   ----   
```

3) changed condition in LC_ActionModifyBreakDivide::doCheckMayTrigger()

` if (m_triggerData->entitiesToCreate.isEmpty() && !m_alternativeActionMode){`

That's all. No need in additional statuses, checks and so on. 

Please don't rely on AI, so far in most cases it's produces quite a stupid code, and is usable rather to absolutely straightforward dumb code or for quick prototyping.

PR #2732, opened 2026-08-04
[PROPOSAL] Add the ability for plugins to implement custom dockwidgets, toolbars & menus

qwilvove:
Hello,

Plugins with custom dockwidgets, toolbars & menus could be very useful to add more specialised features to LibreCAD.

The idea is to add lists in ```PluginCapabilities``` to hold widgets and data about them (just the name of the widget for now), using new classes created on the same model as ```PluginMenuLocation```. Then, during the application init process, load plugins after building the interface, and update it to add widgets defined in plugins (maybe adding application widgets and plugins widgets at the same time is the way to go).
Since the actions are triggered from the plugin (and not only from the application window using execComm), the plugin needs to hold the document interface (during its initialisation), and update it when it is used.

The implementation is very experimental at the moment. Your feedback and advices will be much appreciated.
I'm not sure about calling ```update()``` at the begining of functions in ```doc_plugin_interface.cpp```. Maybe I've missed important points regarding the functions to call to create valid dockwidgets, toolbars & menus.

Further work on the plugin interface might be needed to improve plugin use for LibreCAD. For example, adding more functions to document interface might be interesting (e.g., several functions to draw the same entity type : circle 3 points, circle 2 points, circle centre and point...), as well as a properties page to enable or disable a widget. I wanted to update sample plugin to provide custom widget examples, but that would have made the interface too cluttered.

I hope that this proposal could be interesting for the project.

Best regards,

qwilvove


sand1024:
@qwilvove 
I've checked PR (very briefly so far, sorry) - yes, it seems it could be good direction in plugins support. The only thing is that I'm not sure that visibility of toolbars and widgets is persistent (i.e. if the user disabled one in current editing system, it should stay disabled after restart too - as it is supported for ordinary toolbars). But probably I've missed that. 

And regardless of your PR - a small general comment.

The major issue with plugins still stays - I suppose the overall plugin API provided by Doc_plugin_interface.h is, well, quite limited and outdated. 

In general, the entire plugin API requires revising - and, most probably, this is quite close to potential support of scripting and both plugins and scripting needs some base simplified facade API that will cover the core code.

qwilvove:
@sand1024
Thank you for your review. You were right, widgets and toolbars visibility was not persistent. I've edited the PR to init plugins before widgets, toolbars and menus. From now on, during the application initialisation process, plugins are loaded first, followed by the construction of widgets, toolbars and menus (including widgets provided by the plugins, as the list of loaded plugins is known). The process is simpler and perhaps cleaner, as it is no longer necessary to build the interface twice.

sand1024:
not complete implementation. May be should be integrated later.

PR #1427, opened 2021-11-23
DimAngular horizontal text gap, w.r.t. issue in pull #1409

melwyncarlo:
* DimAngular horizontal text gap, w.r.t. issue in pull #1409

<hr>

Screenshot from 2021-11-23 16-19-38

PR #2755, opened 2026-08-22
Honor horizontal text alignment in angular dimensions (#1396)

dxli:
Angular dimensions ignore the drawing's text alignment setting. Fixes the second bug reported in #1396 / #1409; supersedes #1427, which fixed this against the 2021 code and no longer applies.

## The bug

`RS_DimAngular::doUpdateDim()` never consulted `getInsideHorizontalText()`. It unconditionally ran the quadrant-rotation logic and passed the resulting angle to `createDimText()`, so the label was always aligned with the arc no matter what `$DIMTIH` said.

Angular was the only dimension type left out — `RS_Dimension::createDimensionLine()` dispatches on the same setting for linear dimensions, `LC_DimArc` honors it for arc-length dimensions, and `$DIMTIH` is read from DXF into the style (`rs_filterdxfrw.cpp`, group code 73). Note the policy defaults to `DRAW_HORIZONTALLY`, so this affects ordinary drawings, not just ones that set the variable explicitly.

## The change

Two things, both scoped to the horizontal case:

- **Leave the label unrotated.** The rotation block is now gated on `!horizontalText`; the text keeps `textAngle = 0` and stays at the arc's midpoint instead of being pushed clear of it.
- **Break the arc around the text**, so it isn't drawn straight through the label — the behaviour linear dimensions already get by splitting their dimension line, and the thing @jbergengh asked for in [#1409](https://github.com/LibreCAD/LibreCAD/pull/1409#issuecomment-975885935) ("The dimension lines should have a gap around the numerical value instead of overwriting the text").

The break follows `LC_DimArc`'s approach: the existing arc is trimmed back until its endpoint reaches the text box, and a second arc runs from the text to the far end. One difference — because the text sits on the middle of the arc, half the sweep is a hard upper bound on both walks, so the loops are bounded by a step count instead of `LC_DimArc`'s open-ended `RS_MAXDOUBLE`/`RS_MINDOUBLE` guards.

Aligned text is untouched: same rotation, same offset, same single unbroken arc as before.

61 insertions across the two files, no changes to shared code. Written against the current `addDimArc()` / `createDimText()` helpers rather than raw `setPen`/`setLayer`/`addEntity` calls, which is why #1427 could not simply be rebased — that file has since been refactored so those helpers resolve the pen internally.

## Verification

A throwaway Catch2 test (not included in the commit) built the same dimension twice, differing only in the orientation policy, and inspected the generated child entities:

```
aligned     arcs=1  textAngle=5.30144
horizontal  arcs=2  textAngle=0
```

So aligned text stays rotated over a single arc, horizontal text is unrotated with the arc split in two — exactly the intended difference, and no change to the aligned path.

Full existing suite: **13309 assertions in 773 test cases, all passing**, plus a clean release build.

### DIFF
--- a/librecad/src/lib/engine/document/entities/lc_dimarc.cpp
+++ b/librecad/src/lib/engine/document/entities/lc_dimarc.cpp
@@ -200,12 +200,10 @@ void LC_DimArc::doUpdateDim() {
 
         constexpr double deg360{M_PI * 2.0};
 
-        constexpr double degTolerance{1.0E-3};
-
         /* With regards to Quadrants #1 and #2 */
-        if (((textAngle_preliminary >= -degTolerance) && (textAngle_preliminary <= (M_PI + degTolerance)))
-            || ((textAngle_preliminary <= -(M_PI - degTolerance)) && (textAngle_preliminary >= -(deg360 +
-                degTolerance)))) {
+        if (((textAngle_preliminary >= -g_dimTextQuadrantTolerance) && (textAngle_preliminary <= (M_PI + g_dimTextQuadrantTolerance)))
+            || ((textAngle_preliminary <= -(M_PI - g_dimTextQuadrantTolerance)) && (textAngle_preliminary >= -(deg360 +
+                g_dimTextQuadrantTolerance)))) {
             textPosOffset.setPolar(getDimensionLineGap() * getGeneralScale(), textAngle_preliminary);
             textAngle = textAngle_preliminary + M_PI + M_PI_2;
         }
@@ -293,18 +291,16 @@ void LC_DimArc::doUpdateDim() {
 
     //TODO: the current algorithm to find dimArc1/dimArc2 angles could be
     // costly.
-    constexpr double deltaOffset{1.0E-2};
-
     while (!textRectRotated.inArea(dimArc1->getEndpoint())
         && (dimArc1->getAngle2() < RS_MAXDOUBLE)
         && (dimArc1->getAngle2() > RS_MINDOUBLE)) {
-        dimArc1->setAngle2(dimArc1->getAngle2() + deltaOffset);
+        dimArc1->setAngle2(dimArc1->getAngle2() + g_dimArcTextStep);
     }
 
     while (!textRectRotated.inArea(dimArc2->getStartpoint())
         && (dimArc2->getAngle1() < RS_MAXDOUBLE)
         && (dimArc2->getAngle1() > RS_MINDOUBLE)) {
-        dimArc2->setAngle1(dimArc2->getAngle1() - deltaOffset);
+        dimArc2->setAngle1(dimArc2->getAngle1() - g_dimArcTextStep);
     }
 
     dimArc1->setPen(pen);
--- a/librecad/src/lib/engine/document/entities/rs_dimangular.cpp
+++ b/librecad/src/lib/engine/document/entities/rs_dimangular.cpp
@@ -30,6 +30,7 @@
 #include<iostream>
 
 #include "lc_linemath.h"
+#include "lc_rect.h"
 #include "rs_arc.h"
 #include "rs_constructionline.h"
 #include "rs_debug.h"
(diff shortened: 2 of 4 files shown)


dxli:
Reviewed my own patch against correctness / comment concision / minimum change, and the correctness pass found **a real bug in the first version, which is now fixed** (force-pushed).

## The bug: the two arc pieces overlapped across the text, leaving no gap

The first version trimmed the existing full arc backwards from its far end, stopping once the endpoint **entered** the text box. That stops at the box's *far* edge, so that piece already spans the whole text region — and the second piece did the same from the other side. The two met inside the box: no gap at all, which is the very thing the change exists to produce.

It was invisible to my original test because that test asserted only "2 arcs, text angle 0". I rebuilt the probe to compare the horizontal case against the **aligned case for identical geometry** (the aligned path yields one unbroken arc, so it is the true baseline) and measured the gap directly. Every case came back with a *negative* gap — the pieces summing to more than the whole arc:

```
normal 55deg      trueSweep=0.9599  total=0.9799   GAP=-0.0199
spans angle 0     trueSweep=0.6981  total=0.7763   GAP=-0.0781
tiny arc small r  trueSweep=0.0873  total=0.1745   GAP=-0.0873   (arc drawn twice)
```

The fix is to grow each piece from its own end towards the text, as `LC_DimArc` does, instead of trimming inwards from the far end. After it:

```
normal 55deg      trueSweep=0.9599  total=0.94     GAP=+0.0199
long label        trueSweep=0.5236  total=0.48     GAP=+0.0436
spans angle 0     trueSweep=0.6981  total=0.62     GAP=+0.0781
near-full 350deg  trueSweep=6.1086  total=6.04     GAP=+0.0687
tiny 3deg arc     trueSweep=0.0524  total=0.04     GAP=+0.0124
tiny arc small r  trueSweep=0.0873  total=0.0873   GAP=0  (1 arc, unbroken)
```

## A second hazard the fix exposed

Growing from each end means starting from a piece whose start and end angles are equal — and `RS_Arc` treats that as a **full circle**, not as zero length (`getAngleLength()` maps a zero sweep to 2π). So when the text is wider than the whole arc, both collapsed pieces rendered as full circles: `total=12.57` (4π) in the probe.

Guarded by leaving the arc unbroken when the text box already covers either end, which is also the sensible visual outcome — there is nothing to keep on either side of the label. That is the `tiny arc small r` row above: one arc, no duplicate, no circle.

Worth noting for a separate look: `LC_DimArc` builds its pieces degenerate from the start (`RS_ArcData(centre, radius, startAngle, startAngle, false)`) with no such guard, so the same full-circle hazard appears to be latent there. Not touched here.

## Other checks

- **Angle wrapping.** `setAngle1/2` normalise through `RS_Math::correctAngle`, so I checked an arc spanning angle 0 explicitly (`spans angle 0` above) — correct gap, no blow-up.
- **Stale borders.** The patch mutates an arc that is already in the container; `RS_Dimension::update()` calls `calculateBorders()` after `doUpdateDim()`, so the container bounds are refreshed.
- **Loop bounds.** Both walks are bounded by half the sweep, so neither piece can cross the middle or wrap past the other end.
- **Aligned path untouched**: still one arc, still rotated (5.30 rad in the probe).
- Full suite: **13309 assertions in 773 test cases, passing.**

## Comments and size

Trimmed the doc block to `@brief` + `@param` list, matching the neighbouring `arrow()` and `extensionLine()` helpers, and dropped the prose paragraph that restated the code. The two remaining inline comments each explain a non-obvious *why* (the full-circle degeneracy, and why half the sweep bounds the walk).

On structure: I considered making this a file-local free function to avoid touching the header at all (-5 lines), but every other helper in this file is a private member with a doxygen block, so the member fits the file's idiom; matching convention seemed worth more than five lines. The change is +68/-13 in one file plus the declaration, with no shared code touched.

dxli:
Went through the numeric literals as asked. The PR touches four: `0.001`, `1.0E-2`, and `2.0`/`3.0` (the latter two are plain arithmetic — `M_PI_2 * 3.0`, halving a sweep — not tuning values, so I left them).

## 1. Validating `0.001` — the clause it guards is unreachable

```cpp
if (angle1 > M_PI_2 * 3.0 + 0.001 || angle1 < M_PI_2 + 0.001)
```

`angle1` is `textPos.angleTo(dimCenter) - M_PI_2`. `angleTo()` goes through `RS_Vector::angle()` → `RS_Math::correctAngle()`, which returns **[0, 2π)**, so `angle1` is confined to **[−π/2, 3π/2)**. The first test asks whether it exceeds `3π/2 + 0.001`, which is outside that range by construction.

Sampling the real `correctAngle` over several full turns:

```
angle1 range: [-1.570796327, 4.712388980]
3*pi/2 + 0.001 = 4.713388980
times first clause was TRUE: 0
```

So the condition is effectively just `angle1 < M_PI_2 + 0.001`, and the paired `// quadrant 1 & 4` / `// quadrant 2 & 3` comments describe a split that never happens on the first half. **I have not changed this** — it is pre-existing logic on the *aligned* path, which this PR deliberately leaves alone, and removing it would mean touching text placement for every existing angular dimension. Worth its own change if you want it; happy to do that separately.

As for the value itself: it is not a numerical tolerance. `RS_TOLERANCE_ANGLE` is `1.0e-8`, five orders of magnitude smaller. `0.001` rad (≈0.06°) is a deliberate widening so text landing exactly on a quadrant boundary picks a stable side. The name and comment now say that.

## 2. Both constants were already duplicated, so they now live together

Neither is new to this PR — both already existed in two places:

| value | `RS_DimAngular` | `LC_DimArc` |
| --- | --- | --- |
| `1.0E-2` | arc walking step | `deltaOffset` (line 296) |
| `1.0E-3` | `0.001` quadrant widening | `degTolerance` (line 203) |

Same values, same purposes, different spellings, free to drift. Both are now named constants next to the shared `RS_Dimension` base, and **both classes use them**:

```cpp
constexpr double g_dimArcTextStep = 1.0E-2;
constexpr double g_dimTextQuadrantTolerance = 1.0E-3;
```

`LC_DimArc`'s local `deltaOffset` and `degTolerance` are gone, so the two implementations can no longer disagree about the step size or the quadrant boundary.

## 3. One more thing this turned up

`rs_dimension.h` already forward-declares `RS_Arc` and `RS_ArcData`, so the two forward declarations I had added to `rs_dimangular.h` were redundant. Removed — the header change is down to the one method declaration.

## Verification

Full suite after the refactor: **13309 assertions in 773 test cases, passing** — identical to before, and that suite exercises `LC_DimArc`, so swapping its two constants for the shared ones is behaviour-neutral. Clean build.

Now +83/−22 across four files.

PR #2759, opened 2026-08-23
Add missing Turkish glyphs to the default LFF font

### LINKED ISSUE #2758: [BUG] Turkish characters (ğ, ş, ı, vb.) display as diamond symbols in Text and MText
### Description
Some Turkish specific characters (such as `ğ, Ğ, ş, Ş, ı, İ`) are not rendering properly when using the `Text` or `MText` tools. Instead of the correct character, a vertical diamond symbol (replacement character ``) is displayed.

### Steps to Reproduce
1. Open LibreCAD.
2. Select the **Text** or **MText** tool.
3. Type a word containing Turkish characters (e.g., `ÜĞİŞÇÖ, üğışçö, ılık, kaşık, ağaç`).
4. Place the text on the drawing area.
5. Notice that Turkish-specific characters are replaced with diamond symbols.

### Expected Behavior
All Turkish characters should render correctly using standard or selected compliant fonts.

### Actual Behavior
Specific characters fail to display correctly in the viewport. Instead of the Turkish characters, a vertical diamond symbol (question mark inside a diamond/replacement character ``) is displayed.

### System Information
- **LibreCAD**
Version: v2.2.1.5-32-g8b48125cc
Compiler: Microsoft Visual C++
Build date: August 16, 2026
Qt Version: 5.15.2
Boost Version: 1.87.0
- **OS:** Windows 11 TR

Image

dxli:
Fixes #2758.

## Root cause

Turkish-specific characters (`ğ`, `Ğ`, `ş`, `Ş`, `ı`, `İ`) have no glyph entries in `standard.lff`, the default vector font used by the Text and MText tools:

```
$ grep '^\[011E\]\|^\[011F\]\|^\[0130\]\|^\[0131\]\|^\[015E\]\|^\[015F\]' librecad/support/fonts/standard.lff
(nothing)
```

`RS_Text::update()` and `RS_MText::update()` both react to a missing glyph the same way - they call `font->findLetter()`, and on `nullptr` substitute `QChar(0xfffd)`:

```cpp
if (font->findLetter(letterText) == nullptr) {
    ...
    letterText = QChar(0xfffd);
}
```

and `standard.lff` *does* define `0xfffd` (line 853, the replacement-character glyph) - so the visible symptom is exactly what's reported: every Turkish-specific character renders as the diamond placeholder instead of itself. This is data-completeness, not a rendering-pipeline bug; the missing-glyph fallback is working as designed, it's just falling back for characters that shouldn't need to.

`standard` is confirmed as the actual default: `RS_MText` requests it explicitly (`rs_mtext.cpp:650`), and `RS_FontList::requestFont()` falls back to it whenever a named font can't be found at all (`rs_fontlist.cpp:125`).

## The fix

Six new entries in `librecad/support/fonts/standard.lff`, inserted in the file's existing (roughly hex-ascending) order:

```
[011E] Ğ
C0047
1.4,11.4;2.5,10.85,A-0.32;3.6,11.4,A0.32

[011F] ğ
C0067
0.9,8.4;2.0,7.85,A-0.32;3.1,8.4,A0.32

[0130] İ
C0049
0,11.5;0,12

[0131] ı
0,0;0,6

[015E] Ş
C0053
3.0,-0.1;2.6,-1.6,A-0.4;1.7,-1.9,A-0.4

[015F] ş
C0073
2.3,-0.1;1.9,-1.6,A-0.4;1.0,-1.9,A-0.4
```

Each uses the file's own composite-reference syntax (`C<hex>` includes the base letter's strokes, matching how e.g. `ç` = `C0063` + a cedilla stroke elsewhere in this same font) rather than duplicating the base letter's geometry:

- **ı / İ** are the trivial cases: `ı` is `i`'s stem without its dot; `İ` is `I`'s stem plus a dot drawn in the same style as `i`'s own dot.
- **ğ / Ğ** add a breve (small open arc) above `g` / `G`.
- **ş / Ş** add a cedilla (hook) below `s` / `S`.

Positioning followed the font's *own* established conventions rather than my own eyeballing, once I found them: this font already has `Ś`/`ś` (acute), `Š` (caron), and `Ą`/`ą`/`Ę`/`ę` (ogonek) accented letters. The existing acute/caron marks sit at a fixed 2.5-unit gap above the base letter's own top, for both x-height and cap-height letters alike - so the new breve marks use that exact same gap, landing on the same visual line as the existing `Š`/`Ś` accents rather than an independently-chosen height. The existing ogonek marks are single below-baseline hooks attached at the letter's own right edge; the new cedillas follow that same placement logic, just with a shape (multi-arc, curling left) that matches a cedilla rather than an ogonek, since Turkish orthography uses the former.

Every glyph was rendered and visually checked - full-size and at typical small-annotation scale - against the actual bulge-arc math the LFF parser (`RS_Font::generateLffFont`) uses, before being written into the font file.

## Test

`rs_font_tests.cpp` (new file, new `fonts/tests/` directory alongside the existing per-module pattern in `entities/tests/`) loads the real, installed `standard.lff` directly - no mocking - and asserts all six codepoints resolve to a non-empty glyph via `RS_Font::findLetter()`. A companion test asserts a genuinely-undefined codepoint (Private Use Area `U+E000`) still correctly returns `nullptr`, so the fix can't be "verified" by a test that would pass no matter what.

Verified the test actually catches the bug: reverted `standard.lff` to the pre-fix (master) content and reran -

```
REQUIRE( letter != nullptr )
with expansion:
  nullptr != nullptr
with message:
  G with breve (Ğ)
```

- restored the fix, green again.

## Verification

Built locally on both build systems, from a clean configure at `175e6baad`:

- **CMake** - 0 errors; full app + `librecad_tests` build (1275 targets).
- **qmake** - 0 errors; full app build.

Tests: `[font]` 15 assertions / 2 cases (new), `[i18n]` 99/27, `[entity-encode]` 922/55, `[dwg-write]` 2348/119, `[text]` 27/11, `[mtext]` 88/12 - all passing, no regressions. The full untagged suite hits a pre-existing, unrelated hang in this environment (a test transitively scans `$HOME/library`, which macOS resolves case-insensitively to `~/Library`); targeted tags stood in for it, as they have for every other PR in this line of work.

## Scope

Only `standard.lff` is touched. Two things worth flagging that are explicitly *not* fixed here:

- **Font-selection fallback is whole-font, not per-character.** If a document uses a different named font (`isocp`, `romans`, ...) that's *also* missing these glyphs, `RS_FontList::requestFont()` only substitutes `standard` when the named font can't be found at all - not per missing character within an otherwise-loaded font. So a user on a non-default text style could still see the diamond. Extending coverage to every bundled stroke font is a much larger undertaking than this report calls for.
- **`standard.lff`'s Latin coverage has other gaps.** Even common accented Latin letters like `Ä`, `Ö`, `Ü`, `É`, `È` have no entries either. Out of scope for this report, which is specifically about the six Turkish characters; noting it in case it's worth its own follow-up.


dxli:
Combined into #2773 along with #2769, #2770, #2771 and #2772 - same commit, unchanged.

PR #2769, opened 2026-08-23
Add the accented Latin and Cyrillic letters the translations need

dxli:
`unicode.lff` offers the widest script coverage LibreCAD ships, but it was missing **every** accented Latin letter used by the Turkish, Romanian, Latvian, Hungarian, Czech, Slovak, Polish and Vietnamese translations, plus the Macedonian/Serbian Cyrillic letters.

A missing glyph is silent. `RS_Text::update()` and `RS_MText::update()` substitute `QChar(0xfffd)` when `findLetter()` returns null, and `unicode.lff` defines `0xfffd` — so the text simply renders as replacement diamonds, with no warning.

## Finding the gap

I extracted every character appearing inside a `<translation>` body across all 66 shipped `.ts` files and diffed against the font's 2,693 codepoints. **69 codepoints** used by real translations had no glyph:

| | count | locales |
|---|---:|---|
| accented Latin | 36 | `tr ro lv hu cs sk sl pl et` |
| Vietnamese | 17 | `vi` |
| Macedonian / Serbian Cyrillic | 16 | `mk sr sr_ba sr_me sr_rs` |

*(Methodology note: a naive character diff also flags Tamil/Georgian/Khmer inside the Spanish files. That is not a font requirement — LibreCAD has character-picker UI strings like `[10A0-10FF] Georgian`, and separately several `es_*` files are contaminated with other languages' translations: `es_gt` carries 384 Tamil characters, `es_uy` 1,532 Cyrillic, `es_pe` 1,291 Arabic. Worth its own issue; not addressed here.)*

## Built from the font's own material

Every glyph is composed from what `unicode.lff` already contains — no external font is involved.

The font already carries a complete diacritic vocabulary on its existing accented letters. The marks here are lifted from those verbatim and re-centred on each base letter's measured width:

| mark | taken from | shape |
|---|---|---|
| breve | `Ă` U+0102 | `1.75,11.75;4.25,11.75,A1` |
| cedilla | `Ç` U+00C7 | `2,0;1.5,-2` |
| macron | `Ā` U+0100 | `.5,12;5.5,12` |
| dot above | `Ż` U+017B | `2.5,11.5;2.5,12` |
| acute / grave / circumflex / caron / tilde | `Á À Â Č Ã` | as-is |

Each new letter is a `C<hex>` composite over a base glyph the font already has — exactly how it already builds `Ś`, `Š` and `Ą`. So the new letters match the existing ones by construction rather than by eye.

The Cyrillic follows the same principle: `Ѕ ѕ Ј ј` reuse the Latin letterforms they share; `Ќ ѓ ќ` are base+acute; `Љ Њ љ њ` reuse the soft-sign bowl as a ligature on `Л`/`Н`. Only `Џ џ Ћ ћ ђ` needed genuinely new strokes.

## Why not convert a Unicode TTF

Two independent blockers, both verified rather than assumed:

**Licence.** SIL OFL 1.1 §5 requires the Font Software *"in part or in whole … be distributed entirely under this license, and must not be distributed under any other license"*, and its definition of a Modified Version explicitly covers *"changing formats or … porting the Font Software to a new environment"* — i.e. TTF→LFF. `unicode.lff` declares GPL v2 or later, so OFL-derived glyphs cannot be merged into it. That rules out Noto and essentially every wide-coverage Unicode font. (`amiri-regular.lff` works today precisely because it is a *separate* OFL-declared file, not a merge.)

**Style.** `unicode.lff` is a single-stroke centreline font — 2.0 vertices per stroke across all 2,693 glyphs. `ttf2lff` traces outlines and has no skeletonisation mode, so a converted `O` is two concentric ellipses at ~100 vertices per stroke. Its tuning flags cut file size ~3.8× but leave vertex count essentially unchanged (100.2 → 95.4); the saving is coordinate precision, not geometry.

So for the scripts that need genuinely new letterforms, hand-drawing in the font's own idiom is the only route that satisfies both constraints.

## Test

`unicode_font_tests.cpp` asserts against the **real installed font** that each of the 69 resolves to a non-empty glyph and stays inside a sane drawing envelope. The envelope check is what catches a bad `C<hex>` reference or a mark placed at the wrong height, rather than only catching outright absence.

It earned its keep during development — it caught `Љ`/`Њ` running to x=10.0 (an `A1` semicircle on a 5-unit chord always bulges 2.5, making the ligature twice the width of its base); the bowls now use `A.8` and end at 8.0.

Verified the test fails on injected defects — a corrupted `C<hex>` reference and a deleted glyph both produce `nullptr != nullptr` naming the exact codepoint — and passes once restored.

## Verification

Built locally on both build systems from a clean configure at `175e6baad`:

- **CMake** — 0 errors.
- **qmake** — 0 errors.

Tests: `[unicode]` 512 assertions / 3 cases (new), `[i18n]` 569/27, `[text]` 27/11, `[mtext]` 88/12 — all passing. Glyph count 2,693 → 2,762; file grows ~9 KB.

## Not covered

Seven scripts still have **zero** coverage anywhere in the repo — Thai, Devanagari, Gurmukhi, Lao, Tamil, Georgian (and Arabic, which exists only in the separate `amiri-regular.lff`), together ~383 glyphs for `th hi pa lo ta ka ar`. No GPL-compatible single-stroke font exists for any of them; all 16 Cyrillic letters added here, for instance, exist elsewhere in the repo *only* in the OFL/unknown-licensed `opengost` fonts. Those scripts would need either separate OFL-declared `.lff` files (legal, but outline-style like Amiri) or original stroke design.


dxli:
Combined into #2773 along with #2770, #2771 and #2772 - same commit, unchanged.

PR #2770, opened 2026-08-23
Add the modern Georgian (Mkhedruli) alphabet to unicode.lff (stacked on #2769)

dxli:
**Stacked on #2769.** This branch is built on top of `feat/unicode-lff-latin-cyrillic`, so the diff below currently includes #2769's own commit as well as this one - GitHub will narrow it down to just this PR's commit automatically once #2769 merges into `master`. It extends the same coverage-gap analysis and the same test file (`unicode_font_tests.cpp`) that PR introduces, so please land #2769 first; this rebases cleanly on top with no changes needed.

## Why Georgian, specifically

Of the seven scripts `unicode.lff` had zero coverage for, Georgian is the one where "just add the glyphs" actually works. From the research on the other six:

- No shaping engine exists anywhere in LibreCAD's text pipeline (confirmed by grep and by an empirical probe against the already-shipped `amiri-regular.lff`, which carries all 749 Arabic presentation-form glyphs and never has them selected for plain typed text).
- Devanagari/Gurmukhi/Tamil need conjunct formation and vowel-sign reordering; Arabic needs positional joining; Thai/Lao need zero-advance mark stacking. None of that exists in `RS_Text`/`RS_MText`, which place one glyph per codepoint at the next horizontal advance, unconditionally.
- Georgian (Mkhedruli) is a simple 33-letter unicameral alphabet — no joining, no conjuncts, no reordering, no case. It just needs the 33 glyphs to exist.

And unlike the other six, it needs no external font at all: the standard Mkhedruli letterforms are public domain, so this is original single-stroke authoring, not a derivative of anything under license.

## Methodology — this needed real verification, not eyeballing

Hand-drawing 33 unfamiliar letterforms for a script I don't read is a real correctness risk, so I built an actual verification pipeline rather than trusting my first draft:

1. **Reference, not source.** Downloaded Noto Sans Georgian (OFL, irrelevant to licensing here since none of its data is used) purely to render a labelled sheet of all 33 letters to look at — the same way a type designer studies an existing alphabet before drawing their own strokes. No outline coordinates from it appear anywhere in this file.
2. **Arc math verified against the real engine, not assumed.** An early version of my preview renderer had the bulge convention backwards. I caught this by probing `RS_Font::findLetter()` + `RS_Block::calculateBorders()` directly (`RS_Polyline::addVertex`'s actual `setNextBulge`-before-`addVertex` semantics) against known-good existing glyphs (`O`, `S`, `C`, `A`, `D`) until the renderer's output matched LibreCAD's own geometry exactly, then used that corrected renderer for every letter that follows.
3. **Every glyph rendered side by side with its Noto reference.** Checked for topology — right number of loops, correct ascender/descender pattern, strokes actually connecting — not pixel fidelity, which a 2-vertex-per-stroke single-line font was never attempting.

## What that process caught

- `დ`/`თ` initially rendered as visually disconnected strokes where the reference shows one continuous shape — fixed by redrawing as connected paths.
- **Three closed loops — `ტ`, `წ`, `ჭ` — self-intersected into a bowtie.** Their two semicircular end-caps were built starting from the top chord instead of the bottom one; DXF bulge direction depends on the direction of travel along each chord, and getting that ordering backwards on a closed loop makes the two caps cross instead of closing. Found by auditing *every* closed-loop glyph for that same ordering, not by trusting the ones that happened to look right in a first pass — `წ` in particular looked plausible at a glance and was still wrong.

## A real gap in the test, disclosed rather than glossed over

`unicode_font_tests.cpp`'s envelope check (bounding-box sanity: is the glyph non-empty, does it stay in a sane drawing region) **does not catch a self-intersecting loop**. I confirmed this directly: reintroduced the `ტ` bowtie bug, reran `[unicode]`, and it passed clean — the bug doesn't touch the bounding box, only the interior geometry. Only the visual side-by-side comparison caught it, three separate times.

I'm recording this rather than letting the green test suite imply more confidence than it earns. The test still does real work — it caught wrong glyphs during development the same way it caught the earlier bad `C<hex>` reference and deleted-glyph cases in #2769 — but for closed-loop topology specifically, it is not a substitute for the visual check.

## Test

Extends the existing `unicode_font_tests.cpp` with a third codepoint range (`kGeorgian`, all 33) through the same `checkRange()` used for Latin and Cyrillic.

## Verification

Built locally on both build systems from a clean configure. `[unicode]` 744 assertions / 4 cases (was 512/3 before this PR), `[i18n]` 801/28 — all passing. Glyph count 2,762 → 2,795.

## Honest bottom line

This is best-effort single-stroke Mkhedruli — structurally verified against the real rendering engine and cross-checked against a real reference for topology, but I am not a Georgian reader and this has not had a native-speaker review. I'd treat that as a real prerequisite for merge, not a formality. Happy to hold this open until someone who reads Georgian can confirm the letterforms, or to have it merged provisionally with that follow-up tracked.


dxli:
Combined into #2773 along with #2769, #2771 and #2772 - same commit, unchanged.

PR #2771, opened 2026-08-23
Add Thai consonants, vowels and combining marks to unicode.lff (stacked on #2770)

dxli:
Stacked on #2770 (Georgian), which is stacked on #2769 (Latin/Cyrillic). Until #2770 merges, this diff includes its commits too; it will narrow to just the Thai commit once #2770 lands.

## What

`unicode.lff` had zero Thai coverage - every Thai codepoint fell back to the replacement-character diamond. This adds:

- All 44 modern consonants, U+0E01-0E2E, skipping the two obsolete letters (KHO KHUAT, KHO KHON) that are already excluded from the standard modern alphabet.
- The 11 Lo/Lm vowels and marks that place correctly with LibreCAD's plain sequential glyph advance: sara a/aa/am, the four leading vowels (e/ae/o/ai×2), lakkhangyao, paiyannoi, maiyamok.
- The 16 Mn combining marks: the five tone marks, sara i/ii/ue/uee/u/uu, and phinthu.

71 glyphs total, `unicode.lff` glyph count 2795 → 2866.

## How the geometry was derived

`unicode.lff` is a single-stroke (centerline) font; Thai needs curves and loops that are hard to hand-draft reliably (I tried by hand for the first few consonants and kept getting the topology wrong at a glance). Instead I built a small offline pipeline:

1. Rasterize each glyph from NotoSansThai-Regular.ttf (SIL OFL) with FreeType.
2. Compute its medial-axis skeleton (scikit-image `skeletonize`).
3. Walk the skeleton as a graph, prune skeletonization artifacts (short spurs, thickness-induced duplicate junctions), and decompose it into stroke paths - including a case I initially got wrong, a pure closed ring with no junction (the circle in ำ/SARA AM), which needs walking all the way around back to the start rather than the usual junction-to-junction walk.
4. Simplify each path (Ramer-Douglas-Peucker) and fit a circular arc per segment (least-squares circle fit) to get an LFF bulge, falling back to a straight segment when the fit is close to flat.

Only the resulting stroke geometry is used - no Noto outline data, hinting, or code ships in the file - so this doesn't carry OFL obligations into `unicode.lff` (GPLv2-or-later); it's the same relationship the existing hand-drawn glyphs in this file have to any reference font used while drawing them.

Every glyph was checked by rendering it back and comparing it against the reference glyph, and the three consonant clusters that are genuinely easy to misread as each other were checked pairwise against their siblings, not just against the reference in isolation:

- ผ/ฝ/พ/ฟ (shared "W" base shape, differ by a small top-left loop and/or a top-right ascender spike)
- ฎ/ฏ (identical upper half; ฏ has an extra zigzag in its below-baseline foot)
- ถ/ภ - these turned out to share nearly the *same* outline in Noto itself (see below)

### ถ/ภ needed a deliberate exaggeration, not a faithful trace

thor thung (ถ) and phor samphao (ภ) looked almost identical in the traced output, and a side-by-side crop of the source font confirmed why: in NotoSansThai-Regular they really are nearly the same glyph, differing only in a small, subtle difference in which way the bottom-left foot curls (left-flick vs. tucked-right hook). That's legible in a filled outline font because of the fill weight, but it's exactly the kind of detail a thin single-stroke glyph loses. I deliberately exaggerated that real difference so the two stay unambiguous as LFF glyphs rather than shipping a faithful trace of two letters that are too close to read apart at this weight. Flagging this explicitly since it's a judgment call, not a mechanical trace.

## Known limitation: combining-mark positioning

LibreCAD's text engine (`RS_Text`/`RS_MText`) has no shaping and no zero-advance combining-mark placement - confirmed by grep (no harfbuzz/shaping code anywhere in the entity code) and by how `amiri-regular.lff`'s Arabic presentation forms are already unreachable from ordinary typed text. Every glyph, including combining marks, is placed at the current pen position and the pen always advances by the glyph's own width.

The 16 Mn marks here are drawn at their real vertical offset (above/below the letter body, matching where they sit in the reference font) so they're ready for a correct zero-advance fix later, but until that engine work happens they will render *after* the base letter's advance width instead of stacked on top of it - visibly wrong, not silently wrong. Plain vowel/consonant text (the large majority of real Thai text) is unaffected by this; only tone-marked or vowel-pointed text hits it.

## Test coverage, and what it does not catch

`unicode_font_tests.cpp` gained a `kThai` array and a test case that resolves every codepoint through the real, installed `standard.lff`/`unicode.lff` loader and checks it's non-empty and inside a plausible bounding-box envelope (Thai needed a wider envelope than the existing Latin/Cyrillic/Georgian check - some consonants carry a second stroke past x=9, the marks sit left of their own recentered origin, and three below-baseline marks have nothing above y=1). `[unicode]` is 1242 assertions across 5 cases; the full suite is 14620 assertions / 783 cases, green.

As with the Georgian PR, I want to be explicit about what this test does *not* catch: it's an envelope/non-emptiness check, not a topology check. It would not catch a self-intersecting bowtie stroke or a subtly-wrong loop shape - only visual comparison against the reference catches that, which is what the pairwise confusable-cluster review above was for.

## Honesty check

This is a best-effort, structurally-verified pass - checked against the reference font and cross-checked pairwise for the confusable letters, but **not reviewed by a Thai speaker**. I'd treat native-speaker review as a real prerequisite before merge, same as I flagged for the Georgian PR.


dxli:
Combined into #2773 along with #2769, #2770 and #2772 - same commit, unchanged.

PR #2772, opened 2026-08-23
Add Lao base letters and combining marks to unicode.lff (stacked on #2771)

dxli:
Stacked on #2771 (Thai), which is stacked on #2770 (Georgian), which is stacked on #2769 (Latin/Cyrillic). Until the earlier ones merge, this diff includes their commits too; it narrows to just the Lao commit as each one lands.

## What

`unicode.lff` had zero Lao coverage - every Lao codepoint fell back to the replacement-character diamond. This adds:

- All 56 Lo/Lm base letters: the modern Lao consonants, the extended Pali/Sanskrit transliteration consonants, the Lo/Lm vowels and marks that place correctly with plain sequential glyph advance (sara a/aa/am, the five leading vowels, the semivowel sign nyo, ko la), and the four Khmu-language extension letters (added to Unicode for representing Khmu, a minority language in Laos, with Lao script).
- The 17 Mn combining marks: five tone marks, vowel signs I/II/Y/YY/U/UU, semivowel sign LO, cancellation mark, niggahita, yamakkan, and the Pali virama.

73 glyphs total, `unicode.lff` glyph count 2866 → 2939.

## How

Same pipeline as #2771 (Thai): rasterize each glyph from NotoSansLao-Regular.ttf (SIL OFL) with FreeType, compute its medial-axis skeleton, trace the skeleton into a graph, prune skeletonization artifacts, simplify with Ramer-Douglas-Peucker, and fit a circular arc per segment (least-squares) for the LFF bulge. Only the resulting stroke geometry is used, not any Noto data or code, so this carries no OFL obligations into `unicode.lff` (GPLv2-or-later) - same relationship the existing hand-drawn glyphs have to whatever reference font was used while drawing them.

The pipeline built for Thai carried over to Lao almost unchanged - Noto's Lao and Thai fonts share the same metric conventions (same `sCapHeight`) and broadly similar letterform construction, being related scripts. One real bug from the Thai pass recurred here in the same shape: the Pali virama (U+0EBA), like Thai's phinthu, is a small filled dot whose medial-axis skeleton collapses to ~2 pixels with nothing to trace, so it's hand-authored as a small 4-arc circle at the position FreeType reports, same as phinthu was.

## Confusable-cluster check

Lao has the same historical "W-shaped consonant family" as Thai's ผ/ฝ/พ/ฟ: PHO SUNG/FO TAM/PHO TAM/FO SUNG (ພ/ຝ/ພ/ຟ - Lao's cognates of the same Indic-derived letters). Checked pairwise the same way: the traced output differentiates them correctly (loop-both-sides / loop-left-spike-right / plain / spike-only), matching the pattern found in Thai. The four Pali TTA/TTHA/DDA/DDHA transliteration consonants (rare, low real-world usage) also came out topologically distinct on inspection.

## Known limitation: combining-mark positioning

Same as #2771: LibreCAD's text engine has no shaping and no zero-advance combining-mark placement, so the Mn marks here are drawn at their real vertical offset but currently render after the base letter's advance width instead of stacked on top of it. Plain vowel/consonant text (the large majority of real Lao text) is unaffected; only tone-marked or vowel-pointed text hits this.

## Tests

`unicode_font_tests.cpp` gained a `kLao` array and a test case, following the same envelope-check pattern and bound-widening rationale as the Thai test case (some consonants run past x=9, marks sit left of their recentered origin, four marks - sara u, sara uu, Pali virama, semivowel sign LO - sit entirely below baseline). `[unicode]` is now 1754 assertions across 6 cases; full suite is 15132 assertions / 784 cases, green.

As with Thai and Georgian: this is an envelope/non-emptiness check, not a topology check - it would not catch a self-intersecting or subtly-wrong stroke, only visual comparison against the reference does that.

## Honesty check

Best-effort, structurally verified against the reference font and cross-checked for the one confusable cluster I could identify - but **not reviewed by a Lao speaker**. Flagging native-speaker review as a real prerequisite before merge, same as Thai and Georgian.


dxli:
Combined into #2773 along with #2769, #2770 and #2771 - same commit, unchanged.

PR #2791, opened 2026-09-02
Fix tool buttons staying checked when another tool is chosen (2.2.1)

ROMPEFUEGOS:
### The bug

Since 2.2.1.1, checkable drawing-tool buttons and menu entries can remain highlighted after another tool is chosen or the action finishes. This becomes especially misleading after nested selection actions, when switching drawings, or when moving through print preview.

### Root cause

Commit `e0f5dd6b` disabled the link from a tool `QAction` to the `RS_ActionInterface` it starts. Restoring that link also requires respecting Qt signal order: `QActionGroup::triggered(QAction*)` reaches `relayAction()` before the individual action slot creates the LibreCAD action.

The QActions are application-wide, while action stacks and relative-zero lock state belong to individual graphic views. Cleanup can also run inside nested callbacks, so callback lifetimes and action-stack transitions must be handled explicitly.

### The fix

- Keep a triggered `QAction` pending until `setCurrentAction()` creates the corresponding action, and map by action instance rather than action name.
- Check only the topmost mapped action for the active drawing. Switching drawings clears the previous shared presentation and restores the active drawing tool without canceling either drawing action stack.
- Preserve tool ownership through helper actions. A selection-to-modify hand-off transfers the original QAction; helper, command-line, and relative-zero actions do not steal it.
- Keep the action stack alive while mouse, keyboard, and command callbacks run, and clean finished actions after keyboard, coordinate, and pointer input.
- Resume an underlying action only when cleanup actually removes the stack top, avoiding redundant preview redraws on ordinary mouse moves.
- Use the full `finish()` path when the Text or Image dialog is canceled.
- Disable the drawing-tool buttons in print preview while keeping Zoom Pan and Window Zoom available. The preview window tracks the QActions it starts like any other active window, so the zoom buttons are released when their action ends; the layer, edit and drawing-preferences actions keep working in the preview.
- Synchronize the relative-zero lock button from the active view so each drawing retains its own lock state.
- Safely handle a missing active window. Relative-zero actions remain available in recent actions without taking ownership of the current tool button.

### Verification

- Full CMake build passed on macOS arm64 with Qt 5, DWG support, and `RelWithDebInfo`.
- All changed translation units compile through qmake5 with C++17.
- The built application launched successfully.
- Live two-drawing checks confirmed that independent Line/Circle actions and relative-zero lock states follow their owning drawings across repeated activation.
- Live print-preview checks confirmed that drawing actions are disabled, Zoom Pan and Window Zoom remain enabled, and the parent drawing action state is restored after leaving preview.
- Source checks found no remaining Text/Image caller of the incomplete `setFinished()` path.
- The original branch evidence exercises 51 checks across 16 single-window tool, helper-action, selection, zoom, and relative-zero scenarios.
- Review follow-up: the action-handler gate that rejected every non-zoom action in print preview was removed after a code review found it also blocked Current Drawing Preferences, layer operations, undo/redo and the block-widget buttons while their controls stayed enabled; the preview handler now releases its zoom buttons; the QAction state update was moved off the per-event `hasAction()` path. Full CMake Qt 5 build passed after the change.

Detailed harness output and screenshots are on the contributor [test-evidence branch](https://github.com/ROMPEFUEGOS/LibreCAD/tree/pr-assets-tool-buttons/test-results).

Line tool state before and after the fix

### DIFF
--- a/librecad/src/actions/rs_actiondrawimage.cpp
+++ b/librecad/src/actions/rs_actiondrawimage.cpp
@@ -77,7 +77,7 @@ void RS_ActionDrawImage::init(int status) {
 
         setStatus(SetTargetPoint);
     } else {
-        setFinished();
+        finish();
         //RS_DIALOGFACTORY->requestToolBar(RS2::ToolBarMain);
     }
 }
--- a/librecad/src/actions/rs_actiondrawtext.cpp
+++ b/librecad/src/actions/rs_actiondrawtext.cpp
@@ -69,7 +69,7 @@ void RS_ActionDrawText::init(int status) {
 			showOptions();
 		} else {
 			hideOptions();
-			setFinished();
+			finish();
 		}
 	}
 		break;
--- a/librecad/src/lib/gui/rs_eventhandler.cpp
+++ b/librecad/src/lib/gui/rs_eventhandler.cpp
@@ -24,6 +24,7 @@
 **
 **********************************************************************/
 
+#include <algorithm>
 #include <QRegExp>
 #include <QAction>
 #include <QMouseEvent>
@@ -113,8 +114,6 @@ RS_EventHandler::~RS_EventHandler() {
     defaultAction.reset();
 
     RS_DEBUG->print("RS_EventHandler::~RS_EventHandler: Deleting all actions..");
-    for(const auto& [actionName, qAction]: m_toQAction)
-        qAction->setChecked(false);
     RS_DEBUG->print("RS_EventHandler::~RS_EventHandler: Deleting all actions..: OK");
     RS_DEBUG->print("RS_EventHandler::~RS_EventHandler: OK");
 }
@@ -170,15 +169,6 @@ void RS_EventHandler::mouseReleaseEvent(QMouseEvent* e) {
 
         current->mouseReleaseEvent(e);
 
-        // action may be completed by click. Check this and if it so, uncheck the action
-        if (current->getStatus() < 0){
-            if (m_toQAction.count(current->getName()) == 1) {
-                m_toQAction[current->getName()]->setChecked(false);
-                m_toQAction.erase(current->getName());
-	    }
-	}
-
-
         // Clean up actions - one might be finished now
         cleanUp();
         e->accept();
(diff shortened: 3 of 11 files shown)


dxli:
@ROMPEFUEGOS 

Could you validate the PR for scenarios:

1, create a new drawing A; 2, Line 2 points; 3, create a new drawing B; 4, Circle(C, R); 5, switch back to drawing A; 6, switch back to drawing B

2, create a new drawing A; 2, line 2 points; 3, Print Preview; 4, create a new drawing B; 4, Circle(C, R); 5, switch back to drawing A

The basic idea, active action tracking and drawing switching do not work properly.

Another idea, making all action not checkable, so no tracking is needed.

dxli:
Updated review status for commit `2032f9d70`:

- The PR branch is current with `origin/2.2.1` and GitHub reports it mergeable with a clean merge state.
- Independent live actions are retained per drawing: a Line action in drawing A and a Circle action in drawing B remain on their respective view stacks, while the shared QAction presentation follows the active drawing.
- Print Preview clears drawing-tool presentation, disables/rejects model-editing actions, preserves zoom/navigation, and switching between preview and drawing windows does not cancel another drawing's live action.
- The full CMake Qt 5/DWG macOS arm64 build passed, qmake6 configuration selected C++17 successfully, and the resulting application launched successfully.

No additional source changes were needed after the latest review.

dxli:
Deep-review follow-up pushed in `9a55dfa73`:

- synchronize the shared relative-zero lock button from the active graphic view, so drawings retain independent lock state;
- keep `Zoom Pan` and `Window Zoom` enabled in print preview, matching the action handler's existing preview whitelist while drawing/editing tools remain disabled;
- resume and redisplay options only when cleanup removes the current stack entry, avoiding duplicate print-preview redraws after ordinary mouse moves.

Verification after the update:

- full CMake Qt 5/DWG `RelWithDebInfo` build passed;
- all four changed translation units compiled through qmake5 with C++17;
- live two-drawing test confirmed independent relative-zero lock state is restored on each activation;
- live print-preview test confirmed Zoom Pan and Window Zoom stay enabled, drawing tools stay disabled, and all states restore on return to the drawing.

PR #2814, opened 2026-09-09
Fix HIDDEN2 DXF import mapping

ROMPEFUEGOS:
Fixes the DXFRW `HIDDEN2` alias: it now resolves to `DashLine2` rather than `DashLineTiny`.

Changes:
- exercise `HIDDEN2` from valid R12 LTYPE and LAYER records, plus an entity override
- prevent the analyzer-report push from running on pull-request events, where the publishing secret is unavailable

Verification:
- `QT_QPA_PLATFORM=offscreen /private/tmp/librecad-pr2814-build/librecad_dxf_fast_tests --reporter compact`
  - 515 test cases, 6,500 assertions passed

A matching one-line mapping defect remains on the `2.2.1` branch and should be backported separately.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -132,7 +132,7 @@ jobs:
           git commit -m "latest analyzer report"
 
       - name: Push changes
-        if: github.repository_owner == 'LibreCAD'
+        if: github.event_name != 'pull_request' && github.repository_owner == 'LibreCAD'
         uses: ad-m/github-push-action@master
         with:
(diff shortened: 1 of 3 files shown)


dxli (approved):

ROMPEFUEGOS:
Superseded: #2815 was merged as fc320c183 with these three commits included (the HIDDEN2 mapping fix, the strengthened regression test and the codeql.yml guard), so this one is now redundant. Closing.

Issue #1738, opened 2024-02-14
Line types customization


ghost:
Good morning, I read somewhere that in AutoCAD you can customize the line type simply through a text file that can be reached with the file explorer. Instead in LibreCAD, from what I understand, the line types are predefined, and to modify them you would have to recompile the source code ... but I'm not very familiar with these things. Could a change be made to the code so that LibreCAD works like AutoCAD? Thanks in advance for any help, have a nice day, Simone Aiello

app4soft:
To not duplicate this issue, will add details from other feature request

# Support linetype Au

### Expected behavior

LibreCAD should support custom linetype definitions using setting path to AutoCAD linetype file (`.lin`).

Topic on Forum:

- https://forum.librecad.org/Creating-Custom-Linetypes-td4974845.html
- https://forum.librecad.org/Adding-custom-Line-Types-td5723518.html
- https://forum.librecad.org/Linetypes-aren-t-working-when-I-open-LibreCAD-dxf-files-with-AutoCAD-or-eDrawings-td5706867.html

### Observed behavior

LibreCAD has linetype definitions hardcoded in `rs_linetypepattern.cpp`:

- https://github.com/LibreCAD/LibreCAD/blob/master/librecad/src/lib/gui/rs_linetypepattern.cpp

### Steps to reproduce or sample file

AutoCAD linetype file (`.lin`) samples:

- https://github.com/tuffs/acad-lin
- https://github.com/sidhrika/MyLineTypes

LIN-file syntax highlighting for Notepad++:

- https://github.com/rddim/notepad-plus-plus-autocad

### Operating System and LibreCAD version info

All versions and all OS.

### Related issues

- Support hatch definition with AutoCAD hatch pattern files (`.pat`): https://github.com/LibreCAD/LibreCAD/issues/1651

u2fly:
@dxli @lordofbikes

LibreCAD not fully supports standard ISO line types from DXF (produced by QCAD or other program) due to some unsupported line types, difference in metrics and line type substitution tables.

- ISO-128-2:2022 - https://standards.iteh.ai/catalog/standards/iso/ed38e2d0-844f-4eb5-97a9-8770b7a9acd6/iso-128-2-2022
  - Preview - https://cdn.standards.iteh.ai/samples/83355/10bb39d36fc34caeb80ecd25347ddb0c/ISO-128-2-2022.pdf
[...]

PR #2815, opened 2026-09-09
Add the HIDDEN linetype family so AutoCAD hidden lines survive a DXF round-trip (#2813)

### LINKED ISSUE #2813: HIDDEN linetypes are silently rewritten as DASHED on DXF round-trip; HIDDEN2 imports as the tiny dashed variant
**Describe the bug**

LibreCAD has no native HIDDEN linetype. `RS_FilterDXFRW::nameToLineType()` maps the acad.lin names `HIDDEN`, `HIDDEN2` and `HIDDENX2` onto the DASHED family on import, and `lineTypeToName()` writes them back out as `DASHED`, `DASHEDTINY` and `DASHEDX2`. The substitution is silent: a drawing that comes in from AutoCAD with hidden lines leaves LibreCAD with dashed lines on every entity and layer that used them. (Since abd1f4a8e the raw LTYPE *record* survives as advanced metadata and is re-emitted, so the saved file even carries an orphaned `HIDDEN` record that nothing references.)

Additionally, `HIDDEN2` is tested twice in `nameToLineType()`; the first test (paired with `DASHEDTINY`, added in 4c77bc92a in 2015 as a copy of the `DASHED2` line below it) always wins, so `HIDDEN2` — acad.lin `A,.125,-.0625`, the half-scale hidden line — imports as `DashLineTiny` (0.15x) instead of `DashLine2` (0.5x). The JWW filter has the intended mapping; only the DXF/DWG path is wrong.

**To Reproduce**

1. Save the minimal DXF below (one LINE with linetype `HIDDEN2`) and open it in LibreCAD.
2. Select the line and look at the Properties dock: *Line Type* reads "Dash (tiny)".
3. Save as DXF: the LINE carries `6/DASHEDTINY`.
4. Same file with `HIDDEN` instead of `HIDDEN2`: reads "Dash", saved as `DASHED`.

```
0
SECTION
2
ENTITIES
0
LINE
8
0
6
HIDDEN2
10
0.0
20
0.0
11
10.0
21
0.0
0
ENDSEC
0
EOF
```

**Expected behavior**

Hidden lines keep their name and their acad.lin pattern across a round-trip, and `HIDDEN2` is the 0.5x variant, not the 0.15x one.

**Additional context**

master at 5822ae887 (Linux, Qt 6.8.2 via pixi); the mapping is the same on 2.2.1. Related to #1738 (loading `.lin` files) but independent: the built-in substitution table is wrong on its own. Two PRs follow: (1) the one-line `HIDDEN2` mapping fix with a regression test, and (2) HIDDEN as a first-class family (`HIDDEN`, `HIDDENTINY`, `HIDDEN2`, `HIDDENX2`) following the convention of ca7c096e6, with the enum values appended after `LineSelected` so persisted settings, `.lcp` palettes and `$DIMLTYPE` stay valid.

ROMPEFUEGOS:
### The bug

LibreCAD has no native HIDDEN linetype. On import, `RS_FilterDXFRW::nameToLineType()` maps the acad.lin names `HIDDEN` / `HIDDEN2` / `HIDDENX2` onto the DASHED family, and on save `lineTypeToName()` writes them back as `DASHED` / `DASHED2` / `DASHEDX2` (`DASHEDTINY` for `HIDDEN2` before #2814). The substitution is silent: a drawing that comes in from AutoCAD with hidden lines leaves LibreCAD with dashed lines on every entity and layer that used them. Since abd1f4a8e the raw LTYPE *records* survive in `LC_DwgAdvancedMetadata` and are re-emitted, so the saved file even carries an orphaned `HIDDEN` record that nothing references any more. Layer standards and other CAD packages key on that name.

Reproduce: open a DXF with one `LINE` whose group 6 is `HIDDEN` (fixture `single_HIDDEN.dxf` below), select it: the Properties dock reads *Dash*. Save as DXF: the `LINE` carries `6/DASHED`.

### Root cause

`RS2::LineType` is a closed catalogue of seven families × four variants, and the pen stores only the enum, so a name that has no enum value cannot survive a round-trip. `HIDDEN*` are aliases of `DashLine*` in the substitution table; there is nothing to write them back to.

### The fix

HIDDEN becomes a first-class family, following the convention set by ca7c096e6 (2014, tiny variants for every family) and 4c77bc92a (2015, DXF persistence of the tiny variants):

- `RS2::HiddenLine`, `HiddenLineTiny`, `HiddenLine2`, `HiddenLineX2` = 28..31, **appended after `LineSelected`**. The numeric values are persisted in QSettings (`rs_settings.cpp`), in `.lcp` pen palettes (`lc_penpalettedata.cpp`) and in `$DIMLTYPE` (`lc_dimstyletovariablesmapper.cpp`), and are cast straight into the plugin interface (`doc_plugin_interface.cpp`), so inserting into the middle of the enum is not an option; a comment in `rs.h` now says so.
- LTYPE metrics straight from acad.lin × 25.4, the same derivation the DASHED family uses: `HIDDEN` 6.35/−3.175, `HIDDEN2` 3.175/−1.5875, `HIDDENX2` 12.7/−6.35, `HIDDENTINY` 0.9525/−0.47625. `HIDDEN` is therefore numerically `DASHED2` and `HIDDENX2` is `DASHED`; the point of the family is to preserve the **name**, not to look different.
- Screen patterns with the same rounded-millimetre scheme as DASHED (12/6 for 12.7/6.35), so HIDDEN draws as 6/−3. `rs_painter.cpp` needs no change: its default branch renders any registered pattern. **Visible change:** drawings that carried `HIDDEN` / `HIDDEN2` / `HIDDENX2` were rendered with the DASHED patterns (12/−6, 6/−3, 24/−12 on screen); they now render at half that length (6/−3, 3/−1.5, 12/−6), which is what acad.lin's ratios say and what AutoCAD shows.
- `nameToLineType()` / `lineTypeToName()` map the four names both ways; `writeLTypes()` emits the four records, which the DWG writer (`dwgWriter15`) and the DIMSTYLE writer also resolve against.
- `writeLTypes()` used to skip re-emitting imported raw LTYPE records by checking a hand-written set of the 27 built-in names. Adding the new records without touching that set would have written `HIDDEN` twice for a file that already carried it (caught by the GUI round-trip below), so `writeLType()` now records the names it writes and the skip set is derived from that.
- JWW: the name tables gain the same three names DASHED has (that filter never carried tiny variants), and the LTYPE range loop in `fileExport()`, which stopped at `BorderLineX2`, now also covers the new block. JWW export is not reachable from the UI and `DL_Jww::writeLineType()` is an empty stub, so this is for consistency.
- MakerCAM SVG export gets the four cases; the hidden ladder is the dash ladder halved, mirroring acad.lin.
- Pen toolbar combobox and pen palette list *Hidden*, *Hidden (tiny)*, *Hidden (small)*, *Hidden (large)* right after the Dash family, with their own icon (`linetype08.svg`, the Dash icon with half the dash length). The property sheet inherits both.
- The plugin string bridge (`convLTW` in `doc_plugin_interface.cpp`, the `DPI::LTYPE` property that e.g. the *sameprop* plugin copies between entities) gains the three names, like the DASHED entries next to it; without them a HIDDEN entity would have reached plugins as `BYLAYER` where it used to say `DashLine`.

Left alone on purpose: the public `DPI::LineType` enum in `document_interface.h` (it already lacks the tiny variants and is cast to/from `RS2::LineType` without translation — deserves its own fix), `rs_filterdxf.cpp` (not in the CMake build) and `rs_python_wrappers.cpp` (behind `RS_OPT_PYTHON`). Compatibility note: like every enumerator added before, a pen set to a hidden type in QSettings and then read by an older LibreCAD is not validated on the way in (the `FIXME` in `rs_settings.cpp`); `.lcp` palettes go through `LC_PenInfoRegistry::hasLineType()` and reject it cleanly. This does not implement `.lin` loading (#1738); it fixes the built-in substitution table, which stays the fallback once `.lin` support exists.

Fixes #2813. Builds on #2814 (the `HIDDEN2` mapping fix): that commit is the first of the two here, so this PR shows both until #2814 is merged.

### Verification

Built with `pixi` (conda-forge Qt 6.8.2, GCC 13.3, the same recipe as `run_pixi.yml`) from `origin/master` with and without this change (RelWithDebInfo, `-DBUILD_TESTS=ON`; no warnings). Unit tests and the two CI lanes were run at 5822ae887; the GUI scenarios at b8da4aaf2, four commits earlier, none of which touches linetypes.

Unit (`dxf_roundtrip_tests.cpp`, part of `librecad_dxf_fast_tests`, the lane run by `dwg-dxf-verify.yml`): three new cases — a `LINE` with pen `HiddenLine` exported through `RS_FilterDXFRW` carries `6/HIDDEN`, the LTYPE table has `HIDDEN` with path 6.35/−3.175, importing it back (DXF, and DWG under `DWGSUPPORT`) yields `HiddenLine`, and saving the reimported drawing writes the record once; the `HIDDEN2` regression test now expects `HiddenLine2`; every name `lineTypeToName()` writes maps back to the same enum through `nameToLineType()`, and every hidden variant has a screen pattern. `librecad_dxf_fast_tests`: 517 cases passed; `librecad_dwg_fast_tests`: 909 passed, 37 skipped (fixtures), 0 failed.

GUI, Qt `vnc` platform driven by the harness in `tools/` (pen toolbar, command line, *File › Save as*, Properties dock; saved files checked with `check_dxf_linetype.py` / ezdxf):

| scenario | master (b8da4aaf2) | with this change |
|---|---|---|
| G1 pen toolbar line-type list, entries after *Dash (large)* | Dash Dot… | Hidden, Hidden (tiny), Hidden (small), Hidden (large) |
| G2 pick *Hidden*, draw a line, *Save as* → LTYPE table + group 6 | (no such entry; line saved ByLayer) | `HIDDEN` record 6.35/−3.175, `LINE` with `6/HIDDEN` |
| G3 reopen the G2 file, select, Properties dock | — | Hidden |
| G4 external `single_HIDDEN.dxf` / `single_HIDDEN2.dxf` / `single_HIDDENX2.dxf`, Properties dock | Dash / Dash (tiny) / Dash (large) | Hidden / Hidden (small) / Hidden (large) |
| G5 external `hidden_family.dxf` (3 lines + a layer using HIDDEN), *Save as* → group 6 of the lines, layer linetype, LTYPE records | `DASHED` / `DASHEDTINY` / `DASHEDX2`, layer `DASHED`; orphaned `HIDDEN*` records kept | `HIDDEN` / `HIDDEN2` / `HIDDENX2`, layer `HIDDEN`; 31 LTYPE records, no duplicates |

pen toolbar line-type list, base vs fix

Properties dock on the single-line fixtures, base vs fix

Full set — `results.json`, all screenshots, the fixtures, the saved DXF files and the harness: [branch `pr-assets-hidden-linetype`](https://github.com/ROMPEFUEGOS/LibreCAD/tree/pr-assets-hidden-linetype/test-results/hidden) of the fork.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012gqZ2BjNbthdqRGX2YDugY

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -132,7 +132,7 @@ jobs:
           git commit -m "latest analyzer report"
 
       - name: Push changes
-        if: github.repository_owner == 'LibreCAD'
+        if: github.event_name != 'pull_request' && github.repository_owner == 'LibreCAD'
         uses: ad-m/github-push-action@master
         with:
           repository: ${{github.repository_owner}}/static-analyzer-reports
--- a/librecad/res/controls/controls.qrc
+++ b/librecad/res/controls/controls.qrc
@@ -19,6 +19,7 @@
         <file alias="linetype05.lci">linetype05.svg</file>
         <file alias="linetype06.lci">linetype06.svg</file>
         <file alias="linetype07.lci">linetype07.svg</file>
+        <file alias="linetype08.lci">linetype08.svg</file>
         <file alias="width00.lci">width00.svg</file>
         <file alias="width01.lci">width01.svg</file>
         <file alias="width02.lci">width02.svg</file>
--- a/librecad/res/controls/linetype08.svg
+++ b/librecad/res/controls/linetype08.svg
@@ -1,0 +1,1 @@
+<svg height="12" viewBox="0 0 8.4666665 3.175" width="32" xmlns="http://www.w3.org/2000/svg"><path d="m0 1.4597563h8.4949593" fill="#427bc3" stroke="#00000f" stroke-dasharray="1.35 .45" stroke-dashoffset="0" stroke-width=".9"/></svg>
--- a/librecad/src/lib/engine/rs.h
+++ b/librecad/src/lib/engine/rs.h
@@ -864,7 +864,15 @@ namespace RS2 {
         BorderLineX2   = 25, /**< dash, dash, dot large. */
 
         LineTypeUnchanged = 26, /**< Line type defined by block not entity */
-        LineSelected      = 27 /**< Line type for selected */
+        LineSelected      = 27, /**< Line type for selected */
+
+        // Append new line types after this point. The numeric values are
+        // persisted (QSettings, .lcp pen palettes, $DIMLTYPE) and handed to
+        // plugins, so existing values must never be renumbered.
+        HiddenLine     = 28, /**< hidden line (acad.lin HIDDEN). */
+        HiddenLineTiny = 29, /**< hidden line tiny */
+        HiddenLine2    = 30, /**< hidden line small. */
+        HiddenLineX2   = 31  /**< hidden line large. */
     };
 
     /**
--- a/librecad/src/lib/filters/rs_filterdxfrw.cpp
+++ b/librecad/src/lib/filters/rs_filterdxfrw.cpp
@@ -31665,7 +31665,7 @@ RS2::LineType RS_FilterDXFRW::nameToLineType(const QString &name) {
       uName == "DASHED" || uName == "HIDDEN") {
     return RS2::DashLine;
   }
-  if (uName == "DASHEDTINY" || uName == "HIDDEN2") {
+  if (uName == "DASHEDTINY") {
     return RS2::DashLineTiny;
   }
   if (uName == "DASHED2" || uName == "HIDDEN2") {
--- a/librecad/src/lib/filters/rs_filterdxfrw.cpp
+++ b/librecad/src/lib/filters/rs_filterdxfrw.cpp
@@ -18332,6 +18332,7 @@ void RS_FilterDXFRW::writeLType(const UTF8STRING &lTypeName,
               lTypeName)) {
     ltype = *source;
   }
+  m_builtinLTypeNames.insert(normalizeDwgTableName(lTypeName));
   (void)writeLTypeRecord(ltype);
 }
 
@@ -18360,6 +18361,7 @@ bool RS_FilterDXFRW::writeLTypeRecord(DRW_LType &ltype) {
 }
 
 void RS_FilterDXFRW::writeLTypes() {
+  m_builtinLTypeNames.clear();
   writeLType("CONTINUOUS", "Solid line", 0, 0, {});
   writeLType("ByLayer", "", 0, 0, {});
   writeLType("ByBlock", "", 0, 0, {});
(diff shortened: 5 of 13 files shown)


ROMPEFUEGOS:
About the red *Security scan* check: the analysis itself completed ("Build and analyze" and "Create report repository" are green); what fails is the last step, *Push changes*. Its guard `if: github.repository_owner == 'LibreCAD'` is also true for pull requests, but `secrets.LC_PUSH_ANALYZER_REPORT` is empty for PRs from forks, so `ad-m/github-push-action` exits with `Missing input 'token'`. The same happens on #2811 and #2812 (opened from a fork), while PRs from in-repo branches such as #2810 pass. Adding `&& github.event_name != 'pull_request'` to that guard (or pushing the report only on `push` to master) would keep the check green for fork PRs. Nothing in this PR's code is involved.

PR #2816, opened 2026-09-09
Fix NUL-padded escapes, unreachable codepage mappings, and Big5 coverage

dxli:
Five text-codec fixes, found while checking libdxfrw's codepage tables against
the WHATWG Encoding indexes. The first is a data-corruption bug; the rest are
mappings that were present but unreachable, plus the coverage and performance
work that follows from them.

### 1. `\U+` escapes carried 12 NUL bytes on macOS

A code point the target codepage cannot hold is written as AutoCAD's `\U+XXXX`
escape. On Apple platforms `decodeText()` built it as:

```cpp
std::string str(16, '\0');
snprintf(&(str[0]), 16, "%04X", c);
num = str;          // std::string keeps its size: 16, not 4
```

so the escape came out **19 bytes with 12 trailing NULs** instead of 7. Other
platforms took a `stringstream` branch and were correct.

The damage lands on write, and it fires for any character outside the file's
codepage — a CJK layer name in an `ANSI_1252` drawing is enough:

| output | result |
| --- | --- |
| ASCII DXF | `isSafeAsciiDxfString` rejects NUL → write refused, **text silently lost** |
| binary DXF | `isSafeBinaryDxfString` → same |
| **DWG ≤ R2004** (`putCP8Text`) | **no guard — the NULs go into the file** |
| DWG ≥ R2007 | unaffected (UTF-16 path) |

### 2. windows-1255 byte 0xCA

The Hebrew points are a contiguous run, `0xC0+n → U+05B0+n`. `0xCA` was the
only gap — because U+05BA did not exist when windows-1255 was defined. Unicode
5.0 added it. Microsoft, `iconv` and Python still leave it unmapped, but the
arithmetic of the surrounding run makes the answer unambiguous. An unmapped
byte decodes to *nothing* here, so the string silently shortened.

### 3. 98 double-byte mappings were unreachable

The DBCS decoder scans the range its lead byte owns,
`[leadTable[c-0x81], leadTable[c-0x80])`. `leadTable` is derived data, and it
had drifted out of step with the entries it indexes:

| table | misbucketed | past the sentinel |
| --- | --- | --- |
| cp936 | 56 | 0 |
| cp949 | 30 | 1 |
| cp950 | 10 | 1 |

An entry in the wrong bucket never matches (the comparison includes the lead
byte); an entry past the sentinel is never visited. So 98 characters that were
present in the tables decoded as `?`. The lead tables are regenerated from the
entries themselves — double tables untouched, only offsets change — and each
gains a 128th entry so a stray `0xFF` lead byte from a damaged file cannot read
off the end.

### 4. Big5 gains the hkscs sequences

5,087 sequences that decoded as `?` are added from the WHATWG big5 index.
Only absent sequences are added; **no existing reading changes.** That matters
at `0xF9FE`, where cp950 says U+2593 and big5-hkscs says U+FFED — a genuine
vendor fork rather than a gap, so the cp950 reading stands.

One caveat, stated plainly: 1,713 of the additions (a third) are above
U+FFFF and will show as missing glyphs until the font engine handles
supplementary code points.

Four further sequences (pointers 1133/1135/1164/1166) stand for **two** code
points each — U+00CA/U+00EA plus a combining U+0304/U+030C, the Cantonese
romanisation Ê̄ Ê̌ ê̄ ê̌. They are absent from `index-big5.txt` because, in the
standard's words, indexes are limited to single code points. The table cell
packs the pair as `(first << 16) | second`, which cannot be mistaken for a
real code point: the largest in any of these four tables is U+2F9D4 and the
packed values start above `0x00CA0000`.

Encoding stays asymmetric, which is the standard's design rather than a
shortcut — the Big5 encoder looks up one code point at a time and errors when
there is no pointer. So Ê̄ writes as `88 66` plus `\U+0304`: different bytes,
same text on re-read, where before the character was lost to `?`.

### 5. The double-byte reverse lookup is indexed

`fromUtf8` found the byte pair for a code point by walking the entire double
table per character — 18,594 comparisons for Big5 after the hkscs rows, paid
on every save of a CJK drawing. It now builds the reverse map once per
converter (converters are per file, not per string), filled in table order with
`emplace()` so a repeated code point keeps the first mapping the scan returned.

Encoding 2000 CJK characters to Big5, measured against the code and table this
replaces:

| character position | before | after |
| --- | --- | --- |
| near the start of the table | 0.315 ms | **0.037 ms** |
| near the end | 7.454 ms | **0.034 ms** |

### Verification

Full suite: 2417 cases / 2320 passed / 28 failed / 69 skipped;
72592 assertions / 32 failed — the master baseline unchanged. The deltas are
this PR's own tests: 10 test cases carrying 60 assertions, which is exactly the
TEST_CASE count added over master.

Every one of the 13,503 original cp950 mappings was checked byte-identical
after regeneration, and the four two-code-point rows were added on top of the
18,590 the hkscs commit produced with none changed or removed.

### 6. Two gaps found reviewing the above

**`\U+` cannot express a supplementary code point.** The escape carries exactly
four hex digits and `encodeText()` reads exactly four, but `%04X` is a minimum
width, not a maximum. `U+20021` emitted five digits and read back as `U+2002`
followed by `'1'` — one character silently becoming two wrong ones. The bug
predates this branch, but **item 4 made it reachable**: Big5 gained 1,713
supplementary mappings where such a sequence previously decoded to `?` and never
produced one. It now falls back to the same `?`, which is honestly lossy rather
than quietly wrong.

**cp932 was left with the scan.** Item 5 covered `DRW_ConvDBCSTable` (cp936,
cp949, cp950) and missed `DRW_Conv932Table`, which kept scanning all 7,724
entries per character. The index moves into their shared `DRW_Converter` base
rather than being copied into the second class. Encoding 2000 `U+9F8D` to cp932:
**2.215 ms → 0.037 ms**, same bytes out.

cp932's lead table was checked for the drift the other three had and is clean —
61 entries, sentinel 7724 exactly matching the table length, no misbucketed rows.


dxli:
Superseded by #2811, which now merges this branch in full — all ten commits from `fix/textcodec-escape-and-tables` are contained there (head `d1e75bdf0` is an ancestor of that branch), so the NUL-padded `\U+` escape fix, the indexed double-byte reverse lookup, the windows-1255 0xCA mapping, the Big5/HKSCS and UHC table work and their tests all carry over unchanged.

Closing to keep the review in one place rather than splitting it across two PRs. The branch is left intact in case this needs to be reopened or split back out.

PR #2825, opened 2026-09-10
Add selectable LibreCAD selection pointer

dxli:
## Summary
- Add a crosshair selection pointer based on the design in #1487, as an opt-in Appearance setting (Selection pointer: System arrow / LibreCAD crosshair).
- The system arrow stays the default. The setting only replaces the arrow cursor, which drawing views show when no command sets its own cursor; command cursors are unchanged.
- A preference change applies immediately to views showing the arrow.
- The pointer is a white line with a black outline, 100 px with a 21 px pick box and a centre hotspot, so it shows on both dark and light backgrounds and in print preview.

## Verification
cmake (Qt 6), `QT_QPA_PLATFORM=offscreen`:
- `librecad_tests "[cursor]"`: passes. It checks the arrow and the crosshair following the setting, the centre hotspot, and that the crosshair has both dark and light pixels; with the earlier near-white artwork the contrast check fails.
- Full `librecad_tests`: the same 26 failed assertions in the same 22 test cases as master at 717e8dee8.
- `qg_graphicview.cpp` compiles with the qmake build's flags; `controls.qrc` is already listed in `src.pro`.


dxli:
Validation update: qmake6 6.11.2 completed a clean out-of-source build of `librecad.pro` on macOS/arm64 with `make -j6`. The resulting `LibreCAD.app` starts successfully for `--help`, and qmake generated `:/cursors/librecad_selection_cursor.png` in `qrc_controls.cpp`.

sand1024:
@dxli 
I'm slightly concerned that such cursor may be slightly misleading for users from AutoCAD and some other systems... as far as I remember, the size of cursor box corresponds to the current snap range. I need to check docs for details, but I'm sure there is some relation... 

May be it's better to show a custom overlay (like for snap indicator) with size that is based on current snap range? 
Or, what is even better - generate cursor bitmap dynamically based on it (this should work on runtime faster I suppose, yet will require handling switch between screens)?

Issue #2804, opened 2026-09-07
[BUG] Loop/Race Condition when using Fillet Feature causing "Freeze"


pitkz:
**Describe the bug**
Program gets into a Loop/Race when using Fillet Feature causing "Freeze" (Super Lag -> unusable -> OS suggests killing)

**To Reproduce**
1. Draw some rectangle
2. Select Fillet Feature
3. Program starts lagging, you can see the coordinates of the cursor being updated a couple of times in the GUI, but than it freezes.
4. Console with "-d 6" Debbuging shows:
./LibreCAD-2.2.2_alpha1-628-g3c028612d.AppImage -d 6
[...]
[so much that my console overflows immediately, cant show the last user made "select Fillet" action of the debbuging logs]
entity: getDistanceToPoint: OK
RS_EntityContainer::getDistanceToPoint: OK
RS_EntityContainer::getNearestEntity: OK
RS_Snapper::catchEntity: not found
RS_Snapper::catchEntity
RS_Snapper::catchEntity
RS_EntityContainer::calculateBorders
RS_EntityContainer::calculateBorders: size 1: -20000000000.000000,-20000000000.000000
RS_EntityContainer::calculateBorders: size: 0.000000,0.000000
RS_Snapper::catchEntity
RS_EntityContainer::getNearestEntity
RS_EntityContainer::getDistanceToPoint
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
entity: getDistanceToPoint: OK
entity: getDistanceToPoint
entity: 7
[...]

PR #2828, opened 2026-09-10
Improve snapping performance for Text and MText

dxli:
## Summary
- Nearest-entity queries on a drawing (`getDistanceToPoint()` / `getNearestEntity()` on a graphic or block) go through the entities in drawing order, and skip every entity whose cached borders lie farther away than the nearest entity found so far. Hover catching, the context menu, and on-entity, intersection and distance snapping all go through it, so a mouse move no longer measures every entity and every Text/MText glyph (#2804). Nothing is cached beyond each entity's existing borders.
- The skip knows where borders cannot be trusted. Construction lines reach past them, and so do lines on a construction layer, including polyline, spline and leader segments and the contents of block references. Containers are searched for such lines only when the drawing has a construction layer or a block definition holding a construction line, so far block references are not walked on every query.
- A typed `RS_Snapper::catchEntity()`, and on-entity snapping with Free snap, measure only the visible top-level entities that can lie within the catch or snap range, collected by `RS_EntityContainer::appendNearby()`. With Free snap, a point beyond the snap range leaves no key entity for Delete Free to act on. A typed catch collects its candidates with `LC_ContainerTraverser` and never opens Text/MText; at `ResolveNone` it returns polylines, block references and dimensions themselves. Exact ties still go to the entity drawn last and, across the requested types, to the type listed last.
- Selection by center point applies to entities of the drawing itself only. The center of an arc inside a polyline, block reference or dimension no longer picks that container, so hover and the context menu agree. Top-level arcs, circles and ellipses are still picked by their center.
- Distance queries at `ResolveAllButTexts` and `ResolveAllButTextImage` skip text, so Intersection snap, Trim's limit pick, Info > Properties with children and Polyline segment type no longer land on glyph geometry. Fix `LC_ContainerTraverser`'s inverted `ResolveAllButTextImage` rule, which now yields text whole.
- New `RS_Information::isEditable()`: the generated children of block references, dimensions, hatches, text and splines are not editable. `isTrimmable()` uses it; Fillet and Bevel require both picks to be trimmable; Line Join requires editable lines; Break/Divide, Line Gap, Slice/Divide, Circle by Arc and Line Angle Rel do not divide or replace such children. Fillet on a line inside a block reference used to mark the generated child deleted, and Undo then freed it while the undo history still held it.
- Borders stay current when geometry changes in place: `RS_Circle::setCenter()`/`setRadius()`, the `RS_Ellipse` setters (including `setReversed()`), `RS_Arc::setReversed()`/`setData()`, `RS_Point::setPos()`, the `LC_Hyperbola` setters, `RS_Dimension::updateDim()` and `RS_EntityContainer::stretch()`. `LC_Hyperbola::calculateBorders()` finds the extrema of rotated arcs; it searched along unrotated axes, so the borders missed part of the curve. `RS_Ellipse::scale()` calculates the borders of the scaled ellipse instead of scaling the old ones, which was wrong for non-uniform and negative factors.
- `RS_Spline` rebuilds the lines it is drawn and hit-tested with after move, rotate, scale, mirror and shear; they stayed at the old place, so a moved spline was drawn where it had been and could not be caught anywhere. `RS_Spline::setDegree()` rebuilds the knot vector and the wrapping of closed splines for the new degree; before, the spline failed validation and a rebuild drew nothing.
- `LC_SplinePoints::offset()` rebuilds the control points, so Draw Parallel on a spline through points no longer creates a curve with no control points and invalid borders. Its length is computed for splines with fewer than four control points too.
- `RS_Spline` grips are its control points: the nearest reference point is the nearest control point, and moving it moves that control point and rebuilds the spline. The nearest reference point came from the ends of the drawn lines, so inner control points could not be grabbed, and dragging an end moved only a drawn line until the next rebuild.
- `RS_Circle::scale()` calculates the borders and length of the scaled circle. It scaled the old borders, which a negative factor turned inside out, and kept the old length.
- `LC_DimArrow` rotates about the given center, and its transforms carry the arrow's angle and the point its stem is drawn to. The position was rotated about the origin, away from the rotated outline, and a copy of a transformed arrow pointed the old way.
- "Is Reversed" in the property sheet (arcs) and in the ellipse dialog reverses the direction and keeps the shape, as the arc dialog already did. It set only the flag, which turned the arc into the rest of the circle or ellipse.
- `RS_Ellipse::getNearestPointOnEntity()` considers only points on the ellipse, and refines the nearest one by Newton steps on the elliptic angle. The candidates are the points with the cosine of each root of the quartic, plus the stationary points for a point on an axis, where the quartic has double roots that its solver can miss. Before:
  - at a major-axis vertex the distance came out as the maximum, or the center distance;
  - inside the ellipse on the major axis it missed the nearest points (200 instead of 180.28 at (-300, 0) on a 500 × 250 ellipse);
  - near the vertices it returned no distance, distances below the true one, or distances up to 2603 too large.

  Against a reference that does not use the quartic, on 27,600 points on, near and away from the axes, the distance is now within 1e-12.
- Line Snake's polyline conversion catches top-level lines only.

## Testing
- `cmake --build <build> --target librecad_tests`
- `QT_QPA_PLATFORM=offscreen ./librecad_tests -# "[snap],[catch],[container],[actions],[round],[preview],[information],[onentity],[rs_hatch],[hyperbola],[#rs_ellipse_tests],[#rs_spline_tests],[#lc_snappercatchtests],[#lc_circletests],[#lc_arctests],[#lc_dimarrowtests]"` (142 test cases pass)
- New `librecad/src/lib/actions/tests/lc_snappercatchtests.cpp` covers:
  - what typed and untyped catches return at each level (polylines, block references, Text/MText, hidden polylines);
  - ties by drawing order and by requested type;
  - a randomized drawing where the pruned search picks the same entity at the same distance as measuring every entity, at every level;
  - arc centers at top level and inside a polyline;
  - construction lines and construction layers (lines, polyline and spline segments, block references);
  - rotated hyperbolas;
  - entities edited in place (circles, points, reversed and replaced arcs, reversed elliptic arcs);
  - a spline moved by Modify > Move;
  - the major vertex of an ellipse inside a block reference;
  - on-entity snapping (text skipped, snap range with and without Free snap, key entity, inside a solid hatch);
  - `isEditable()`/`isTrimmable()`;
  - Fillet, Line Gap, Arcs to Lines, Spline from Polyline and Lengthen on drawing geometry versus generated geometry.
- `lc_hyperbola_tests.cpp`: borders of rotated arcs on both branches, and setters keep the borders current.
- `rs_ellipse_tests.cpp`:
  - nearest points for the vertices and for points on both axes, for several ratios and rotations, against a sampled minimum;
  - exact distances for points offset along the normal near each vertex (15 of these 180 points missed by more than 1e-9 without the Newton refinement);
  - `scale()` with uniform, non-uniform and negative factors, and `setReversed()`, against freshly calculated borders and length;
  - `revertDirection()` keeps the shape of an elliptic arc and swaps its ends.
- `rs_spline_tests.cpp`:
  - move, rotate, scale, mirror and shear rebuild the drawn lines;
  - `setDegree()` on open and closed splines;
  - the nearest reference point and `moveRef()` work on the control points, on open and closed splines;
  - `LC_SplinePoints::offset()` keeps control points and borders;
  - the length of a spline through points with three control points.
- New `lc_circletests.cpp`: `scale()` with uniform, non-uniform and negative factors keeps borders and length current.
- New `lc_arctests.cpp`: `revertDirection()` keeps the shape, swaps the ends and leaves the cached points consistent.
- New `lc_dimarrowtests.cpp`: a rotated, moved or mirrored arrow is where expected, and matches a copy built from its position and angle.
- The action test scaffolding shared with `lc_action_modify_round_tests.cpp` moved to `librecad/src/test/lc_actiontestsupport.h`.
- Fast suite (`librecad_tests` with the CMake fast-test excludes): 2417 test cases, 2378 passed, 21 skipped (missing DWG fixtures), 18 failed. All 18 failures are in `dwg_write_smoke_tests.cpp` (DWG export), which this change does not touch.
- Cost per whole-drawing query in an unoptimized build, with RS_DEBUG at `D_NOTHING` as the application runs, measuring every entity vs the pruned search. Both pick the same entity at the same distance at every sampled point.

  | Drawing | Every entity | Pruned |
  |---|---|---|
  | 100k short lines | 16.5 ms | 14.5 ms |
  | 73k lines, rectangles, circles and arcs | 13.6 ms | 10.6 ms |
  | 5k texts of 20 glyph lines + 20k lines | 21.7 ms | 3.8 ms |
  | 2k block references of 50 lines + 20k lines | 21.3 ms | 3.3 ms |
  | Same, with a construction layer | 21.5 ms | 5.3 ms |
- `git diff --check`


sand1024:
@dxli 
Hi! 
Just noticed your comment regarding adding cache - and wondered why you've decided to limit it to snapping only and invalidate it often?  

I've also considered adding spatial index - yet I supposed that it could be used for lot's of tasks - snapping, catching entities, selection of icons for drawing (querying ones that are within the viewport or within bound - like for selection by intersection and so on). 

So it could be built on the start (when the document is read) and be updated as the document is edited.

I think there is a good place to update the cache (after initial fill) - it is where undo cycle is committed - so entities may be added/removed to the cache (of course, this requires that there are no direct modifications of the document's structure, yet that's ok). 

Potentially, it's possible to support per-layer caches, yet it's hard to say whether it's practical without performance measuring. 

So I'm sure that there is a good room for improvements there - from the first glance, caching should increase the overall performance and allow handling larger drawings. Yet of course, it has to be measured for performance numbers.

PR #2854, opened 2026-09-14
Compare the dash offset only when a dash pattern is drawn

ROMPEFUEGOS:
Builds on #2853, which is merged into this branch. Review the diff against #2853's head
(`071a41e64..a427ad839`): 8 files, +312/−16. The branch also carries #2852 (already on master)
and merge commits, so a squash merge is simplest.

### The bug

The renderers skip pen setup when `m_lastPaintEntityPen.isSameAs(pen, patternOffset)` holds.
The last term compared the cached pen's dash offset, which is the resolved pen's offset and
normally `0`, with the painter's **running** offset, which `RS_Painter::updateDashOffset()`
decrements for every dashed entity drawn. After the first dashed entity the two never match,
so every later entity misses the cache, including solid ones whose offset changes nothing.

Measured by the author with a gdb breakpoint on `RS_Painter::setPen()`, one print pass of
2000 lines with identical pens (on the earlier revision 34ccbf2e6):

| Fixture | master | this fix |
|---|---:|---:|
| all solid | 1 | 1 |
| one dashed last, 1999 solid | 2 | 2 |
| one dashed first, 1999 solid | 2000 | 2 |
| all dashed | 2000 | 2000 |

The same revision rendered 31 images byte-identical to master. Neither table was re-measured
on the current revision.

### The fix

- `RS_Pen::isSameAs()` gains `comparePatternOffset`; the offset is compared only when it is set.
- Each renderer records, right before `painter->setPen()`, whether the pen it installs has a
  dash pattern (`m_lastPaintedPattern`), and passes that on the next request. This is the pen
  after the selection and highlight overrides: a selected entity is drawn with `DashLineTiny`
  whatever its own line type. A cache hit needs the same resolved pen and the same
  highlighted/selected/overlay/visual-snap flags, so the previous answer is exact.
- The cached pen stores the running offset it was installed at, so a dashed pen is kept while
  the offset has not moved.
- The resolved pen is cached with `updateBy()` before it is modified, instead of being copied.
- `LC_GraphicViewRenderer` invalidates the cache after `setPenForOverlayEntity()`'s direct pen
  arms (reference and snap marks) and after overlay drawables (crosshair, info cursor, UCS marks)
  draw, since those set the painter's pen without the cache. Keeping solid pens after a dash
  would otherwise let the next overlay entity keep such a pen.

Applies to `LC_GraphicViewRenderer` (normal and draft), `LC_PrintViewportRenderer` and
`LC_PrintPreviewViewRenderer`.

### Tests

- `rs_pen_tests.cpp`: `isSameAs()` with and without the offset term.
- `lc_graphicviewrenderer_tests.cpp`:
  - a run of selected entities keeps advancing the selection dash phase;
  - on the widget, draft and print paths, a solid pen is kept after the running offset moves,
    and a dashed pen is kept only while the offset stays where it was installed;
  - a snap mark or an overlay drawable invalidates the cache.

### Verification

cmake (Qt 6), `QT_QPA_PLATFORM=offscreen`:

- `librecad_tests` `[pen],[painter]`: all 90 assertions in 11 test cases pass.
- Full `librecad_tests`: the same 26 failed assertions in the same 22 test cases as master
  at 717e8dee8; this branch only adds passing tests.
- The new renderer tests fail against the previous head of this PR (5 assertions in 2 test
  cases): the invalidation cases, and the unmoved dashed pen on all three renderers.
- The changed renderer sources compile with the qmake build's flags.


ROMPEFUEGOS:
@dxli, one case is still wrong after #2854. While panning, texts are drawn as draft by default, and `setPenForEntity()` and `setPenForDraftEntity()` (screen width 0) share the pen cache. When a text and a line with the same solid pen are drawn one after the other, the second reuses the pen set up for the first: a line after a text comes out 1 px wide, and a text after a line gets the line's width. Before #2854 this stopped once a dashed line had been drawn in the pass; now it does not.

[`546f28f14`](https://github.com/ROMPEFUEGOS/LibreCAD/commit/546f28f14b113f89b1f9777f37655f5ce0ab5c7a) on `ROMPEFUEGOS:fix/pen-cache-draft-key` records which of the two installed the cached pen (+7/−2), with a test that fails without it. Happy to open a PR. One side effect: in "Draft" mode with `Highlight with screen linewidth in "Draft" mode` on, the parts of a highlighted block or dimension that share its pen are now always drawn at 1 px, as they already are with `Show entity handles on mouse hover` on, the default.

Checked on 8d9c8760f and again on e27468774, with and without the commit:
- Tests: no new failures, and the new case fails without the fix. qmake builds (8d9c8760f), and the branch builds on the five CI systems.
- The 31 print and widget images are byte-identical. In pixel scenes a 0.50 mm line after a text is 8 px (1 px without the fix) and a text after such a line 1 px (8 px without it).
- `RS_Painter::setPen()` counts are unchanged outside panning; while panning, at most one more per switch between a text and a line.

ROMPEFUEGOS:
Opened as #2879, rebased onto 161965f8e and re-verified there.

PR #2883, opened 2026-09-18
Move the DXF color number helpers into the engine

ROMPEFUEGOS:
### What moves

`RS_FilterDXFRW::numberToColor()` and `RS_FilterDXFRW::colorToNumber()` become `LC_ColorNumbers::numberToColor()` and `LC_ColorNumbers::colorToNumber()` in a new engine unit, `librecad/src/lib/engine/lc_colornumbers.{h,cpp}`, next to `LC_LineTypeNames`. This is the follow-up #2880 left out to stay a pure move.

Both bodies move unchanged; only the qualified name differs, so the move can be checked as a copy. The palette they read, `DRW::dxfColors`, stays in libdxfrw's drw_objects.h, which the engine already includes through lc_dwgadvancedmetadata.h, so no dependency is added.

All 44 call sites are requalified and no forwarders stay behind in the filter: 20 in rs_filterdxfrw.cpp, 7 in rs_filterdxf1.cpp, 1 in rs_filtershp.cpp, 8 in lc_dimstyletovariablesmapper.cpp, 2 in lc_tolerance.cpp, 6 in qg_dlgoptionsdrawing.cpp. The three commented-out calls in rs_dimension.cpp are renamed with them.

### What it buys

Five translation units stop including rs_filterdxfrw.h. lc_dimstyletovariablesmapper.cpp, lc_tolerance.cpp, rs_filterdxf1.cpp and rs_filtershp.cpp called these helpers through it, and rs_dimension.cpp only mentioned the filter in comments. Two of them were also getting something else through that header, so they now include it themselves: lc_tolerance.cpp needs rs_mtext.h, and rs_filterdxf1.cpp needs `<cstring>`. In the engine only rs_hatch_tests.cpp still includes the filter: it constructs one and tests its hatch spline helpers. qg_dlgoptionsdrawing.cpp keeps it, since it also calls `numberToUnit()` and `isVariableTwoDimensional()`.

### What it leaves alone

RS_FilterJWW keeps its own pair. It has different signatures and reads jwwlib's palette, which differs from libdxfrw's in 201 of the 255 colors (index 12 is 165,0,0 there and 204,0,0 here), so merging the two would change what JWW reads. RS_FilterDXF holds a third pair but is in neither build list.

No behaviour change: the same numbers map to the same colors, and DXF and DWG output is unaffected.

### Verification

- Test gate: the eleven Catch2 lanes on 3cc2e8cff and on this branch: 7286 tests, the same 41 pre-existing failures, none new.
- qmake6: builds the application and its ten plugins.
- CI: the five pixi builds are green on this commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MCVxVHLgir3pBpPxV3v7Av


librecad/src/ui/dialogs/settings/options_drawing/qg_dlgoptionsdrawing.cpp
```diff
@@ -1117,9 +1118,9 @@ void QG_DlgOptionsDrawing::_toRemoveSetupLegacyDimsTab(RS2::LinearFormat& linear
     const int dimclrd = m_graphic->getVariableInt("$DIMCLRD", 0);
     const int dimclre = m_graphic->getVariableInt("$DIMCLRE", 0);
     const int dimclrt = m_graphic->getVariableInt("$DIMCLRT", 0);
-    cbDimClrD->setColor(RS_FilterDXFRW::numberToColor(dimclrd));
-    cbDimClrE->setColor(RS_FilterDXFRW::numberToColor(dimclre));
-    cbDimClrT->setColor(RS_FilterDXFRW::numberToColor(dimclrt));
+    cbDimClrD->setColor(LC_ColorNumbers::numberToColor(dimclrd));
```
sand1024, line 1121: this is obsolete legacy code that will be removed.... so no need to modify it. actually, in nearest future, the entire settings dialog will be chaged

ROMPEFUEGOS: Those two functions are still compiled: only their callers sit behind `ENABLE_LEGACY_DIMENSIONS_TAB`, so once the helpers leave the filter the old spelling stops building. The rename is the smallest change that keeps it compiling, and it goes away with the rest when you replace the dialog. Happy to handle it differently if you prefer.

sand1024: @ROMPEFUEGOS 
I didn't checked PR in details - yet just wondering - is this PR just about trivial extraction of the method into dedicated utility class?

ROMPEFUEGOS: Yes, that's all it is: the two functions move unchanged from `RS_FilterDXFRW` into an engine namespace, as #2880 did for the linetype names, so the dimension-style mapper and `LC_Tolerance` no longer include the DXF filter header just to reach them. No behaviour change; the rest of the diff is the requalified call sites.

sand1024: @ROMPEFUEGOS 
Oh, I see, got it. 

As for LC_Tolerance - it is just a stub for now. It will be implemented (together with needed creation/editing commands) as soon as other dimensions entities will be rewritten to support rendering according to dimension styles.

I've added support of dimension styles, yet didn't rework dimensions rendering so far, so not all attributes from dim styles affects rendering.

dxli (approved):

PR #2908, opened 2026-09-22
Stop offering JWW as an export format

dxli:
## Decision: stop offering JWW as an export format

JWW becomes import-only, like SHP. `RS_FilterJWW::canExport()` now returns false, and `fileExport()` returns false without touching the file. I did not implement the writers. The evidence is below.

JWW import is not changed by this PR.

## What happens on master today

A probe test at a3ee08bd7 exported a drawing with a line, a circle and a spline:

| Call | Result |
|---|---|
| `RS_FilterJWW::canExport(…, FormatJWW)` | **true** |
| `RS_FilterJWW::fileExport(g, "x.jww", FormatJWW)` | returns false; **no file is created** (debug log: "can't write file") |
| `RS_FileIO::fileExport(g, "x.jww", FormatJWW)` | false, no file |

So JWW export never writes a file at all; it isn't writing an empty one. `DL_Jww::out()` returns `NULL`, so `fileExport()` stops before the first byte. The empty `DL_Jww::write*` bodies are never reached.

## History

- **b91e60cd2** (2011-08-04, "Added experimental support for loading and saving JWW files"). The writers were empty from this first commit. `DL_Jww::out()` had no `return` statement at all until tin-pot's 2011-12-29 MSVC change (0fcee161f) added `return NULL`. The commented-out parameter names date from 6692cb312 (2012, "remove compiler warnings"); nothing was ever removed from the bodies.
- **7f2ab483d** (2013-09-02, "Remove Jww from Save dialogs"). Commit message: *"Jww writing has been disabled for ages, see DL_Jww::out() … Don't pretend we can do it."* It put the dialog entries behind `JWW_WRITE_SUPPORT`, a macro that nothing defines, but it left `canExport()` returning true.
- The unreachable code was still being maintained: #2815 (for issue #2813) and #2823 (2026-09) both edited the linetype table inside `RS_FilterJWW::fileExport()`.
- `DL_Jww`'s constructor never initialises `version`, so every `jww.getVersion()==VER_R12` check in the writer would have read an uninitialised member.

## Why completing the writers from `dl_jww-copy.cpp` would be wrong

- `dl_jww-copy.cpp` is dxflib's `dl_dxf.cpp` renamed: it has the RibbonSoft header, and git follows it back to "Updated DXF Library (dxflib) to 2.2.0.0". Its `out()` opens a `DL_WriterA`, and `writeLine()` emits `LINE` / `AcDbLine` group codes. `RS_FilterJWW::fileExport()` is the matching DXF skeleton: `sectionTables`, `tableLayers`, `BLOCK_RECORD`, `$ACADVER` header variables and so on. Filling in the stubs would put **DXF text into a `.jww` file**.
- A real `.jww` is a binary MFC archive that begins with `JwwData.` and a version number (see `JWWDocument::ReadHeader/WriteHeader` and `jwdatafmt.txt`). The probe wrote DXF text (19,053 bytes, `999 dxfrw … $ACADVER`) under a `.jww` name, and `RS_FilterJWW::fileImport` rejected it ("Cannot open JWW file", 0 entities). LibreCAD could not have re-opened its own export, and Jw_cad would not have read it either.
- A correct writer would be a new feature built on `JWWDocument::Save()`, which has no caller today. The probe shows it can round-trip one `CDataSen` line through its own `Read()` (14,162 bytes, starting `JwwData.`, version 600). Beyond that it would need a JWW header (16 layer groups × 16 layers, scales, paper size), pen and linetype numbering, Shift_JIS text, and approximations for entities JWW lacks, such as splines. There are no JWW sample files in the repo and no Jw_cad to validate against (it's Windows-only), so compatibility couldn't be checked here. If there's demand, that belongs in its own PR.

### Spline knots

The request to write each spline's own knot vector, rather than uniform knots, only applies if the writer is kept. The uniform-knot loop was in `RS_FilterJWW::writeSpline`, which commit 2 deletes. `RS_FilterDXFRW::writeSpline` already writes `getUnwrappedKnotVector()`.

## Commits

A third commit only corrects a test comment: `DL_Jww::out()` returned NULL from 0fcee161f (December 2011); at jwwlib's first commit it had no return statement at all. The first commit's message repeats the older wording.

1. **Stop offering JWW as an export format**: `canExport()` is false and `fileExport()` is an inline `return false`, following `RS_FilterSHP`. Removes the `JWW_WRITE_SUPPORT` branches (`qg_filedialog.cpp`, `lc_filedialogservice.cpp`, and the `DEFINES -=` in `src.pro`) and the unconditional "JWW (*.jww)" entry in the static `QG_FileDialog::getSaveFileName`, which has no callers. Adds `jww_export_tests.cpp`.
2. **Remove the unreachable JWW writer** (−6,731 lines, 4,935 of them the unbuilt `dl_jww-copy.cpp`; can be dropped). Removes `RS_FilterJWW`'s `DL_WriterA` helpers (`writeVariables` … `writeImageDef`, `getEntityAttributes`), the matching `DL_Jww::out()` + 44 `write*` stubs, the writer-only `checkVariable()` / `getVersion()` / uninitialised `version`, and the unbuilt `dl_jww-copy.cpp`. The static conversion helpers (`lineTypeToName` etc.) stay because they are public and used by `dxf_roundtrip_tests`. `DL_WriterA` itself stays because `DL_Jww::toReal()` uses `DL_WriterA::strReplace`.

## Tests

`librecad/src/lib/filters/tests/jww_export_tests.cpp` (tag `[jww]`), 4 cases, 14 assertions:

- `RS_FilterJWW` can import JWW but reports `canExport` false (JWW and JWC).
- `RS_FileIO` has an import filter for JWW and **no export filter**.
- Exporting a line + circle + spline as JWW fails and creates no file, both through `RS_FileIO` and directly through the filter.
- A refused JWW export leaves an existing `.jww` byte-for-byte untouched.

Against the old filter (the test built with `rs_filterjww.{h,cpp}` from a3ee08bd7), 2 cases fail: `canExport` is true, and `getExportFilter` is non-null. The file-safety assertions pass before and after; they lock in the invariant.

Full `librecad_tests` (macOS, Release): 2538 cases, 22 failed, 26 failed assertions. That's the same failure set as the baseline once the two intended `[jww]` failures are removed. Every failure is pre-existing, in `dwg_write_smoke_tests` (20) and `acad_table_tests` (2).

Builds (macOS arm64, Qt 6 Homebrew):
- CMake/Ninja: `all` (app) and `--target librecad_tests` are both clean.
- qmake6: full `librecad.pro` (jwwlib, the app, plugins and tools) is clean. The only warnings in touched files are already on master: `DL_Jww::stringToInt` unused parameters, and `DEFAULT_FILTERS_INDICES` unused.
- No new app sources. The test file goes only in `CMakeLists.txt`, like every other `filters/tests/*.cpp`, because qmake has no Catch2 target.

## Found along the way (not changed here)

- **JWW import hangs on master.** `RS_FilterJWW::toNativeString` has two `do { … } while (!cap.isNull())` loops. `cap` starts as `""` and is never reassigned, because the Qt6 port in ab3e44e28 dropped `cap = regexp.cap();`, and an empty `QString` is never null. So the loops never exit. Every entity reaches them through `addLayer`, so opening any `.jww` spins forever: `sample` shows `DL_Jww::CreateSen → RS_FilterJWW::addLayer → toNativeString`. 2.2.1 still has `cap = regexp.cap()` and is not affected.
- **Save on a drawing opened from a `.jww` overwrites it with DXF.** Every MDI window starts with format type `FormatDXFRW` and loading doesn't change it, so Save writes DXF into the original path. With the hang patched locally, `doSave` reported success, replaced a 14,162-byte `JwwData.` file with 19,186 bytes of DXF that `RS_FilterJWW` can no longer open, and left only the `~` backup. The same probably applies to `.dwg`, `.shp`, `.lff` and `.cxf`.
- **DXF export crashes on a spline with an empty `weights` list, and fails after "append spline point".** `RS_Spline::getUnwrappedWeights()` copies `controlPoints.size()` weights without a bounds check. The append/add/remove-point actions clear `knotslist` and leave `weights` the wrong size.


dxli:
Closing this PR because JWW export capability must remain available. The current branch removes the JWW export path and its dialog entry; that is not acceptable for the active format-support work. Retaining the existing path also keeps the source-format save work in PR #2909 compatible.

